PR #4, opened 2018-02-08
chore(docs): add git commit hash to sourcemap doc

jahtalab:
cc @roncohen

### DIFF
--- a/docs/sourcemap.asciidoc
+++ b/docs/sourcemap.asciidoc
@@ -7,12 +7,12 @@ you can generate source maps for your bundles and upload them to the APM server.
 
 Here is an example to generate source maps with webpack and configure the JavaScript agent accordingly.
 
-First you need to configure webpack to generate source map and provide the app version to your app:
+First you need to configure webpack to generate source map and provide the service version to your app:
 
 [source,js]
 ----
 var webpack = require('webpack')
-var appVersion = require("./package.json").version
+var serviceVersion = require("./package.json").version
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin')
 module.exports = {
   entry: 'app.js',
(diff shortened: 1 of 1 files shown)


docs/sourcemap.asciidoc
```diff
@@ -66,4 +91,28 @@ request.post({url: 'http://localhost:8200/v1/client-side/sourcemaps',formData: f
...
+Or use current git commit hash:
+
+[source,sh]
+----
+APPVERSION=`git rev-parse --short HEAD` && \
+curl http://localhost:8200/v1/client-side/sourcemaps -X POST \
+    -F sourcemap=@./dist/app.min.js.map \
+    -F service_version="$APPVERSION" \
+    -F bundle_filepath="http://localhost/app.min.js" \
+    -F service_name="service-name"
```
simitt: I am a bit confused about the usage of `app_version`. As we have renamed `app` to `service` wouldn't it be more consistent and easier to understand if you'd also use `service` over `app` locally?

jahtalab: I think the root of confusion is because we are using service for what most people would call an "app" and I did voice my concern over this. So devs that come to this documentation, they are used to calling their frontend application an "app", but at some point we should make the transition and make the point that our configuration is called "service".

jahtalab: What I'm saying is that the `service_version="$APPVERSION"` is there to show the user that they're suppose to use `service_version` for what they call "App version".

simitt: I would have probably written it the following: 
```
SERVICE_VERSION=`git rev-parse --short HEAD` && \
curl http://localhost:8200/v1/client-side/sourcemaps -X POST \
   -F sourcemap=@./dist/app.min.js.map \
   -F service_version="$SERVICE_VERSION" \
   -F bundle_filepath="http://localhost/app.min.js" \
   -F service_name="service-name"
```
You are also using `service-name` and not `app-name`, so I don't see how this is consistent.
Anyways, you know the community better than I do, and if you think it is less confusing this way then it is fine with me.

jahtalab:
@simitt , Thanks for the review! I changed app to service as you suggested. I wanted to make the transition smooth but it seems that it might add confusion 😄

PR #37, opened 2018-07-10
fix link to apm server rum config

jalvz:


jalvz:
@jahtalab I am fixing this link as right now the docs build is failing. When you have time, you might want to change it.

PR #38, opened 2018-07-10
fix link to apm server rum config (#37)

jalvz:


jalvz:
@jahtalab pinging you here too

PR #113, opened 2018-12-05
chore: update js-core to v2.1.1

jahtalab:
add standalone-html test

### DIFF
--- a/test/e2e/standalone-html/index.html
+++ b/test/e2e/standalone-html/index.html
@@ -10,13 +10,35 @@
 
 <body>
 </body>
-<script src="https://unpkg.com/elastic-apm-js-base@1.0.0/dist/bundles/elastic-apm-js-base.umd.min.js"></script>
+<div id="app"></div>
+<script src="/dist/bundles/elastic-apm-js-base.umd.min.js"></script>
 <script>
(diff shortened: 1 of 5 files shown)


test/e2e/standalone-html/index.html
```diff
@@ -10,13 +10,35 @@
...
   })
+
+  function checkDtInfo(payload) {
+    console.log('distributed tracing data', payload)
+    if (typeof payload.traceId !== 'string') {
+      throw new Error('Wrong distributed tracing payload: ' + req.responseText)
+    }
+  }
+
+  req = new window.XMLHttpRequest()
```
vigneshshanmugam: any reason not to use fetch? 

Is it because we run tests on IE 10 as well?

jahtalab: In this case I wanted the test to be close to what is usually used in a setup like this. For example it's likely that static websites are just using XHRs without a polyfill for fetch.

Btw, We only run unit tests in all browsers. e2e tests are running on chrome at the moment.

vigneshshanmugam: got it, thanks for the info. You can merge it

PR #118, opened 2018-12-10
chore: update getting started doc with unpkg link

### LINKED ISSUE #97: Provide a way to download the agent and document that in getting started
Options:
* https://unpkg.com/elastic-apm-js-base
* Adding dist folder to the github repo

vigneshshanmugam:
+ Fixes #97

### DIFF
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -48,7 +48,7 @@ Alternatively, you can use a script element and the `elasticApm` global object t
 
 [source,html]
 ----
-<script src="elastic-apm-js-base/dist/bundles/elastic-apm-js-base.umd.min.js"></script>
+<script src="https://unpkg.com/elastic-apm-js-base@2/dist/bundles/elastic-apm-js-base.umd.min.js" crossorigin></script>
 <script>
   elasticApm.init({
     serviceName: '',
@@ -56,5 +56,7 @@ Alternatively, you can use a script element and the `elasticApm` global object t
   })
 </script>
 ----
+
+NOTE: For production usage, we recommend hosting the file `elastic-apm-js-base/dist/bundles/elastic-apm-js-base.umd.min.js` in your CDN for better performance.
 
 NOTE: Currently our minified JavaScript bundle is about 15KB (gzipped).


docs/intro.asciidoc
```diff
@@ -57,4 +57,6 @@ Alternatively, you can use a script element and the `elasticApm` global object t
 </script>
 ----
 
+NOTE: For production usage, we recommend hosting the file `elastic-apm-js-base/dist/bundles/elastic-apm-js-base.umd.min.js` in your CDN for better performance.
```
jahtalab: I think we should also mention that elastic does not maintain `unpkg.com`

docs/intro.asciidoc
```diff
@@ -48,7 +48,7 @@ Alternatively, you can use a script element and the `elasticApm` global object t
 
 [source,html]
 ----
-<script src="elastic-apm-js-base/dist/bundles/elastic-apm-js-base.umd.min.js"></script>
+<script src="https://unpkg.com/elastic-apm-js-base@2/dist/bundles/elastic-apm-js-base.umd.min.js" crossorigin></script>
```
jahtalab: I think we can remove the version from the url, so it always gets the latest version.

Good call to add `crossorigin` 👍

roncohen:
this is a nice change!

I think we should be clear about the recommended deployment method. If the point is that people should copy from unpkg.com and deploy on their own CDN, i would write that out so there's no confusion. And I think we should generally try to include the major version in the filename/url when it makes sense to help people do the right thing when they deploy.

docs/intro.asciidoc
vigneshshanmugam: yeah makes sense, Lets remove the version and go with latest.

vigneshshanmugam:
@roncohen @jahtalab Can you have a look now?

docs/intro.asciidoc
```diff
@@ -44,11 +46,16 @@ var apm = initApm({
 })
 ----
 
-Alternatively, you can use a script element and the `elasticApm` global object to load and initialize the agent:
+=== Using Script Tags
+
+Add a <script> tag to the HTML page and use the `elasticApm` global object to load and initialize the agent:
+
+NOTE: Please download the latest version of JavaScript agent from `https://github.com/elastic/apm-agent-js-base/releases/latest` or
```
bmorelli25: Is there a reason this link and the unpgk link are formatted as code instead of links?
```suggestion
NOTE: Please download the latest version of JavaScript agent from https://github.com/elastic/apm-agent-js-base/releases/latest[GitHub] or
```

docs/intro.asciidoc
```diff
@@ -44,11 +46,16 @@ var apm = initApm({
...
+`https://unpkg.com/elastic-apm-js-base/dist/bundles/elastic-apm-js-base.umd.min.js` and host the file in your Server/CDN before deploying to production.
```
bmorelli25:
```suggestion
https://unpkg.com/elastic-apm-js-base/dist/bundles/elastic-apm-js-base.umd.min.js[UNPKG] and host the file in your Server/CDN before deploying to production.
```

vigneshshanmugam:
Thanks @bmorelli25 ..I have applied your suggestion :)

docs/intro.asciidoc
```diff
@@ -19,6 +19,8 @@ You will also need to {apm-server-ref}/rum.html[enable frontend endpoints in the
 [[getting-started]]
 == Getting started
 
+=== Using Package Managers
```
jahtalab: I think you should add `[float]`, this will make it generated in the same page instead of a separate one. 
Also there should be page ID so we can refer to these sections individually, something like `[[using-package-managers]]`

docs/intro.asciidoc
```diff
@@ -44,11 +46,16 @@ var apm = initApm({
 })
 ----
 
-Alternatively, you can use a script element and the `elasticApm` global object to load and initialize the agent:
+=== Using Script Tags
```
jahtalab: The same here with id and float!

docs/intro.asciidoc
```diff
@@ -44,11 +46,16 @@ var apm = initApm({
...
+
+Add a <script> tag to the HTML page and use the `elasticApm` global object to load and initialize the agent:
+
+NOTE: Please download the latest version of JavaScript agent from https://github.com/elastic/apm-agent-js-base/releases/latest[GitHub] or
```
jahtalab: Thanks for adding the github release page 👍

Issue #111, opened 2018-12-04
Revisit bundling steps 


vigneshshanmugam:
+ [ ] Migrate to webpack 4
+ [ ] Use babel env instead of babel-preset-es2015. Its deprecated already
+ [ ] Uglify js config can be tuned for better results.
+ [ ] Upgrade babel to 7

Should reduce the bundle size of our RUM agent.

PR #123, opened 2018-12-12
feat: Bundling -  move to webpack 4

vigneshshanmugam:
+ First part of #111


jahtalab:
We don't have any e2e tests in js-core, we only have unit tests (using karma and saucelabs). But unit tests actually run against a real APM server which makes them closer to e2e tests, however they don't test our bundling (since that happens in js-base)

In the js-base repo we have e2e tests (some of them test the final bundle) but those are only running in chrome at the moment. I think testing the final bundle across browsers could expose the issues that are created because of our bundling.

vigneshshanmugam:
@jahtalab That makes total sense, thanks for the explanation. I will make the e2e tests that we have and run it across all browsers instead of chrome only.

vigneshshanmugam:
Blocked by #134

PR #138, opened 2019-01-14
feat: add OpenTracing support

jahtalab:
add a separate bundle for opentracing
minor fixes

Requires https://github.com/elastic/apm-agent-js-core/pull/23 to be released!

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "webpack-cli": "^3.1.2"
   },
   "dependencies": {
-    "elastic-apm-js-core": "^2.3.0",
+    "elastic-apm-js-core": "^2.4.0",
     "es6-promise": "^4.2.4"
   },
   "standard": {
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "run-script": "node ./scripts/run-script.js",
     "test": "npm run lint && npm run build && npm run buildE2eBundles && npm run runNodeTests && npm run runUnitTests && npm run run-e2e && npm run build-docs",
     "lint": "standard",
-    "format": "prettier-standard src/**/*.js test/**/*.js"
+    "format": "prettier-standard 'src/**/*.js' test/**/*.js"
   },
   "files": [
     "src/**/*.js",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "run-script": "node ./scripts/run-script.js",
     "test": "npm run lint && npm run build && npm run buildE2eBundles && npm run runNodeTests && npm run runUnitTests && npm run run-e2e && npm run build-docs",
     "lint": "standard",
-    "format": "prettier-standard src/**/*.js test/**/*.js"
+    "format": "prettier-standard 'src/**/*.js' test/**/*.js"
```
vigneshshanmugam: is this required? It works without quotes for me.

test/e2e/standalone-html/index.html
```diff
@@ -9,13 +9,16 @@
   <body>
     <div id="app"></div>
     <script src="/dist/bundles/elastic-apm-js-base.umd.min.js"></script>
+    <script src="/dist/bundles/elastic-apm-opentracing.umd.min.js"></script>
```
vigneshshanmugam: are we planning to release opentracing support as separate script instead?

src/opentracing/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+function createTracer (apmBase) {
+  return createElasticTracer(apmBase.serviceFactory)
+}
+
+if (window && window.elasticApm) {
+  window.elasticApm.createTracer = createTracer.bind(window.elasticApm, window.elasticApm)
+}
+
+module.exports = {
```
vigneshshanmugam: we dont need to do this, build script takes care of it.

package.json
jahtalab: without quotes it doesn't format the top level files `/src/*.js`

test/e2e/standalone-html/index.html
jahtalab: I think that's a better approach, considering the feature might not be used in a lot of use-cases. I would prefer to keep our main bundle size small.

What do you think?

jahtalab:
@vigneshshanmugam , Would you please have another look at this one?
The docs PR is: https://github.com/elastic/apm-agent-js-base/pull/140

PR #142, opened 2019-01-17
chore: add license checker during linting phase

vigneshshanmugam:
+ to be OSS complaint

### DIFF
--- a/scripts/license-checker.js
+++ b/scripts/license-checker.js
@@ -1,0 +1,111 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2017-present, Elasticsearch BV
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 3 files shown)


vigneshshanmugam:
@jahtalab Rebased against master.

scripts/license-checker.js
```diff
@@ -0,0 +1,111 @@
...
+  'CC0-1.0',
+  'CC-BY-3.0',
+  'CC-BY-4.0',
+  'Unlicense'
+]
+
+/**
+ * Packages that must be ignored during license check
+ */
+const IGNORE_LIST = ['colors@0.6.2']
```
jahtalab: Why is this necessary?

vigneshshanmugam: They have a custom license which is already fixed in colors@1.0 so this is a workaround to avoid the check for older versions.

vigneshshanmugam:
```
license: Custom: http://i.imgur.com/goJdO.png
```

PR #162, opened 2019-02-04
chore: Monorepo first draft

jahtalab:


dev-utils/test.js
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  process.on('exit', exitHandler)
+
+  process.on('SIGINT', exitHandler)
+
+  process.on('uncaughtException', exitHandler)
+}
+
+function startSelenium (callback, manualStop) {
```
vigneshshanmugam: dead code, we don't use selenium anymore for E2E

dev-utils/test.js
```diff
@@ -0,0 +1,236 @@
...
+  }
+  if (runSelenium) {
+    startSelenium(runWdio)
+  } else {
+    runWdio()
+  }
+}
+
+module.exports = {
+  runUnitTests: runUnitTests,
```
vigneshshanmugam: nit: Use object shorthand properties. Can you also add https://eslint.org/docs/rules/object-shorthand to eslint rules

package.json
```diff
@@ -1,39 +1,18 @@
...
-    "runNodeTests": "node ./scripts/run-script.js runNodeTests",
-    "buildE2eBundles": "node ./scripts/run-script.js buildE2eBundles",
-    "run-e2e": "node ./scripts/run-script.js runE2eTests",
-    "run-script": "node ./scripts/run-script.js",
-    "test": "npm run lint && npm run build && npm run buildE2eBundles && npm run runNodeTests && npm run runUnitTests && npm run run-e2e && npm run build-docs",
     "lint": "eslint . && node scripts/license-checker",
-    "format": "eslint . --fix"
+    "build-docs": "sh ./scripts/build_docs.sh apm-agent-js-base ./docs ./build",
+    "format": "eslint . --fix",
+    "test": "npm run lint && lerna bootstrap && lerna run test --stream && npm run build-docs"
```
vigneshshanmugam: bootstrap shouldnt be called during test TBH, before setup one is supposed to run bootstrap and have lerna configure the links.

packages/rum-core/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "prepare-major": "sh ../../dev-utils/prepare-release major && npm run push-release",
+    "push-release": "git push origin master --follow-tags && npm publish",
+    "changelog": "node node_modules/conventional-changelog-cli/cli.js  -p angular -i CHANGELOG.md -s",
+    "karma": "karma start",
+    "karma-bench": "karma start karma.bench.conf.js",
+    "karma-coverage": "karma start --coverage",
+    "runUnitTests": "node ../../dev-utils/run-script.js runUnitTests true packages/rum-core",
+    "test": "npm run runUnitTests",
+    "lint": "standard",
+    "format": "prettier-standard 'dev-utils/**/*.js' src/**/*.js test/**/*.js"
```
vigneshshanmugam: Its better to remove prettier-standard and follow the eslint configs that we had for js-base.. its much better and works for licensing header checks.. 

We can do this also as separate PR if you feel its too much.

packages/rum-core/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "files": [
+    "src/**/*.js",
+    "test/**/*.js",
+    "dev-utils/**/*"
```
vigneshshanmugam: nit: remove test and dev-utils from publishing.

jahtalab: I've removed `dev-utils` since it's on the root now. we still use test from js-core in js-base, let's address this separately.

packages/rum-core/package.json
jahtalab: The lint script is on the root package.json, I've removed these scripts now!

dev-utils/test.js
jahtalab: Let's address this in a separate PR.

vigneshshanmugam: Sure

jahtalab:
Thanks @vigneshshanmugam , I've addressed some of the review comments.

I believe these Items are already addressed in this PR:

1. A single change on the rum-base should go through the below steps
    1. Lint all files (including core and base)
    2. Unit tests running on rum-core
    3. Unit tests running for rum-base( bundle rum-core if necessary)
    4. Integration and E2E tests after bundling core

Regarding publish, I think we should be able to publish each package separately. But the build scripts are add to the root package in `dev-utils` directory. 

For some of the other items you mentioned we already have an issue or an item in the check list here: https://github.com/elastic/apm-agent-js-base/issues/146

Please add a check list item (or an issue) if one doesn't already exist, and we can address them separately.

vigneshshanmugam:
@jahtalab Yeah I just wanted to list out the items that would be good to have, We can address it in separate PR's as this PR is a draft and we can continue working on it. 

Yeah the publishing part, It would be better to go through lerna publish as we are leveraging lerna already for the repo management, We can do this again as separate PR and lets not block this PR. 

> Please add a check list item (or an issue) if one doesn't already exist, and we can address them separately.

Sure, I will do that. 

Right now this PR is in a good state, we can merge it to unblock ourselves and work separately on specific tasks. What do you think?

dev-utils/test.js
vigneshshanmugam: @hamid can you remove this dead code and selenium deps?

vigneshshanmugam: Postponing it till we find a way to run E2E locally without Sauce labs

PR #164, opened 2019-02-05
chore: add changelog generation code

vigneshshanmugam:
+ part of elastic/apm-agent-js-base#146 
+ We rely on Lerna for publishing the packages which means Lerna will do most of the things like 
   - creating tags
   - changelog creation
   - pushing the tags to github
   - publishing the package to npm

**Note:** `CHAGELOG.md will be created on package level basis and we have to manually copy paste if we need changelog at root folder.`

### Follow up Tasks that depend on this PR
- [x] Fixing github release script for js-base elastic/apm-agent-js-base#166 

LERNA would enable support for github releases, we can completely eliminate our custom release script once its landed https://github.com/lerna/lerna/issues/1513


package.json
```diff
@@ -8,6 +8,7 @@
     "url": "https://github.com/elastic/apm-agent-js-base.git"
   },
   "scripts": {
+    "release": "lerna publish",
```
jahtalab: How can we still tell it to release only one package?

Imagine there're changes in both packages but we only want to release one of the at the moment.

vigneshshanmugam: The changes are automatically picked up, So if changes are in js-core both would be published.. if the changes are only on js-base, Only the js-base would be published... 

I will work on individual releasing and docs(Release.md) in separate PR.

vigneshshanmugam:
@jahtalab If the branches are not in sync, publish wont happen as lerna restricts it and has the same functionality that our scripts have.. Thats why I did remove them..

Even for individual releases, I am going to use lerna and rely on that for tag creation since it can do lots of things for us.

packages/rum-core/package.json
```diff
@@ -4,25 +4,16 @@
...
-    "changelog": "node node_modules/conventional-changelog-cli/cli.js  -p angular -i CHANGELOG.md -s",
     "karma": "karma start",
     "karma-bench": "karma start karma.bench.conf.js",
     "karma-coverage": "karma start --coverage",
     "runUnitTests": "node ../../dev-utils/run-script.js runUnitTests true packages/rum-core",
     "test": "npm run runUnitTests"
   },
   "files": [
-    "src/**/*.js",
-    "test/**/*.js"
```
jahtalab: There's a reference to `elastic-apm-js-core/test` in `test/e2e.js` in the js-base. would you please either refactor that one or revert this change?

package.json
jahtalab: Thanks! looking forward to that one then 👍

packages/rum/test.config.js
```diff
@@ -1,26 +1,49 @@
...
  * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
- * 
+ *
  */
 
-module.exports = require('../../test.config')
\ No newline at end of file
+const testUtils = require('../../dev-utils/test')
```
jahtalab: I think this should be in the root directory, since we will be using it for other packages!

packages/rum-core/package.json
vigneshshanmugam: that shouldn't be a problem, since we are in mono repo this should just work without any problem.. I just removed it from npm publishing part.

packages/rum/test.config.js
vigneshshanmugam: It just feels odd to have test file in the root since we dont have any tests dir and thought it belonged to base since it we are using it only here.

vigneshshanmugam: I am moving it to devutils/test.js as function

packages/rum-core/package.json
jahtalab: Yeah, but let's not have invalid reference!

PR #166, opened 2019-02-05
chore: fix release script for github artificats

vigneshshanmugam:
+ part of elastic/apm-agent-js-base#146 
+ This happens after we do `lerna publish` which is covered in another PR elastic/apm-agent-js-base#164 
+ fixes elastic/apm-agent-js-base#157 
+ The tagging issue should not be a problem anymore since we rely on tag creation via lerna /cc @jahtalab

### DIFF
--- a/scripts/release-github.js
+++ b/scripts/release-github.js
@@ -28,9 +28,9 @@ const url = require('url')
 const fs = require('fs')
 const https = require('https')
 const releaseAssets = require('gh-release-assets')
-const { version } = require('../package.json')
+const { version } = require('../packages/rum/package.json')
 
-const BUILD_DIR = path.resolve(__dirname, '../dist/bundles')
+const BUILD_DIR = path.join(__dirname, '../packages/rum/dist/bundles')
 const GITHUB_URL = 'https://api.github.com/repos/elastic/apm-agent-js-base'
 
 function createRelease (token) {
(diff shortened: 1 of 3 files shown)


scripts/release-github.js
```diff
@@ -28,9 +28,9 @@ const url = require('url')
 const fs = require('fs')
 const https = require('https')
 const releaseAssets = require('gh-release-assets')
-const { version } = require('../package.json')
+const { version } = require('../packages/rum/package.json')
```
vigneshshanmugam: we would still have to decide how to version the release if we want to release two packages separately

jahtalab: For now let's keep using `v3.0.0` and publish artefacts for the js-base. We can revisit this later

vigneshshanmugam: Do you want to follow different conventions for js-base and js-core then? It will collide with other packages then, We have to follow some other convention. Thats why its better to do it via <packageName>@version as lerna does.

vigneshshanmugam:
@jahtalab Regarding commit tags yes, I would just reply on lerna for it. Since we are using for both individual publishing and publishing all packages. I am writing doc and fixing the scripts in another PR #171

scripts/release-github.js
```diff
@@ -46,7 +46,7 @@ function createRelease (token) {
   /**
    * To match the package version with tags
    */
-  const tagVersion = 'v' + version
+  const tagVersion = `${name}@${version}`
```
vigneshshanmugam: @jahtalab We are relying on lerna tag creation here. Check details here https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-core#lernapackage

PR #167, opened 2019-02-06
chore: move to build stages to reduce test time

vigneshshanmugam:
+ Trying to combine builld matrics and build stage to reduce test time. 

Tasks
- [x] APM server docker should happen only for tests
- [x] Docs should be separate step after all tests passes and should happen only on master branch
- [x] Matrix test includes e2e and tests from js-core

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,9 @@ services:
   - docker
 node_js:
 - '8'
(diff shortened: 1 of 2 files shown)


vigneshshanmugam:
Closing in favor of #168

PR #171, opened 2019-02-07
chore: publish individual packages via lerna

vigneshshanmugam:
+ Publish individual packages via Lerna 
+ `release.md` which contains information on publishing

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,5 @@
+# Releasing
+
+## Publishing all packages
+
+### Publishing single package
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,9 +15,11 @@ To publish all the packages run `npm run release`. It will run `lerna publish`(w
 - Generating `CHANGELOG.md` using conventional commit messages.
 - Publishing the packages via `npm publish`
 
+**Note: To override the semantic version bumb, use `npm run release -- (minor|patch|major)`**
+
 ### Releasing single package
 
-To publish a single package, run `npm run release-package -- elastic-apm-js-base` which uses `lerna publish` under the hood to publish a single package to NPM
+To publish a single package, run `npm run release-package -- elastic-apm-js-base` which uses `lerna publish` under the hood to publish single package to NPM
 
 ### Releasing artifacts to GitHub
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -8,9 +8,9 @@ Before releasing a new version of packages to the NPM and GitHub, Its advisable 
 
 ### Releasing all packages
 
-To publish all the packages run `npm run publish`. It will run lerna publish (we use --independent) so it will prompt the version number for
-every package. `lerna publish` will take care of the following
+To publish all the packages run `npm run release`. It will run `lerna publish`(we use --independent) so it will prompt the version number for every package.
 
+- Automatically determining a semantic version bump (based on the types of commits landed)
 - Creating annonated Git tags for all the packages that has changed since last release
 - Generating `CHANGELOG.md` using conventional commit messages.
 - Publishing the packages via `npm publish`
@@ -21,4 +21,6 @@ To publish a single package, run `npm run release-package -- elastic-apm-js-base
 
 ### Releasing artifacts to GitHub
 
-Use `npm run release-github` in the root directory, the script takes care of building the `elastic-apm-js-base` files and creating a release in GitHub with the previous annonated tag.
+Use `npm run release-github` in the root directory, the script takes care of building, testing the `elastic-apm-js-base` package and creating a release in GitHub with the previous annonated tag.
+
+**Note: Make sure you pass the GITHUB_TOKEN (with push access) in your environment variable while releasing**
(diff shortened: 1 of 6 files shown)


RELEASE.md
```diff
@@ -0,0 +1,24 @@
...
+
+Before releasing a new version of packages to the NPM and GitHub, Its advisable to follow the below steps in order
+
+- Update the dependencies `npm run clean`
+- Bootstrap the deps using `npm run bootstrap`
+- Make sure all tests passes `npm run test`
+
+### Releasing all packages
+
+To publish all the packages run `npm run publish`. It will run lerna publish (we use --independent) so it will prompt the version number for
```
jahtalab: Do you mean `npm run release` instead of `npm run publish`?

RELEASE.md
```diff
@@ -0,0 +1,24 @@
...
+every package. `lerna publish` will take care of the following
+
+- Creating annonated Git tags for all the packages that has changed since last release
+- Generating `CHANGELOG.md` using conventional commit messages.
+- Publishing the packages via `npm publish`
+
+### Releasing single package
+
+To publish a single package, run `npm run release-package -- elastic-apm-js-base` which uses `lerna publish` under the hood to publish a single package to NPM
```
jahtalab: Is it possible to merge this with the publish script? Maybe something to accept a variable like `scope` to provide the package name.

RELEASE.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Releasing artifacts to GitHub
+
+Use `npm run release-github` in the root directory, the script takes care of building the `elastic-apm-js-base` files and creating a release in GitHub with the previous annonated tag.
```
jahtalab: I think we should have one script in the package.json that does all of this including doing the tests and other checks

RELEASE.md
```diff
@@ -0,0 +1,24 @@
...
+Before releasing a new version of packages to the NPM and GitHub, Its advisable to follow the below steps in order
+
+- Update the dependencies `npm run clean`
+- Bootstrap the deps using `npm run bootstrap`
+- Make sure all tests passes `npm run test`
+
+### Releasing all packages
+
+To publish all the packages run `npm run publish`. It will run lerna publish (we use --independent) so it will prompt the version number for
+every package. `lerna publish` will take care of the following
```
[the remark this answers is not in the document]
vigneshshanmugam: Yes lerna publish accepts

`[major | minor | patch | premajor | preminor | prepatch | prerelease]`

RELEASE.md
vigneshshanmugam: oops yeah, you are right.

RELEASE.md
vigneshshanmugam: scope in general is tied to `lerna run` command.. Since its our own custom script, We can use scope though. Do you want like that?

RELEASE.md
vigneshshanmugam: Already covered it here, https://github.com/elastic/apm-agent-js-base/pull/166/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R11

so it will be `release-github: "npm run build && npm run test && node scripts/release-gihtub.js"`

RELEASE.md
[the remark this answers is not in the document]
vigneshshanmugam: lerna publish under the hood automatically determining a semantic version bump since we follow the conventional commits, 

But if we want to override this behaviour, We could still pass in like `npm run release -- (minor|major|patch)`

RELEASE.md
jahtalab: I don't have a strong preference in regards to the name of the variable, although I do think `scope=elastic-apm-js-base` is more clear than just ` -- elastic-apm-js-base`. 

But my main point was to have one script that handles the release instead of two.

RELEASE.md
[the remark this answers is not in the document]
jahtalab: Would you please add that to the RELEASE as well?

lerna.json
```diff
@@ -8,6 +8,7 @@
     "version": {
       "allowBranch": "master",
       "conventionalCommits": true,
+      "message": "chore(release): publish",
```
jahtalab: Can we also have the published versions in the commit message?

vigneshshanmugam: We can only do that in fixed versioning, message wont work for independent versions. 

> If the message contains %s, it will be replaced with the new global version version number prefixed with a "v". If the message contains %v, it will be replaced with the new global version version number without the leading "v". Note that this placeholder interpolation only applies when using the default "fixed" versioning mode, as there is no "global" version to interpolate when versioning independently.

RELEASE.md
vigneshshanmugam: Hmmm, the reason i kept it at separate is because of failures. We can do that later once we feel conformable with the release process.

RELEASE.md
```diff
@@ -0,0 +1,28 @@
...
+### Releasing all packages
+
+To publish all the packages run `npm run release`. It will run `lerna publish`(we use --independent) so it will prompt the version number for every package.
+
+- Automatically determining a semantic version bump (based on the types of commits landed)
+- Creating annonated Git tags for all the packages that has changed since last release
+- Generating `CHANGELOG.md` using conventional commit messages.
+- Publishing the packages via `npm publish`
+
+**Note: To override the semantic version bumb, use `npm run release -- (minor|patch|major)`**
```
jahtalab: "bump"

vigneshshanmugam: oops, Fixed

PR #173, opened 2019-02-11
feat(rum-core): measure all resource entries in page load

vigneshshanmugam:
+ Turned out to be a bit more extensive that I thought, There are multiple resource types that are being captured by `perf.getEntriesByType("resource")`.. List is specified here - https://github.com/web-platform-tests/wpt/blob/b0020d5df18998609b38786878f7a0b92cc680aa/resource-timing/resource_initiator_types.html#L93

+ I have added as check to create Spans for all known resource types except (embed, video, audio etc) which does not make much sense since its uncommon, We can support it in future. 

+ Refactored the code a bit to look nicer. 

+ fixes https://github.com/elastic/apm-agent-js-base/issues/161

### DIFF
--- a/packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
@@ -25,7 +25,10 @@
 
 const Span = require('./span')
 const Url = require('../common/url')
-const { RESOURCE_INITIATOR_TYPES } = require('../common/constants')
+const {
+  RESOURCE_INITIATOR_TYPES,
+  SPAN_THRESHOLD
+} = require('../common/constants')
 
 /**
  * Navigation Timing Spans
@@ -50,10 +53,6 @@ const eventPairs = [
   ],
   ['loadEventStart', 'loadEventEnd', 'Fire "load" event']
 ]
-/**
- * Maximum duration of the span that is used to decide if the span is valid - 30 seconds
- */
-const SPAN_THRESHOLD = 5 * 60 * 1000
 
 function isValidSpan (transaction, span) {
   const duration = span.duration()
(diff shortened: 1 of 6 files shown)


packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -25,72 +25,139 @@
...
+      !name
     ) {
-      // is ajax request? test for css/img before the expensive operation
-      var foundAjaxReq = false
-      for (var j = 0; j < filterUrls.length; j++) {
-        var idx = name.lastIndexOf(filterUrls[j])
+      continue
+    } else if (!isValidPerformanceTiming(startTime, responseEnd)) {
+      continue
+    } else if (RESOURCE_INITIATOR_TYPES.indexOf(initiatorType) !== -1) {
```
jahtalab: This doesn't have to be part of the else if. since the above ifs have continue in them you can start another if else block after.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -99,49 +166,28 @@ function createResourceTimingSpans (entries, filterUrls) {
       if (foundAjaxReq) {
```
jahtalab: Maybe `if(!foundAjaxReq) { spans.push(...) }` is more clear!

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -25,72 +25,139 @@
...
+  [
+    'domContentLoadedEventStart',
+    'domContentLoadedEventEnd',
+    'Fire "DOMContentLoaded" event'
+  ],
   ['loadEventStart', 'loadEventEnd', 'Fire "load" event']
 ]
-const spanThreshold = 5 * 60 * 1000
+/**
+ * Maximum duration of the span that is used to decide if the span is valid - 30 seconds
```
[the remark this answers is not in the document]
vigneshshanmugam: my bad, thanks for the catch 👍

PR #174, opened 2019-02-11
perf(rum-core): avoid url parsing on resource timing entries

vigneshshanmugam:
+ Resource timing entries name have fully resolved URL as part of the SPEC and we dont really need to  parse them again based on the window base URL. 

+ The context information is sent only when the Entry Name is different from Span name -> Happens when the resource contains query strings in URL. small optimisation for reducing the payload to APM server.

### DIFF
--- a/packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
@@ -24,11 +24,11 @@
  */
 
 const Span = require('./span')
-const Url = require('../common/url')
 const {
   RESOURCE_INITIATOR_TYPES,
   SPAN_THRESHOLD
 } = require('../common/constants')
+const { stripQueryStringFromUrl } = require('../common/utils')
 
 /**
  * Navigation Timing Spans
@@ -102,15 +102,17 @@ function createResourceTimingSpan(name, initiatorType, start, end) {
   if (initiatorType) {
     kind += '.' + initiatorType
   }
+  const spanName = stripQueryStringFromUrl(name)
+  const span = new Span(spanName, kind)
+  /**
+   * Add context information when the span name and entry name are different
+   */
+  if (spanName !== name) {
+    span.addContext({
+      http: { url: name }
+    })
+  }
 
-  const parsedUrl = new Url(name)
-  const spanName = parsedUrl.origin + parsedUrl.path
-  const span = new Span(spanName || name, kind)
-  span.addContext({
-    http: {
-      url: name
-    }
-  })
   span._start = start
   span.ended = true
   span._end = end
--- a/packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
@@ -104,15 +104,7 @@ function createResourceTimingSpan(name, initiatorType, start, end) {
   }
   const spanName = stripQueryStringFromUrl(name)
   const span = new Span(spanName, kind)
-  /**
-   * Add context information when the span name and entry name are different
-   */
-  if (spanName !== name) {
-    span.addContext({
-      http: { url: name }
-    })
-  }
-
+  span.addContext({ http: { url: name } })
   span._start = start
   span.ended = true
   span._end = end
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -238,12 +238,12 @@ class PerformanceMonitoring {
   }
 
   createTransactionDataModel(transaction) {
-    var configContext = this._configService.get('context')
-    var stringLimit = this._configService.get('serverStringLimit')
-    var transactionStart = transaction._start
+    const configContext = this._configService.get('context')
+    const stringLimit = this._configService.get('serverStringLimit')
+    const transactionStart = transaction._start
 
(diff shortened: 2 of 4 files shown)


packages/rum-core/src/performance-monitoring/transaction.js
```diff
@@ -91,11 +91,10 @@ class Transaction extends SpanBase {
     if (this.ended) {
       return
     }
-    var transaction = this
-    var opts = extend({}, options)
+    const opts = extend({}, options)
 
-    opts.onEnd = function (trc) {
-      transaction._onSpanEnd(trc)
+    opts.onEnd = trc => {
```
jahtalab: I think it's better to keep this a function since we're assigning it on the options. It's not going to be a problem with our current code, but I think it's better to be explicit.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -115,14 +115,16 @@ function createResourceTimingSpans (entries, filterUrls) {
...
-          http: {
-            url: name
-          }
-        })
+        var spanName = stripQueryStringFromUrl(name)
+        var span = new Span(spanName, kind)
+        /**
+         * Add context information when the span name and entry name are different
+         */
+        if (spanName !== name) {
```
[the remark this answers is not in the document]
vigneshshanmugam: Its unnecessary payload info at the cost of convenience, I would prefer first.

packages/rum-core/src/performance-monitoring/transaction.js
vigneshshanmugam: We transpile it anyway which gets converted to Es5. I did this because some of the code were Es6 and rest were Es5. Its better to move everything to ES6 and bundle accordingly

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -64,11 +72,13 @@ class PerformanceMonitoring {
...
       if (
-        (task.source === XMLHTTPREQUEST_SOURCE && !globalState.fetchInProgress) ||
+        (task.source === XMLHTTPREQUEST_SOURCE &&
+          !globalState.fetchInProgress) ||
         task.source === FETCH_SOURCE
       ) {
         if (event === SCHEDULE && task.data) {
-          var spanName = task.data.method + ' ' + stripQueryStringFromUrl(task.data.url)
+          var spanName =
+            task.data.method + ' ' + stripQueryStringFromUrl(task.data.url)
```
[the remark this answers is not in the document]
vigneshshanmugam: In the case of XHR - browser throws if its not string.. 
In the case of Fetch - We check and modify it already in our patch logic.

Felt unnecessary IMO.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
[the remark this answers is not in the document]
jahtalab: Do you have the payload size that we're saving with this change? of course, it depends on the use-case but let's have some numbers before making the change.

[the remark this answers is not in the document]
vigneshshanmugam: Yeah check this one, step 2 has it. https://github.com/elastic/apm-agent-js-base/issues/158#issuecomment-459349290

vigneshshanmugam:
@jahtalab Do you need me to change anything?  Or shall i rebase and we can merge this?

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -282,30 +282,27 @@ class PerformanceMonitoring {
...
     }
   }
 
   sendTransactions(transactions) {
-    var payload = transactions
-      .map(this.createTransactionPayload.bind(this))
-      .filter(function(tr) {
-        return tr
-      })
+    const payload = transactions.map(tr => this.createTransactionPayload(tr))
```
jahtalab: I think the filter part is necessary to avoid having undefined in the array due to filtering!

```
      .filter(function(tr) {
        return tr
      })
```

vigneshshanmugam: Good catch, we need to call filter first and map to avoid null values in array.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -102,15 +102,17 @@ function createResourceTimingSpan(name, initiatorType, start, end) {
   if (initiatorType) {
     kind += '.' + initiatorType
   }
+  const spanName = stripQueryStringFromUrl(name)
+  const span = new Span(spanName, kind)
+  /**
+   * Add context information when the span name and entry name are different
+   */
+  if (spanName !== name) {
```
jahtalab: @vigneshshanmugam , I think we should consider this a breaking change since we're removing a field that users might have relied on!

vigneshshanmugam: I am not sure here, I don't have any numbers on how much users search for context in kibana search UI. What we can do is to split this change and do in separate PR and not combine with this performance optimisation.

jahtalab: I think generally we should keep in mind how each change affects the users, since we're open source we might never know some of the use-cases!

Yeah, let's split to avoid having multiple changes in the same PR.

PR #178, opened 2019-02-11
chore: modify notice generation script [skip ci]

vigneshshanmugam:
+ Generate notices for all packages. 
+ Notice and Readme is automatically published with NPM, so no need of including in files array in package.json

### DIFF
--- a/dev-utils/dep-info.js
+++ b/dev-utils/dep-info.js
@@ -26,7 +26,7 @@
 const { join } = require('path')
 const { readFileSync, writeFileSync, readdirSync } = require('fs')
 
-function getFileContent(directory, filename) {
+function getFileContent (directory, filename) {
   var files = readdirSync(directory)
   var licenseFile = files.find(f => f.toLowerCase().startsWith(filename.toLowerCase()))
   if (licenseFile) {
@@ -35,11 +35,11 @@ function getFileContent(directory, filename) {
   }
 }
 
-function generateDependencyInfo(deps, modulesPath) {
+function generateDependencyInfo (deps, modulesPath, rootLicense) {
   var allLicenses = []
   deps.forEach(d => {
     var modulePath = join(modulesPath, d)
-    var license = getFileContent(modulePath, 'LICENSE')
+    var license = getFileContent(modulePath, 'LICENSE') || rootLicense
     var dep = {
       name: d
     }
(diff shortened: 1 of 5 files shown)


packages/rum/NOTICE.txt
```diff
@@ -6,27 +6,6 @@ Copyright (c) 2017-present, Elasticsearch BV
 ---
 This product relies on elastic-apm-js-core
 
-MIT License
```
vigneshshanmugam: @jahtalab Since we have the license in root repo for `js-core`, It wont be copied here. Should we do something here? Or is it good to be like this?

jahtalab: I think we need to include the full text of the license with the artefacts we publish (e.g. to npm). Also, most packages I've seen have a license file in their package when installed from npm.

vigneshshanmugam: When Lerna does publishing, the LICENSE is automatically included for all packages, So dont worry about that part. What i meant here is if we have to manually traverse the root directory for  copying LICENSE in Notice.txt

vigneshshanmugam: LICENSE, README and NOTICE - all the three will be published while doing lerna publish for all the packages, 

The reason why elastic-apm-js-core LICENSE info missed in notice.txt is because we have a root level LICENSE and not package level, We either have to change script
1. to copy license in each package before running the `genNotice` script
2. Traverse the root directory for all deps under `/packages/`

dev-utils/dep-info.js
```diff
@@ -56,32 +56,37 @@ function generateDependencyInfo (deps, modulesPath) {
...
-  const { dependencies, name } = JSON.parse(
-    readFileSync(path.join(rootDir, './package.json'), 'utf8')
-  )
-  var depInfo = generateDependencyInfo(
-    Object.keys(dependencies),
-    path.join(rootDir, './node_modules')
-  )
-  var allLicenses = `
+function generateNotice() {
+  const packagesDir = join(__dirname, '../packages')
```
jahtalab: Why not accept a parameter here instead? IMO, it's more usable if it has parameter instead of being hardcoded!

packages/rum/NOTICE.txt
jahtalab: Ok, I misunderstood then! What about adding a parameter to the generateNotice with a list of internal package names and a path to the the license file in the root?

vigneshshanmugam: Hmmm, feel like too much explicit info, We can grab the internal package names from the directory passed and then we can read the root level. I will go with option 2 and also take your consideration in to account.

dev-utils/dep-info.js
vigneshshanmugam: Sure, just thought this script wont be used anywhere.

Issue #108, opened 2018-12-04
Refactor agent and rename


alvarolobato:
We need to refactor the packaging in the agent to:
github repo: apm-agent-rum-js
npm: `@elastic/apm-rum`

- [x] Change the name of this repo
- [x] update the docs `conf.yml` to fetch the new repo. Even though Github automatically sets a redirect for the old url, we should update it to be safe. https://github.com/elastic/docs/pull/676
- [x] rename the bundle to `elastic-apm-rum` https://github.com/elastic/apm-agent-rum-js/pull/202
- [x] Publish the module to npm under the new name
- [x] Build docs to new `rum-js` URL: https://github.com/elastic/docs/pull/958
- [x] Redirect all docs to the new url https://github.com/elastic/website-www.elastic.co/issues/3674
- [x] Remove `js-base` form the code base (docs: https://github.com/elastic/apm-agent-rum-js/pull/378)
- [x] Fix Kibana instructions https://github.com/elastic/kibana/pull/32903 https://github.com/elastic/kibana/pull/32902 https://github.com/elastic/kibana/pull/32901 https://github.com/elastic/kibana/pull/32896
- [x] Update instructions in the solutions page https://github.com/elastic/content/issues/2972
- [x] Add deprecation warning to the old npm package

Targeting version 4.

Postponed:
- [x] Fix Kibana agent key https://github.com/elastic/kibana/issues/32643 (planned for 7.1)

jahtalab:
@bmorelli25 , Is there a way for us to change the [url prefix](https://github.com/elastic/docs/blob/master/conf.yaml#L918) for the docs while still be able to redirect old prefix to the new one.

bmorelli25:
So we're looking at changing the repo from `js-base` to `rum-js`? We can swap the prefix in the doc build and ask the web team to set up a redirect from: 
`https://www.elastic.co/guide/en/apm/agent/js-base/*` to `https://www.elastic.co/guide/en/apm/agent/rum-js/*`

Timing would be key to making this all work. When do you anticipate making this change?

jahtalab:
I'm planning to do the required changes in the following steps:

* Publish the module to npm under the new name (it's currently being done)
* Change the name of this repo, we need to update the docs `conf.yml` to fetch the new repo. Even though Github automatically sets a redirect for the old url, we should update it to be safe.
[...]

PR #179, opened 2019-02-11
BREAKING CHANGE(rum): resolve main field to source file

vigneshshanmugam:
+ This is a bug, We should allow the users of our library to do necessary optimisations(Preferring ES6 or ES5) on the lib.

### DIFF
--- a/packages/rum/package.json
+++ b/packages/rum/package.json
@@ -2,7 +2,7 @@
   "name": "elastic-apm-js-base",
   "version": "3.0.0",
   "description": "Elastic APM JavaScript agent",
-  "main": "dist/bundles/elastic-apm-js-base.umd.js",
+  "main": "src/index.js",
   "repository": {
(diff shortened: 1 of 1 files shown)


packages/rum/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "elastic-apm-js-base",
   "version": "3.0.0",
   "description": "Elastic APM JavaScript agent",
-  "main": "dist/bundles/elastic-apm-js-base.umd.js",
+  "main": "src/index.js",
```
jahtalab: @vigneshshanmugam, We should at least consider this a breaking change.

 Let's wait until the next major version and then we can merge this one.

vigneshshanmugam: Sure, I am good with that.

vigneshshanmugam:
Blocked by #108

PR #183, opened 2019-02-14
chore: delete package level lock files

vigneshshanmugam:
+ Lock files for packages should not be checked in since it would be wrong because of hoisting and symlinking. only root package-locks are useful for monorepos.


vigneshshanmugam:
@jahtalab Npm install would be broken if we have lock files and its recommended in the case of monorepo to only have root level lock files which stores the info. Root level lock files should guarantee that versioning ( yarn workspaces do this AFAIK) https://yarnpkg.com/lang/en/docs/workspaces/#toc-limitations-caveats

While publishing the package, we don't publish lock files to npm as well.

jahtalab:
I don't mean to publish package-lock but I mean when we install the packages before making the  bundle (on your local machine for example), it's better to have a predictable dependency tree! I don't think the root level package-lock would guarantee the versioning of dependencies of each package!

vigneshshanmugam:
My bad, I meant the the tools like lerna/yarn would hoist those dependencies used by those packages and would version them accordingly. In case of version mismatch, those deps would be automatically installed instead of symlinking (yarn does this). Right now our package-lock file is wrong since js-base has no information about our js-core deps since its a symlink. 

If we want to have lock files and do this correctly, we have to include the dependencies `elastic-apm-js-base` and `elastic-apm-js-core` and link them through file protocol. Then our installation would be guaranteed as you would expect. otherwise this is never going to be the case.

PR #186, opened 2019-02-15
chore: add repository.directory field to package.json[skip ci]

vigneshshanmugam:
+ check this https://github.com/npm/rfcs/blob/latest/implemented/0010-monorepo-subdirectory-declaration.md for explanation.

### DIFF
--- a/packages/rum-core/package.json
+++ b/packages/rum-core/package.json
@@ -6,7 +6,7 @@
   "main": "src/index.js",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/elastic/apm-agent-js-base.git",
+    "url": "https://github.com/elastic/apm-agent-js-base.git",
     "directory": "packages/rum-core"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


packages/rum-core/package.json
```diff
@@ -4,7 +4,11 @@
   "description": "Elastic apm core",
   "license": "MIT",
   "main": "src/index.js",
-  "repository": "https://github.com/elastic/apm-agent-js-base/tree/master/packages/rum-core",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/elastic/apm-agent-js-base.git",
```
jahtalab: Isn't just using `https://...` simpler?

vigneshshanmugam: yeah it is, Its just a preference that it is indeed a git url. I am fine with modifying it as well.

jahtalab: Yeah, let's just use the `https:`

PR #187, opened 2019-02-18
chore: consolidate karma and webpack configs across packages

vigneshshanmugam:
+ elastic/apm-agent-js-base#170 - Creating single tunnel for same account is not possible, So this PR tries to focus n starting tunnels to faster the test. We use the same tunnel identifier to reduce the time to start a new sauce connect
+ Did some heavy refactoring around webpack stats getting generated around the test logs 
+ Karma configs are better organised across the project and also reduces the confusion from global configs.
+ Global config for testing is at a single place, No need to do`window.globalConfigs` everywhere. It was super confusing to run different tests and also change urls at multiple files to make tests pass in local. 

**Review per commit** for better understanding. I have rebased it based on functionality to make it easier for review.


dev-utils/run-script.js
```diff
@@ -1,71 +0,0 @@
-/**
- * MIT License
```
jahtalab: I think these scripts make the running on the local machine easier!

dev-utils/test.js
```diff
@@ -225,36 +230,12 @@ function runE2eTests(configFilePath, runSelenium) {
...
-    env
-  }
-  // if (env.sauceLabs) {
-  //   config.useMocks = true
-  // }
-  return config
-}
-
 module.exports = {
-  runUnitTests,
```
jahtalab: I think these scripts also give us more control over how to run the tests!

packages/rum/package.json
```diff
@@ -15,7 +15,7 @@
     "karma": "karma start",
     "karma-coverage": "karma start --coverage",
     "serve": "npm run script serveE2e ./ 8000",
-    "runUnitTests": "npm run script runUnitTests",
+    "runUnitTests": "karma start --single-run",
```
jahtalab: Previously we did this with `karma start` but IMO using the script that we wrote gives us more control!

dev-utils/test.js
vigneshshanmugam: With the launchSaunceconnect manul option removed. this script would simply be an alias to runKarma which we had before. It doesn't really feel needed IMO.

packages/rum/package.json
vigneshshanmugam: This still uses the config from `karma.config.js` so I believe we control everything from there. The script we had effectively did `singleRun: true`. We were not using any custom functionality so I felt its better to remove it altogether.

jahtalab: Let's keep using our own scripts, In most cases making a small changes is more easier for us than with using original script!

packages/rum/test/e2e/integration-test.js
```diff
@@ -28,31 +28,30 @@ const puppeteer = require('puppeteer')
 async function runIntegrationTest(pageUrl) {
   const browser = await puppeteer.launch({
     headless: true,
-    args: [
-      '--no-sandbox',
-      '--disable-setuid-sandbox',
-      '--remote-debugging-port=9222'
```
jahtalab: Don't we need the remote debugging port?

vigneshshanmugam: I don't think so. With the option enabled the tests hang in local so i removed it in favor of enabling test logs.

dev-utils/test-utils.js
```diff
@@ -25,48 +25,8 @@
 
 const path = require('path')
 const fs = require('fs')
-const karmaUtils = require('./karma')
-const saucelabsUtils = require('./saucelabs')
-
-function runKarma(configFile) {
```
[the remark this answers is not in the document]
vigneshshanmugam: I dont have any strong preference here, I will add them back.

dev-utils/test-config.js
```diff
@@ -23,38 +23,55 @@
...
-function launchSauceConnect(userConfig, done) {
-  var sauceConnectLauncher = require('sauce-connect-launcher')
-
-  var config = {
-    username: userConfig && (userConfig.username || process.env.SAUCE_USERNAME),
-    accessKey:
-      userConfig && (userConfig.accessKey || process.env.SAUCE_ACCESS_KEY),
+function getSauceConnectOptions() {
+  return {
+    username: process.env.SAUCE_USERNAME,
```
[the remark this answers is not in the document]
vigneshshanmugam: Sure makes sense, but we might be sending lots of unwanted keys to sauceConnect options. thats why separated and kept them in same file `test-config.js`

dev-utils/karma.js
```diff
@@ -109,28 +118,18 @@ var baseConfig = {
           test: /\.js$/,
           loader: 'babel-loader',
           options: {
-            presets: [
-              [
-                '@babel/preset-env',
-                {
-                  targets: {
-                    ie: '10'
```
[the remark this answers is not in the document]
vigneshshanmugam: No just using babel.config.js.. removing duplicates

dev-utils/test-config.js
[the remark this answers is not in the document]
jahtalab: Ok, I think that's fine then!

dev-utils/run-script.js
```diff
@@ -23,29 +23,18 @@
...
 
-function runUnitTests(launchSauceConnect, directory) {
-  var testConfig = testUtils.getTestEnvironmentVariables()
-  testConfig.karmaConfigFile = path.join(
-    __dirname,
-    '..',
-    directory,
-    './karma.conf.js'
-  )
-  if (launchSauceConnect !== 'false') {
```
jahtalab: The `launchSauceConnect`  is useful when debugging issues with saucelabs, you can start the sauceconnect individually in another terminal and don't have to start it on each run!

dev-utils/run-script.js
```diff
@@ -57,15 +46,9 @@ function runScript(scripts, scriptName, scriptArgs) {
   }
 }
 
-var scripts = {
+const scripts = {
   runUnitTests,
-  generateNotice,
-  launchSauceConnect: function launchSauceConnect() {
```
jahtalab: `launchSauceConnect ` is useful for debugging saucelabs issues locally, see my comment above!

dev-utils/saucelabs.js
```diff
@@ -1,60 +0,0 @@
...
-
-  var config = {
-    username: userConfig && (userConfig.username || process.env.SAUCE_USERNAME),
-    accessKey:
-      userConfig && (userConfig.accessKey || process.env.SAUCE_ACCESS_KEY),
-    logger: console.log,
-    noSslBumpDomains: 'all'
-  }
-
-  var tryConnect = function(maxAttempts, currAttempts, done) {
```
jahtalab: We should use this script to `launchSauceConnect` the logic here make sure to attempt three times if the first try fails!

packages/rum-core/package.json
```diff
@@ -13,7 +13,7 @@
     "karma": "karma start",
     "karma-bench": "karma start karma.bench.conf.js",
     "karma-coverage": "karma start --coverage",
-    "runUnitTests": "node ../../dev-utils/run-script.js runUnitTests true packages/rum-core",
+    "runUnitTests": "node ../../dev-utils/run-script.js runUnitTests packages/rum-core",
```
jahtalab: The `true` here is for the `launchSauceConnect` parameter mentioned above!

packages/rum/run-script.js
```diff
@@ -24,20 +24,14 @@
...
-const testUtils = require('../../dev-utils/test')
+const JasmineRunner = require('jasmine')
+const express = require('express')
+const serveIndex = require('serve-index')
+const testUtils = require('../../dev-utils/test-utils')
 const { runIntegrationTest } = require('./test/e2e/integration-test')
-const projectDirectory = path.join(__dirname, './')
 const { generateNotice } = require('../../dev-utils/dep-info')
 
-function runUnitTests(launchSauceConnect) {
```
jahtalab: We can consolidate this with the one defined above!

packages/rum/test/e2e/general-usecase/app.e2e-spec.js
```diff
@@ -111,8 +110,6 @@ describe('general-usercase', function() {
     })
     expect(span).toBeDefined()
 
-    if (isChrome()) {
```
jahtalab: We don't need the `isChrome` anymore?

packages/rum-core/package.json
vigneshshanmugam: Yes i am aware, We don't use launchSauceConnect function anymore.. Handed it via karma-sauce-launcher.

dev-utils/saucelabs.js
vigneshshanmugam: Retry strategy is already handled here https://github.com/elastic/apm-agent-js-base/blob/982846b9d351ea5fd99800295da14a9e7e581f49/dev-utils/test-config.js#L34

packages/rum/test/e2e/general-usecase/app.e2e-spec.js
vigneshshanmugam: Handled in webdriver.js itself.

dev-utils/run-script.js
vigneshshanmugam: To be honest, you can do this via switch of a flag `sauceConnect` (true|false) in code since you are already debugging.

jahtalab: IMO, passing it through the command line is easier since we don't have to change the code!

vigneshshanmugam: I have reenabled the function as per your suggestion.

jahtalab:
Seems like the tests are failing because of the SauceConnect issue!

vigneshshanmugam:
@jahtalab Yeah was missing tunnelIdenfitier in karma launcher. fixed now.

dev-utils/karma.js
```diff
@@ -23,9 +23,13 @@
...
  */
 
-// dependencies
-// npm i --save-dev jasmine karma karma-sauce-launcher karma-failed-reporter karma-jasmine karma-spec-reporter webpack karma-webpack karma-chrome-launcher karma-sourcemap-loader babel-core babel-loader babel-preset-es2015 babel-plugin-istanbul
-var baseLaunchers = {
+const { Server } = require('karma')
+const { EnvironmentPlugin } = require('webpack')
+const { getWebpackEnv, getSauceConnectOptions } = require('./test-config')
+
+const BABEL_CONFIG_FILE = require.resolve('elastic-apm-js-base/babel.config.js')
```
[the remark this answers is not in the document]
vigneshshanmugam: IMO it doesn't belong in the root and `babel.config.js` is pretty specific to how we bundle the whole app. We configure the individual configs through `babelrc.js` like for e2e, we can have different things for different tests.

dev-utils/webdriver.js
```diff
@@ -32,10 +32,22 @@ var logLevels = {
...
+}
+
 function assertNoBrowserErrors(whitelist) {
   return new Promise((resolve, reject) => {
+    /**
+     * browser.log API is available only in chrome
+     */
+    if (!isChrome()) {
+      resolve()
+    }
```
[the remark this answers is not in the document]
vigneshshanmugam: good catch, fixing it

packages/rum/test/e2e/general-usecase/app.js
```diff
@@ -23,12 +23,12 @@
...
-var createApmBase = require('../e2e')
-
-var active = Math.random() < 1
-var serverUrl = 'http://localhost:8003'
+const createApmBase = require('../e2e')
+const { createTracer } = require('../../../src/opentracing')
 
-var elasticApm = createApmBase({
+const active = Math.random() < 1
+const serverUrl = 'http://localhost:8003'
```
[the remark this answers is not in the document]
vigneshshanmugam: I left it because in other places, We are having in .html. I will set up some bundling config for standalone test and fix this.

dev-utils/karma.js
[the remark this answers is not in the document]
jahtalab: I agree, since it is specific we should have a parameter here to accept the config path, right?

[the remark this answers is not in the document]
vigneshshanmugam: We don't need to, babel takes care of that for us. It recursively finds the folder level when applying the config.

vigneshshanmugam:
@jahtalab Please review the last commit where i extracted the test-utils and also handled the config for mockBackEnd url as suggested.

PR #188, opened 2019-02-18
chore: add readmes for individual packages

jahtalab:


packages/rum-core/README.md
```diff
@@ -1 +1,5 @@
-TODO: Provide description.
+# Elastic APM JavaScript core (for development)
+
+This is the core JavaScript module for Elastic APM. 
+
```
vigneshshanmugam: we can include usage section for both packages. like `npm install <package-name>` and few info from here https://www.npmjs.com/package/elastic-apm-js-core

jahtalab: I've referenced our contributing guide!

PR #194, opened 2019-02-25
feat(rum-core): Add task API

jahtalab:
closes elastic/apm-agent-js-base#101

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -589,9 +589,9 @@
       }
     },
     "@babel/plugin-transform-destructuring": {
-      "version": "7.2.0",
-      "resolved": "https://registry.npmjs.org/@babel/plugin-transform-destructuring/-/plugin-transform-destructuring-7.2.0.tgz",
-      "integrity": "sha512-coVO2Ayv7g0qdDbrNiadE4bU7lvCd9H539m2gMknyVjjMdwF/iCOM7R+E8PkntoqLkltO0rk+3axhpp/0v68VQ==",
+      "version": "7.3.2",
+      "resolved": "https://registry.npmjs.org/@babel/plugin-transform-destructuring/-/plugin-transform-destructuring-7.3.2.tgz",
+      "integrity": "sha512-Lrj/u53Ufqxl/sGxyjsJ2XNtNuEjDyjpqdhMNh5aZ+XFOdThL46KBj27Uem4ggoezSYBxKWAil6Hu8HtwqesYw==",
       "dev": true,
       "requires": {
         "@babel/helper-plugin-utils": "^7.0.0"
@@ -918,6 +918,24 @@
         }
       }
     },
+    "@babel/polyfill": {
+      "version": "7.2.5",
+      "resolved": "https://registry.npmjs.org/@babel/polyfill/-/polyfill-7.2.5.tgz",
+      "integrity": "sha512-8Y/t3MWThtMLYr0YNC/Q76tqN1w30+b0uQMeFUYauG2UGTR19zyUtFrAzT23zNtBxPp+LbE5E/nwV/q/r3y6ug==",
+      "dev": true,
+      "requires": {
+        "core-js": "^2.5.7",
+        "regenerator-runtime": "^0.12.0"
+      },
+      "dependencies": {
+        "regenerator-runtime": {
+          "version": "0.12.1",
+          "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.12.1.tgz",
+          "integrity": "sha512-odxIc1/vDlo4iZcfXqRYFj0vpXFNoGdKMAUieAlFYO6m/nl5e9KR/beGf41z4a1FI+aQgtjhuaSlDxQ0hmkrHg==",
+          "dev": true
+        }
+      }
+    },
     "@babel/preset-env": {
       "version": "7.2.3",
       "resolved": "https://registry.npmjs.org/@babel/preset-env/-/preset-env-7.2.3.tgz",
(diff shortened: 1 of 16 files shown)


packages/rum/test/e2e/react/index.html
```diff
@@ -2,11 +2,12 @@
 <html>
 
 <head>
-  <title>test</title>
+  <title>React test</title>
   <base href="/test/e2e/react/">
-  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; connect-src http://localhost:8000 http://localhost:8200 https://frontend-test-apm-server.elastic.co">
+  <meta http-equiv="Content-Security-Policy" content="default-src 'self';img-src 'self' data:; connect-src http://localhost:8000 http://localhost:8200 http://localhost:8001">
```
vigneshshanmugam: lets remove this line. Feels leeky. If we want to do this right, then we have to inject it from server with the variables configured for server URL. This is the reason why I was standardising all the test configs in my PR #187

packages/rum-core/src/common/patching/fetch-patch.js
```diff
@@ -94,13 +94,18 @@ function patchFetch(callback) {
       promise.then(
         function(response) {
           resolve(response)
-          task.data.response = response
-          invokeTask(task)
+          // invokeTask in the next execution cycle to let the promise resolution complete
+          setTimeout(() => {
```
vigneshshanmugam: I am not sure if this would be the right approach. Any reason why you went with setTimeout instead of Promise.resolve?

vigneshshanmugam: After reading through a bit, promise.resolve is more appropriate here than setTimeout and promise.resolve also guarantees that our next function would be in the order after the previous resolution happened since both of the resolutions will be part of microTask queue.

packages/rum/test/e2e/react/webpack.config.js
```diff
@@ -37,9 +42,23 @@ module.exports = {
...
       {
         test: /.jsx?$/,
-        include: [__dirname, path.resolve(__dirname)],
         use: {
-          loader: 'babel-loader'
+          loader: 'babel-loader',
+          options: {
+            presets: [
+              [
+                '@babel/preset-env',
```
vigneshshanmugam: does it work without preset-react?

vigneshshanmugam: You dont need to pass these explicity here. We have babel.config.js for test directory which has these options by default.

packages/rum/test/e2e/react/index.html
jahtalab: We need to make sure that the agent works when A CSP is present in the page! so this is not for  adding A CSP to secure our tests, it is for testing if the agent works properly if the CSP is in the page! Let's keep these until we have added it from the server. Should be a separate PR IMO.

packages/rum-core/src/common/patching/fetch-patch.js
jahtalab: Yeah, It was my temp solution will use promise.resolve.

packages/rum/test/e2e/react/index.html
vigneshshanmugam: Sure. makes sense

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -80,6 +80,8 @@ class PerformanceMonitoring {
           var spanName =
             task.data.method + ' ' + stripQueryStringFromUrl(task.data.url)
           var span = transactionService.startSpan(spanName, 'external.http')
+          var taskId = transactionService.addTask()
```
roncohen: does this mean users will have to create a task every time they add a span?

roncohen: if yes, how about handling that behind the scenes for users? If you _really_ want to end the transaction, you should end all the spans?

jahtalab: No the spans would work as they are today. i.e we don't wait for the spans to finish and if the transaction is finished we truncate the running spans.  

The users would have to add tasks for any async operation they want to wait for before ending the transaction (whether there's a span for it or not). The most common async operation are XHRs which the agent automatically adds the task for them. Going forward we will also add automatic tasks for async operations we think are useful.

We did discuss adding a parameter to the `startSpan` method to also create a task for it, but I think we should do it as future improvements.

roncohen: OK. In which situations would you like to add a span, but want to finish the transaction before the span ends? And could you, in those situations, just call `transaction.end()` instead? 

I also went looking for the motivation to add the tasks API, but didn't see anything in the original issue or here (as a side node, the description for this PR is "Add e2e tests for react app"). This PR also doesn't referer to the original issue, so people have to go hunting to find it. 

Is the problem that people who create custom spans have no way to control the transaction duration today because the TaskAPI isn't public? In that case, it seems more appropriate to add a task whenever users add a span.

jahtalab: Thanks @roncohen for pointing that out! I've update the issue description and linked to it from the PR!

Please make comment on the original issue if something needs clarifying!

package-lock.json
```diff
@@ -1,14986 +0,0 @@
-{
-  "name": "elastic-apm-rum",
```
vigneshshanmugam: Deleted by mistake?

packages/rum-core/test/common/fetch-patch.spec.js
```diff
@@ -56,8 +56,10 @@ describe('xhrPatch', function() {
       expect(events.map(e => e.event)).toEqual(['schedule'])
       promise.then(function(resp) {
         expect(resp).toBeDefined()
-        expect(events.map(e => e.event)).toEqual(['schedule', 'invoke'])
-        done()
+        Promise.resolve().then(function() {
```
vigneshshanmugam: Do we need this still even after doing Promise.resolve instead of setTimeout?

packages/rum/test/e2e/react/webpack.config.js
```diff
@@ -37,9 +42,11 @@ module.exports = {
     rules: [
       {
         test: /.jsx?$/,
-        include: [__dirname, path.resolve(__dirname)],
         use: {
-          loader: 'babel-loader'
+          loader: 'babel-loader',
+          options: {
+            plugins: ['@babel/plugin-transform-destructuring']
```
vigneshshanmugam: Plugin is not required. It is part of preset-env by default. https://github.com/babel/babel/blob/master/packages/babel-preset-env/src/available-plugins.js#L14

package-lock.json
jahtalab: No! the package-lock.json was making the tests fail see: https://travis-ci.org/elastic/apm-agent-js-base/jobs/498791322#L508


I will add it back once the tests are green for the original changes. But I think we need to reconsider our use of package-lock.json

packages/rum-core/test/common/fetch-patch.spec.js
jahtalab: Yes, we actually need this because of the Promise.resolve in the fetch patch, since at this point the invoke task is not called yet.

packages/rum/test/e2e/react/webpack.config.js
jahtalab: The tests on Safari 9 fail without these because of bundling with unsupported ES6 features.

roncohen:
Thanks for elaborating on motivation for the Task API @jahtalab.

I'll ask here, because it relates to the implementation of the task API:

It seems counterintuitive to me that a span wouldn't also add a task. E.g. whenever you create a span you'll also need to remember to add a task. 

In which situations would you like to add a span, but want to finish the transaction before the span ends? And could you, in those situations, just call `transaction.end()` instead?

packages/rum/test/e2e/react/webpack.config.js
vigneshshanmugam: Weird, because we do transform them to Es5 along with the plugins using our `.babelrc.js` in tests so I am not sure bundling would be the reason.

package-lock.json
vigneshshanmugam: seems like a cache issue to me.

jahtalab:
@roncohen ,

Let's first consider how transactions work today. Spans are added automatically or manually to the current transactions and at some point we (or the user) call the `transaction.end`, which means **today we don't wait for spans to finish** and truncate any active spans when the `transaction.end` is called.

The task API is an improvement to how things work today. It gives us more control over when to end a transaction. We still don't wait for spans to finish and **users are not expected to add tasks whenever they start a span**, this part works as it was before. 

Furthermore consider this scenario, there's a recurring job in the page, for example a WebSocket that continuously receives status updates, even though the user wants to measure the operation related to an status update, they don't want to wait for it to finish. Even though it's possible to use `transaction.end` in this case, as I mentioned in the original issue, the motivation for the task API is that users don't always know at which point to end a transaction.

Finally, the task API will most likely be used for our internals and maybe only advanced users if they need to.

package-lock.json
jahtalab: I also removed the cache, but it kept happening! It's a weird issue!

packages/rum/test/e2e/react/webpack.config.js
jahtalab: The bundle actually was not transformed to ES5! The issue was with the generated bundle for this.

vigneshshanmugam: Its still failing now right? Do you have any error related to this? Seems like a bundling problem if this is the case.

roncohen:
> consider this scenario, there's a recurring job in the page, for example a WebSocket that continuously receives status updates, even though the user wants to measure the operation related to an status update, they don't want to wait for it to finish.

In this case, wouldn't be better to add a boolean argument to "createSpan" (that defaults to true) that lets users decide if the transaction should wait for a span created for the WebSocket to finish? 

Taking this a step further (just an idea): since "tasks" are a completely new concept that we've invented, could we instead of exposing that API, ask users to create a "blocking" span (as described above) if they want to keep the transaction open? e.g. while the route change is ongoing, they just create a span and end it whenever they are done with that part. It will show up on the timeline, but it's probably better than if there's a hole in the timeline and they might wonder what the app was doing during that time? Seems like it would be a vastly simpler explanation.

jahtalab:
We can certainly add a boolean parameter to the `startSpan` method, I've added it to future improvements in the issue. I don't think the it's value should be true by default though. I like the idea of calling this boolean parameter `blocking` or something similar, I agree it's simpler to explain. 
Let's reiterate on this on a separate PR.

packages/rum/test/e2e/react/webpack.config.js
jahtalab: The tests are passing now! 
The errors were happing in the Safari 9 on saucelabs. The bundles were still in ES6 code before this change.

vigneshshanmugam: Can you move this plugin alone to `test/e2e/.babelrc.js`, To avoid the duplication of babel configs.

vigneshshanmugam:
@jahtalab Lets merge once tests passes.

PR #198, opened 2019-03-04
feat(rum-core): add service env to metadata payload

### LINKED ISSUE #151: Add service.runtime to the payload
In order to be consistent with the data model we should provide a value for `service.runtime`, further more this value could be used for a visualisation in UI. Could use a constant string like `Browser` for the value.

vigneshshanmugam:
+ fixes elastic/apm-agent-js-base#48 
+ Users can add an option to override the environment where the service being monitored is deployed. Add the doc for the same.

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -71,11 +71,7 @@ class ApmServer {
       language: {
         name: 'javascript'
       },
-      environment: sanitizeString(cfg.get('environment'), stringLimit),
-      runtime: {
-        name: 'browser',
-        version: ''
-      }
+      environment: sanitizeString(cfg.get('environment'), stringLimit)
     }
     return serviceObject
   }
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -58,18 +58,23 @@ class ApmServer {
   }
 
   createServiceObject() {
-    var cfg = this._configService
-    var stringLimit = cfg.get('serverStringLimit')
+    const cfg = this._configService
+    const stringLimit = cfg.get('serverStringLimit')
 
-    var serviceObject = {
+    const serviceObject = {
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/common/apm-server.js
```diff
@@ -58,18 +58,23 @@ class ApmServer {
...
+        name: sanitizeString(cfg.get('agentName'), stringLimit),
+        version: sanitizeString(cfg.get('agentVersion'), stringLimit)
       },
       language: {
         name: 'javascript'
+      },
+      environment: sanitizeString(cfg.get('environment'), stringLimit),
+      runtime: {
+        name: 'browser',
+        version: ''
```
jahtalab: It seems strange to send an empty string for version. I know it's required but maybe runtime is not a good fit for the RUM agent either. 

I think we should postpone adding runtime until kibana starts using it. WDYT?

vigneshshanmugam: Kibana 6.6 already uses this value. This is addressed in next PR with 'N/A'.  I would just send it since  we are touching the metadata anyway.

jahtalab: Sorry, I meant Kibana doesn't uses it for the RUM agent.

vigneshshanmugam: Cool, but that shouldn't stop us from reporting N/A right? Since we are already working on the metadata stuff, I would just keep it as N/A since we would need to do it anyway in future.

jahtalab: I'm not sure if we would need to do it in the future, since it's not used in the UI for the RUM agent we might never actually need it and if that's the case we just have added something to the payload that will never be used.

vigneshshanmugam: Kibana 6.6. UI, 
screenshot 2019-03-05 at 16 25 39

vigneshshanmugam: Removed runtime from the payload.

PR #208, opened 2019-03-12
chore(rum): unhandled promise rejection in integration test

vigneshshanmugam:
+ When the Integration tests timeout, we get unhandled promise rejection warnings, This PR fixes that and instead logs error to stdout.

### DIFF
--- a/packages/rum/test/e2e/integration-test.js
+++ b/packages/rum/test/e2e/integration-test.js
@@ -75,6 +75,7 @@ async function runIntegrationTest(pageUrl) {
     }
   } catch (err) {
     console.log('[Intgration Tests Error]', err)
+    throw err
   } finally {
     await browser.close()
   }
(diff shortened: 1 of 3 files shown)


vigneshshanmugam:
@jahtalab Agree with you, renamed the commit.

packages/rum/test/e2e/integration-test.js
```diff
@@ -73,6 +73,8 @@ async function runIntegrationTest(pageUrl) {
     result.response = {
       status: transactionResponse.status()
     }
+  } catch (err) {
+    console.log('[Intgration Tests Error]', err)
```
jahtalab: Show re-throw the error as well!

vigneshshanmugam: I dont think so, we will are back to the same unhanded rejection and would break the flow again in here https://github.com/elastic/apm-agent-rum-js/blob/87bbf600c84f8832a79e78457d43bafc0d1a6f7b/packages/rum/run-script.js#L90-L97.

We should catch and log the error and simply send a empty result which should break the tests.

vigneshshanmugam: I have done a proper fix now according to how our integration test env requires and also fixed the timeout caused by not calling done

PR #212, opened 2019-03-12
chore: fix e2e tests and migrate to webdriver 5

### LINKED ISSUE #209: Failsafe tests log failure in the console but the tests pass
It seems that failsafe test failures doesn't make the whole test to fail, see [Travis logs](https://travis-ci.org/elastic/apm-agent-rum-js/jobs/505145111#L835).

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#209 
+ Migrated to webdriver 5 which is has full W3C WebDriver compliancy


dev-utils/test-utils.js
```diff
@@ -183,7 +183,7 @@ function runKarma(configFile) {
 }
 
 function runE2eTests(configFilePath, runSelenium) {
-  const wdio = new Launcher(configFilePath)
+  const wdio = new Launcher(configFilePath, {})
```
vigneshshanmugam: required because of the new CLI update, otherwise breaks.

vigneshshanmugam:
@jahtalab Made it work after several attempts.. good for review now

packages/rum/run-script.js
```diff
@@ -139,37 +139,37 @@ function runSauceTests(serve = 'true') {
...
-        console.log('Sauce Tests Failed', err)
-      })
-      .then(() => {
-        servers.map(s => s.close())
-        process.exit(0)
+  testUtils.runSauceConnect(sauceConnectOpts, async sauceConnectProcess => {
+    try {
+      await runAll('test:unit', loggerOpts)
+      await runAll(['test:e2e:supported', 'test:e2e:failsafe'], {
+        parallel: true,
```
jahtalab: Will this make the saucelab to timeout more often, since we could surpass the concurrency limit?

vigneshshanmugam: No it does not, I was trying to run unit tests parallel that was the issue i mentioned yesterday.

jahtalab: But this still runs 'test:e2e:supported' and 'test:e2e:failsafe' in parallel right?

vigneshshanmugam: yes this is still run in parallel and does no harm to the timeouts  as far i tested it.. If we include deprecate supportfor more browsers in future, we might hit the limit and would have to run it in series for the time being.

PR #213, opened 2019-03-13
chore: group lint and bundle size under check stage

jahtalab:
Grouping the tests under one stage would take advantage of our concurrency setting.

Also, this allows the bundle size check to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,22 +19,23 @@ env:
     - MODE=saucelabs
 
 stages:
-  - lint
+  - check
   - test
-  - bundlesize
   - docs
 
 jobs:
+  allow_failures:
+    - script: npm run bundlesize
   include:
     - stage: docs
       script: npm run build-docs
       name: 'Build Docs'
-    - stage: bundlesize
-      script: npm run bundlesize
-      name: 'Analyse size of JS bundles'
-    - stage: lint
+    - stage: check
       script: npm run lint
       name: 'Lint'
+    - stage: check
+      script: npm run bundlesize
+      name: 'Analyse size of JS bundles'
     - stage: test
       env:
         - STACK_VERSION=6.5.0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,22 +19,19 @@ env:
...
 
 stages:
-  - lint
+  - check
   - test
-  - bundlesize
-  - docs
 
 jobs:
+  allow_failures:
```
vigneshshanmugam: Since you are already touching this, Please disable the bundle size check on master build. It makes only sense on Pull requests and raises red flag before merging to master.

.travis.yml
```diff
@@ -59,7 +56,9 @@ jobs:
       env:
         - STACK_VERSION=7.0.0-alpha1-SNAPSHOT
         - SCOPE=@elastic/apm-rum
-
+    - stage: test
```
vigneshshanmugam: any reason why we are running docs as part of test? It's better to keep it as separate stage IMO.

jahtalab: Travis won't start the next stage until the previous stage is finished. With our concurrency level having two stages improves overall speed.

vigneshshanmugam: Got it. makes total sense to do this.

.travis.yml
jahtalab: With `allow_failures` it won't make the whole test to fail only one job will fail status will be green including in master.

vigneshshanmugam: Yeah i understand, But my point is to not even run the bundle size check on master since it does not help us in any way.

vigneshshanmugam: We can do disable in another PR.

jahtalab: I think it's ok to have it on master as well. Currently we can't make decisions regarding the bundle size per PR (so far we only increase the bundle limit). We should have periodic audits of the bundle and having this check on master gives us the current status of this check.

PR #216, opened 2019-03-15
feat(rum): better log message on invalid configuration

### LINKED ISSUE #86: Validate config options before setting them
We should validate the value provided by the user for config options according the defined type and restrictions (varies between config options).

Some config options that need validation:
* `serviceName`
* `serverUrl`
* `pageLoadTraceId`
* `pageLoadSpanId`
* `pageLoadSampled`
* `pageLoadTransactionName`

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#86 #248 #247 
+ Logs error on invalid configs
+ Improved the error message to help the users find problem
+ Included the fix for removing trailing slash on serverUrl

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -33,12 +33,6 @@ class ApmServer {
   constructor(configService, loggingService) {
     this._configService = configService
     this._loggingService = loggingService
-    this.logMessages = {
-      invalidConfig: {
-        message: 'RUM agent configuration is invalid!',
-        level: 'warn'
-      }
-    }
 
     this.errorQueue = undefined
     this.transactionQueue = undefined
@@ -194,15 +188,6 @@ class ApmServer {
     this.throttleAddTransaction(transaction)
   }
 
-  warnOnce(logObject) {
-    if (logObject.level === 'warn') {
-      logObject.level = 'debug'
-      this._loggingService.warn(logObject.message)
-    } else {
-      this._loggingService.debug(logObject.message)
-    }
-  }
-
   ndjsonErrors(errors) {
     return errors.map(function(error) {
       return NDJSON.stringify({ error })
@@ -227,10 +212,6 @@ class ApmServer {
   }
 
   _send(data = [], type = 'transaction') {
-    if (!this._configService.isValid()) {
-      this.warnOnce(this.logMessages.invalidConfig)
-      return
-    }
     if (data.length === 0) {
       return
     }
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -154,12 +154,6 @@ class Config {
   }
 
   getEndpointUrl() {
-    /**
-     * Remove all trailing slash for serverUrl since serverUrlPrefix
-     * includes a forward slash for the path
-     */
-    this.config.serverUrl = this.config.serverUrl.replace(/\/+$/, '')
-
     return this.config.serverUrl + this.config.serverUrlPrefix
   }
 
@@ -214,6 +208,16 @@ class Config {
   }
 
   setConfig(properties = {}) {
+    /**
+     * Normalize config
+     *
+     * Remove all trailing slash for serverUrl since serverUrlPrefix
+     * includes a forward slash for the path
+     */
+    if (properties.serverUrl) {
+      properties.serverUrl = properties.serverUrl.replace(/\/+$/, '')
+    }
+
     this.config = merge({}, this.defaults, this.config, properties)
     this._changeSubscription.applyAll(this, [this.config])
   }
@@ -223,26 +227,28 @@ class Config {
   }
 
   /**
-   * Validate the config once set via setConfig aganist the required parameters
-   * Also normalizes the config on specific keys
+   * Validate the config aganist the required parameters and
+   * generates error messages
    */
-  validate() {
+  validate(properties = {}) {
     const requiredKeys = ['serviceName', 'serverUrl']
     const errors = []
 
-    for (let i = 0; i < requiredKeys.length; i++) {
-      const key = requiredKeys[i]
-      if (this.config[key] == null || this.config[key] === '') {
+    Object.keys(properties).forEach(key => {
+      if (requiredKeys.indexOf(key) !== -1 && !properties[key]) {
         errors.push(errors.length === 0 ? 'Missing ' + key : key)
       }
-    }
+    })
     /**
      * Invalid characters on serviceName
      */
-    if (!/^[a-zA-Z0-9 _-]+$/.test(this.config.serviceName)) {
+    if (
+      properties.serviceName &&
+      !/^[a-zA-Z0-9 _-]+$/.test(properties.serviceName)
+    ) {
       errors.push(
         'serviceName ' +
-          this.config.serviceName +
+          properties.serviceName +
           ' contains invalid characters! (allowed: a-z, A-Z, 0-9, _, -, <space>)'
       )
     }
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -154,8 +154,13 @@ class Config {
   }
 
   getEndpointUrl() {
-    var url = this.get('serverUrl') + this.get('serverUrlPrefix')
-    return url
+    /**
+     * Remove all trailing slash for serverUrl since serverUrlPrefix
+     * includes a forward slash for the path
+     */
+    this.config.serverUrl = this.config.serverUrl.replace(/\/+$/, '')
+
+    return this.config.serverUrl + this.config.serverUrlPrefix
   }
 
   set(key, value) {
@@ -240,14 +245,6 @@ class Config {
           this.config.serviceName +
           ' contains invalid characters! (allowed: a-z, A-Z, 0-9, _, -, <space>)'
       )
-    }
-    /**
-     * Remove all trailing slash for serverUrl since serverUrlPrefix
-     * includes a forward slash for the path
-     */
-    const serverUrl = this.config.serverUrl
-    if (serverUrl) {
-      this.config.serverUrl = serverUrl.replace(/\/+$/, '')
     }
 
     return errors
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -228,29 +228,34 @@ class Config {
 
   /**
    * Validate the config aganist the required parameters and
-   * generates error messages
+   * generates error messages with missing and invalid keys
    */
   validate(properties = {}) {
     const requiredKeys = ['serviceName', 'serverUrl']
-    const errors = []
-
+    const errors = {
+      missing: [],
+      invalid: []
+    }
+    /**
+     * Check when required keys are missing
+     */
     Object.keys(properties).forEach(key => {
       if (requiredKeys.indexOf(key) !== -1 && !properties[key]) {
-        errors.push(errors.length === 0 ? 'Missing ' + key : key)
+        errors.missing.push(key)
       }
     })
     /**
-     * Invalid characters on serviceName
+     * Invalid values on the config
      */
     if (
       properties.serviceName &&
       !/^[a-zA-Z0-9 _-]+$/.test(properties.serviceName)
     ) {
-      errors.push(
-        'serviceName ' +
-          properties.serviceName +
-          ' contains invalid characters! (allowed: a-z, A-Z, 0-9, _, -, <space>)'
-      )
+      errors.invalid.push({
+        key: 'serviceName',
+        value: properties.serviceName,
+        allowed: 'a-z, A-Z, 0-9, _, -, <space>'
+      })
     }
 
     return errors
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -217,17 +217,40 @@ class Config {
     return this._changeSubscription.subscribe(fn)
   }
 
-  isValid() {
+  /**
+   * Validate the config once set via setConfig aganist the required parameters
+   * Also normalizes the config on specific keys
+   */
+  validate() {
     const requiredKeys = ['serviceName', 'serverUrl']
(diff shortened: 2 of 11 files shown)


packages/rum-core/src/common/config-service.js
```diff
@@ -66,52 +72,53 @@ class Config {
   constructor() {
     this.config = {}
     this.defaults = {
-      serviceName: '',
-      serviceVersion: '',
-      environment: '',
+      // String Opts
```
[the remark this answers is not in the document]
vigneshshanmugam: I did the other way around because if we decided to remove config service class and use a POJO for `config.js`, It would be ideally easier to have normalize for each options like `String`, `Boolean` and so on.. which makes our refactoring easier in future.

packages/rum-core/src/common/apm-server.js
```diff
@@ -178,30 +178,17 @@ class ApmServer {
   }
 
   addError(error) {
-    if (this._configService.isActive()) {
```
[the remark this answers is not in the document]
vigneshshanmugam: We talked about this one in offline, so marking it as resolved.

vigneshshanmugam:
The reason why i went with changing both at the same time is due to the fact the validation logic is now different and we test all the configs instead of the two keys that we test before and we also need to normalize and validate at the same go because of using defaults. I dont think we would gain much benefit by moving in to two PR's. But lets dicuss about this on monday and come to a conclusion.

jahtalab:
The reason I suggested that is because I think config validation should be considered a breaking change while the no-op is only related to the internal workings of the agent.

vigneshshanmugam:
Config validation - Goes in major version
Noop agent - goes in to minor.

vigneshshanmugam:
Noop agent is covered in #239 PR. I will modify this PR to error on invalid configs which would go in next breaking change.

packages/rum/src/apm-base.js
```diff
@@ -99,9 +99,33 @@ class ApmBase {
...
+      return
+    }
+    const loggingService = this.serviceFactory.getService('LoggingService')
+    const message = 'RUM Agent configuration is invalid: ' + errors.join(', ')
+
+    /**
+     * TODO: Deactivate the agent when configuration is invalid
+     * on next major version
+     */
+    // configService.setConfig({ active: false })
```
vigneshshanmugam: We can do this on next breaking change to align with other agents.

vigneshshanmugam:
@jahtalab Inorder to make this as enhancement, I have removed the part where we set the agent to inactive mode for invalid configs. Instead we log proper warnings. I have commented out that part which we can include in next breaking changes. So far, this PR is good to go on next minor release.

packages/rum-core/src/common/apm-server.js
```diff
@@ -227,10 +212,6 @@ class ApmServer {
   }
 
   _send(data = [], type = 'transaction') {
-    if (!this._configService.isValid()) {
```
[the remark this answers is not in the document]
vigneshshanmugam: We already do it in apm-base.

packages/rum/src/apm-base.js
```diff
@@ -99,9 +99,33 @@ class ApmBase {
...
+   * logging error to the console
+   */
   config(config) {
-    var configService = this.serviceFactory.getService('ConfigService')
+    const configService = this.serviceFactory.getService('ConfigService')
     configService.setConfig(config)
+    /**
+     * Skip validation when agent is not active
+     */
+    if (!configService.isActive()) {
```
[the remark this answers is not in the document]
vigneshshanmugam: This is when agent is inactive, We should not do any extra work when 
```
apm.init({active: false}) 
```

is on the page, It should basically be a noop.

packages/rum-core/src/common/config-service.js
```diff
@@ -217,17 +217,40 @@ class Config {
...
+        'serviceName ' +
+          this.config.serviceName +
+          ' contains invalid characters! (allowed: a-z, A-Z, 0-9, _, -, <space>)'
+      )
+    }
+    /**
+     * Remove all trailing slash for serverUrl since serverUrlPrefix
+     * includes a forward slash for the path
+     */
+    const serverUrl = this.config.serverUrl
```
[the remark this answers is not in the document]
vigneshshanmugam: Good point, This conversion should be done only once. I will move it to `this._config` in apm-base. Having in setConfig makes it to check everytime.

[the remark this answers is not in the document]
vigneshshanmugam: Best place is to do it on getEndPointUrl where we construct the server Url.

packages/rum/src/apm-base.js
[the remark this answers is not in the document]
vigneshshanmugam: Validation should run whenever the agent is started and config is set. Users would be using our API and wont be calling set config manually. Plus this aligns with other agents like `Node.js`. Only thing missing is deactivating the agent when config is invalid which I added as TODO and will go in next breaking change.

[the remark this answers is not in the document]
jahtalab: What I'm worried about is that we will end up in an invalid state if the agent is activated again. IMO we should either not set the config at all or set it with validation. I don't consider this too much of an extra work after all the user is calling this function and expecting to set the config.

packages/rum-core/src/common/config-service.js
```diff
@@ -154,8 +154,13 @@ class Config {
   }
 
   getEndpointUrl() {
-    var url = this.get('serverUrl') + this.get('serverUrlPrefix')
-    return url
+    /**
+     * Remove all trailing slash for serverUrl since serverUrlPrefix
+     * includes a forward slash for the path
+     */
+    this.config.serverUrl = this.config.serverUrl.replace(/\/+$/, '')
```
jahtalab: I don't think this is also a good place to do it, we're creating side-effect in a method that is supposed to only 'get' the url. I think the best place to do it is on setConfig maybe we can have two methods one for validation and one for sanitising values

packages/rum/src/apm-base.js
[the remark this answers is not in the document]
vigneshshanmugam: We validate the agent and log errors when the agent is activated again. Yeah its not a lot of work, The reason why I skipped the validation is to stop the errors from being getting logged since the agent is anyways inactive.

packages/rum-core/src/common/config-service.js
vigneshshanmugam: Hmmm, setConfig is super generic.  We have to do this transformation only once and not every-time when user sets a config. Let me know where you think would be the best place.

jahtalab: Decided to fix the `serverUrl` in setConfig before replacing the configuration.

packages/rum/src/apm-base.js
jahtalab: We decided to make the change into a feature version.

vigneshshanmugam:
@jahtalab Addressed the concerns and also rebased.

packages/rum-core/test/common/config-service.spec.js
```diff
@@ -126,21 +126,27 @@ describe('ConfigService', function() {
...
+    })
+    expect(errors3).toEqual(['Missing serviceName', 'serverUrl'])
 
-    configService.setConfig({ serverUrl: 'test' })
-    result = configService.isValid()
-    expect(result).toBe(true)
+    // Invalid characters in serviceName
+    const errors4 = configService.validate({ serviceName: 'abc.def' })
+    expect(errors4).toEqual([
+      'serviceName abc.def contains invalid characters! (allowed: a-z, A-Z, 0-9, _, -, <space>)'
```
jahtalab: Maybe something like `serviceName "abc.def" contains invalid characters` would be better!

packages/rum-core/test/common/config-service.spec.js
```diff
@@ -126,21 +126,27 @@ describe('ConfigService', function() {
...
+    // Invalid required key serviceName
+    const errors2 = configService.validate({ serviceName: undefined })
+    expect(errors2).toEqual(['Missing serviceName'])
+
+    // Invalid required key serviceName & serverUrl
+    const errors3 = configService.validate({
+      serviceName: undefined,
+      serverUrl: ''
+    })
+    expect(errors3).toEqual(['Missing serviceName', 'serverUrl'])
```
jahtalab: It seems kind of strange to me that the `errors` contain different kind of messages depending on how many errors there are. Maybe making the `validate` function more generic in a way that returns structured error and then the caller would be responsible to generate the error messages would be better, something like:

```
const errors = {
  missingKeys: ['serviceName', 'serverUrl'],
  invalidValues: [{
    key: 'serviceName',
    value: 'abc.def',
    allowed: 'a-z, A-Z, 0-9, _, -, <space>'
  }]
}
```

packages/rum/src/apm-base.js
```diff
@@ -99,9 +99,21 @@ class ApmBase {
...
-    configService.setConfig(config)
+    const configService = this.serviceFactory.getService('ConfigService')
+    const errors = configService.validate(config)
+    if (errors.length === 0) {
+      configService.setConfig(config)
+    } else {
+      const loggingService = this.serviceFactory.getService('LoggingService')
+      const message = 'RUM Agent configuration is invalid: ' + errors.join(', ')
+      configService.setConfig({ active: false })
+      loggingService.error(message)
```
jahtalab: Maybe use `warn` instead of `error` to avoid breaking users tests in dev environment.

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -89,12 +92,11 @@ describe('ApmBase', function() {
     req.send()
     const tr = apmBase.getCurrentTransaction()
     expect(tr).toBeUndefined()
-    expect(loggingService.info).toHaveBeenCalledWith('RUM agent is inactive')
```
jahtalab: Isn't this more clear compared to moving it to the `callFake`?

packages/rum/test/specs/index.spec.js
```diff
@@ -52,16 +52,6 @@ describe('index', function() {
...
     spyOn(apmServer, '_postJson').and.callThrough()
 
-    try {
-      throw new Error('ApmBase test error')
-    } catch (error) {
-      apmBase.captureError(error)
-      if (isPlatformSupported()) {
-        expect(apmServer.errorQueue).toBeUndefined()
-        expect(apmServer.sendErrors).not.toHaveBeenCalled()
-        expect(apmServer._postJson).not.toHaveBeenCalled()
```
jahtalab: Why we don't need this test anymore? this deals with whether or not the platform is supported!

packages/rum-core/src/common/config-service.js
```diff
@@ -217,17 +226,34 @@ class Config {
...
     const requiredKeys = ['serviceName', 'serverUrl']
+    const errors = []
 
-    for (let i = 0; i < requiredKeys.length; i++) {
-      const key = requiredKeys[i]
-      if (this.config[key] == null || this.config[key] === '') {
-        return false
+    Object.keys(properties).forEach(key => {
+      if (requiredKeys.indexOf(key) !== -1 && !properties[key]) {
+        errors.push(errors.length === 0 ? 'Missing ' + key : key)
```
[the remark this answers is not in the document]
vigneshshanmugam: Let's stick to `Missing config` or `Missing options`. both options and config conveys the same thing.

packages/rum-core/test/common/config-service.spec.js
vigneshshanmugam: Very good suggestion. I will change the validate function to output in this format.

packages/rum/src/apm-base.js
vigneshshanmugam: I thought about a bit, But it's actually an error. Agent wont work which is a serious message and we need the users to take action. `warnings` might be ignored silently.

packages/rum/test/specs/index.spec.js
vigneshshanmugam: Probably i screwed up while rebasing. Good catch.

packages/rum/test/specs/apm-base.spec.js
vigneshshanmugam: It didn't work for me in other tests, so I changed it to keep it consistent.

packages/rum/src/apm-base.js
```diff
@@ -105,14 +105,22 @@ class ApmBase {
...
-      const message = 'RUM Agent configuration is invalid: ' + errors.join(', ')
-      configService.setConfig({ active: false })
+      let message = "RUM Agent isn't correctly configured: "
+
+      if (missing.length > 0) {
+        message += 'Missing config - ' + missing.join(', ')
+      }
+
+      invalid.forEach(({ key, value, allowed }) => {
+        message += `, ${key} "${value}" contains invalid characters! (allowed: ${allowed})`
```
jahtalab: Just a minor thing, if there are no missing errors but there are invalid ones the message would be: 

`RUM Agent isn't correctly configured: , ${key} "${value}" contains invalid characters! (allowed: ${allowed})`.

There's an extra comma at the start.

vigneshshanmugam: good catch. fixing it.

packages/rum/src/apm-base.js
```diff
@@ -111,13 +111,15 @@ class ApmBase {
...
 
       if (missing.length > 0) {
-        message += 'Missing config - ' + missing.join(', ')
+        message += 'Missing config - ' + missing.join(separator)
       }
 
       invalid.forEach(({ key, value, allowed }) => {
-        message += `, ${key} "${value}" contains invalid characters! (allowed: ${allowed})`
+        message += missing.length > 0 ? separator : ''
+        message += `${key} "${value}" contains invalid characters! (allowed: ${allowed})`
```
jahtalab: I think it still doesn't work! for example if `missing.length==0` but there are multiple `invalid` errors. In this case they will be concatenated without any spacing.

vigneshshanmugam: fixed.

PR #222, opened 2019-03-20
chore: use 7.0.0-beta1 for tests

jahtalab:
* docs: update source map endpoint


docs/sourcemap.asciidoc
```diff
@@ -130,10 +130,14 @@ Here's how to add the secret token to a curl command:
...
     -F service_version="$SERVICEVERSION" \
     -F bundle_filepath="http://localhost/app.min.js" \
     -F service_name="service-name" \
     -H "Authorization: Bearer <token>"
-----
\ No newline at end of file
+----
+
+
+NOTE: Please note that previously `/v1/rum/sourcemaps` was provided by the APM Server. However, this endpoint
```
vigneshshanmugam: NOTE: The old endpoint `/v1/rum/sourcemaps` provided by the APM server has been deprecated and will be removed starting with APM server version 7.0+.
```suggestion
NOTE: Please note that previously `/v1/rum/sourcemaps` was provided by the APM Server. However, this endpoint 
```

jahtalab: I've reworded it a little bit. I'm not a fan of using "old" in the docs though!

PR #224, opened 2019-03-20
fix(rum): keep page load transaction until load

### LINKED ISSUE #223: avoid ending page load transaction prematurely
With the recent addition of the new TASK API #194, We have added a check when a task is completed to check if the transaction has ended. With the result of this check `detechFinish` which is called during `removeTask`, We are closing the page load transaction prematurely. 

code where its broken - https://github.com/elastic/apm-agent-rum-js/blob/829ffd79b62e8c2dbfe1509050ccff676c9587a8/packages/rum-core/src/performance-monitoring/transaction.js#L150-L156

As a result of that, most of our resource timing and navigation timing information is lost and our marks is missing lots of useful information. 
Screenshot 2019-03-20 at 15 48 02

jahtalab:
Fixes #223

### DIFF
--- a/packages/rum/src/apm-base.js
+++ b/packages/rum/src/apm-base.js
@@ -61,15 +61,20 @@ class ApmBase {
     )
     var configService = this.serviceFactory.getService('ConfigService')
 
+    const pageLoadTaskId = 'page-load'
     var tr = transactionService.startTransaction(
       configService.get('pageLoadTransactionName'),
       'page-load'
     )
+
+    if (tr) {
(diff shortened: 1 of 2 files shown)


jahtalab:
Good point @vigneshshanmugam, will do that 👍!

jahtalab:
@vigneshshanmugam , I've added it to the general-usecase test. I'm refactoring react tests at the moment will add a test there in another PR.

vigneshshanmugam:
@jahtalab Sounds good 👍

PR #227, opened 2019-03-21
chore(rum): update browser versions for android and ios in tests

### LINKED ISSUE #205: Upgrade Android tests on saucelabs to 5.1
I've received the following email from saucelabs:

> In our continuing efforts to provide the best performing, most up-to-date platform for our users, Sauce Labs will discontinue support for the following operating systems and Appium versions on Sauce emulators and simulators on April 4th.

> iOS 9.3, 10.0 and 10.2
>Android 4.4 and 5.0
> Appium versions 1.7.2 and below on both iOS and Android

> We made this decision due to reduced market share of these versions, along with diminished usage by our customers. We strongly recommend customers to periodically update their tests to run on the latest iOS and Appium versions and the most widely used Android versions. Our Platform Configurator has the most up-to-date list of  operating systems and Appium versions supported on Sauce emulators and simulators.

Related: https://github.com/elastic/apm-agent-rum-js/issues/132
cc @vigneshshanmugam

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#205 
+ Unit and E2E tests are run on same set of browsers to catch different bugs during E2E.
+ Safari 11.0 - 11.3 has a bug in fetch spec, window.fetch() without params doesn't get rejected. 
+ Ios emulators have bug when setting script timeout in appium. Removed them from the E2E tests till new appium version is available in sauce labs
+ `window.URL` is broken in IE 11 - it took me 2 hours to figure out why this happens and have no clue till i remote debugged IE 11. 
+ linted .jsx files with proper fixes. 
+ Remove chrome 46 and firefox 42  since the global usage was 0% and replace with other versions which has some usage stats.

### DIFF
--- a/dev-utils/webdriver.js
+++ b/dev-utils/webdriver.js
@@ -40,11 +40,6 @@ function isChrome() {
   return browserName.indexOf('chrome') !== -1
 }
 
-function isIOSDevice() {
-  const platformName = browser.capabilities.platformName.toLowerCase()
-  return platformName === 'ios'
-}
-
 function assertNoBrowserErrors(whitelist) {
   return new Promise((resolve, reject) => {
     /**
@@ -53,9 +48,6 @@ function assertNoBrowserErrors(whitelist) {
     if (!isChrome()) {
       return resolve()
     }
-    // TODO: Bug in ChromeDriver: Need to execute at least one command
-    // so that the browser logs can be read out!
-    browser.execute('1+1')
     var response = browser.getLogs('browser')
     var failureEntries = []
     var debugLogs = []
@@ -138,6 +130,5 @@ module.exports = {
       // done()
     }
   },
-  isChrome,
-  isIOSDevice
+  isChrome
 }
--- a/dev-utils/webdriver.js
+++ b/dev-utils/webdriver.js
@@ -40,6 +40,11 @@ function isChrome() {
   return browserName.indexOf('chrome') !== -1
 }
 
+function isIOSDevice() {
+  const platformName = browser.capabilities.platformName.toLowerCase()
+  return platformName === 'ios'
+}
+
 function assertNoBrowserErrors(whitelist) {
   return new Promise((resolve, reject) => {
     /**
@@ -133,5 +138,6 @@ module.exports = {
       // done()
     }
   },
-  isChrome
+  isChrome,
+  isIOSDevice
 }
--- a/packages/rum/test/e2e/manual-timing/app.e2e-spec.js
+++ b/packages/rum/test/e2e/manual-timing/app.e2e-spec.js
@@ -28,9 +28,6 @@ const { verifyNoBrowserErrors } = require('../../../../../dev-utils/webdriver')
 describe('manual-timing', function() {
   it('should run manual timing', async function() {
     browser.url('/test/e2e/manual-timing/index.html')
-    browser.setTimeout({
-      script: 10000
-    })
     browser.waitUntil(
       () => {
         return $('#test-element').getText() === 'Passed'
--- a/packages/rum/test/e2e/manual-timing/app.e2e-spec.js
+++ b/packages/rum/test/e2e/manual-timing/app.e2e-spec.js
@@ -23,17 +23,11 @@
  *
  */
 
-const {
-  verifyNoBrowserErrors,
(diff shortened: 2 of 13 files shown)


jahtalab:
Tests are failing on Mobile Safari 11.0.0

```
Mobile Safari 11.0.0 (iOS 11.0.1) xhrPatch should support native fetch exception FAILED
	Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL
```

packages/rum/test/e2e/manual-timing/app.e2e-spec.js
```diff
@@ -28,9 +28,6 @@ const { verifyNoBrowserErrors } = require('../../../../../dev-utils/webdriver')
 describe('manual-timing', function() {
   it('should run manual timing', async function() {
     browser.url('/test/e2e/manual-timing/index.html')
-    browser.setTimeout({
```
vigneshshanmugam: The reason why I removed it because the latest appium version in saucelabs seems to have bug and there is no workaround other than using the default script timeout of 30 seconds. https://w3c.github.io/webdriver/#timeouts

vigneshshanmugam: Lol. IE uses 0 seconds and fails.. Trying another way.

vigneshshanmugam:
Failure is due to downtime in docker.elastic.co. Will restart once its up.

vigneshshanmugam:
@jahtalab PR is good for review.

dev-utils/webdriver.js
```diff
@@ -40,6 +40,11 @@ function isChrome() {
   return browserName.indexOf('chrome') !== -1
 }
 
+function isIOSDevice() {
```
roncohen: not used anywhere. Did you intend to use it here? https://github.com/elastic/apm-agent-rum-js/pull/227/files#diff-058a3bfe50baa67d640de5bc3c79f90bR41

packages/rum/wdio.conf.js
```diff
@@ -25,37 +25,24 @@
...
-    deviceName: 'android emulator',
-    browserName: 'browser',
-    platformVersion: '5.0',
-    platformName: 'android'
-  }
-].map(capability => ({ tunnelIdentifier, ...capability }))
+
+/**
+ * Skip the ios platform on E2E tests because of script
+ * timeout issue in Appium
```
roncohen: does this means we'll not be running E2E tests on iOS?

vigneshshanmugam: Yeah, hoping for saucelabs to release a new version of appoium dist. They currently only have 1.9 whereas new appium is already on 1.12. We never ran before, But other tests like integration and unit is covered.

dev-utils/webdriver.js
vigneshshanmugam: Good catch, Wanted to skip timeouts for this version but did not work. I ll give another try or remove this completely.

vigneshshanmugam: Removed since it did not help.

dev-utils/webdriver.js
```diff
@@ -48,9 +48,6 @@ function assertNoBrowserErrors(whitelist) {
     if (!isChrome()) {
       return resolve()
     }
-    // TODO: Bug in ChromeDriver: Need to execute at least one command
-    // so that the browser logs can be read out!
-    browser.execute('1+1')
```
vigneshshanmugam: Not required anymore with webdriver 5.0 which we upgraded with this PR #212

packages/rum/wdio.conf.js
roncohen: OK

nik9000:
@elasticmachine, run docs build

nik9000:
@elasticmachine, run docs build

vigneshshanmugam:
Race condition on E2E tests becomes more prone when switching browsers because of the way we schedule task which might not be consistent in all browsers due to microtasks execution order bugs.

vigneshshanmugam:
Looks good after rebasing with master. The errors were due to this bug we had before https://github.com/elastic/apm-agent-rum-js/issues/223

PR #230, opened 2019-03-27
perf(rum): avoid setting config twice during agent initialization

vigneshshanmugam:
+ Small optimisation to remove setting the config twice during initialisation. We already have them as defaults in the config-service and this does not provide any value.


vigneshshanmugam:
I am not sure why its required, the defaults is already `js-base` in `config-service`. https://github.com/elastic/apm-agent-rum-js/blob/59671e49048c775ab70013e53290095eb106ebae/packages/rum-core/src/common/config-service.js#L72

IMO, it shouldn't be required. Its up to the user if he decides to set some other name for the agent which is think its another issue since we should not let user change the agent name and version.

vigneshshanmugam:
Preceded by #236

PR #232, opened 2019-03-27
[docs] CORS documentation

### LINKED ISSUE #66: Add a note about configuring CORS to getting started
if APM server is deployed in an origin different to the page's origin. The browser will make an OPTIONS request to the APM server prior to making the original POST request.

* We should explain why this process is necessary (we can reference https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS).
* We should also mention that users need to configure the `allow_origins ` in their APM server configuration (by default the APM server allows all origins). 
* We can reference [APM Server - RUM configuration](https://www.elastic.co/guide/en/apm/server/current/configuration-rum.html)
* We should also mention which cross-origin headers are sent and received (browser <--> apm server), this is useful for configuring proxies.
* We should mention that the preflight request will fail if RUM endpoint is disabled on the APM server.

It might make sense to add a guide to the docs and reference it in the getting started.

Any thoughts on this @bmorelli25, @vigneshshanmugam ?

bmorelli25:
CORS documentation! Closes #66:
- [x] Explain why this process is necessary
- [x] Mention that users need to configure the `allow_origins ` in their APM server configuration
- [x] Reference [APM Server - RUM configuration](https://www.elastic.co/guide/en/apm/server/current/configuration-rum.html)
- [x] Mention that the preflight request will fail if RUM endpoint is disabled on the APM server.
- [x] Mention which cross-origin headers are sent and received (browser <--> apm server), this is useful for configuring proxies.
---
~I have some questions:~
* ~Where can I find the cross-origin headers to complete the last checkbox above?~
* ~It's not immediately clear to me if any configuration needs to be done in the rum agent itself, or if all configuration is done in the Server?~
* ~Should we mention the `distributeTracingOrigins` config option here as well? If so, what's important?~

### DIFF
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -19,6 +19,10 @@ You will also need to {apm-server-ref-v}/rum.html[enable frontend endpoints in t
 [[getting-started]]
 == Getting started
 
+* <<using-package-managers>>
+* <<using-script-tags>>
+* <<configuring-cors>>
+
 [float]
 [[using-package-managers]]
 === Using Package Managers
@@ -71,7 +75,7 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
 NOTE: Currently our minified JavaScript bundle is about 16KB (gzipped).
 
 [float]
-[[Configuring CORS]]
+[[configuring-cors]]
 === Configuring CORS
 
 If APM Server is deployed in an origin different than the page's origin,
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -76,22 +76,37 @@ NOTE: Currently our minified JavaScript bundle is about 16KB (gzipped).
 
 If APM Server is deployed in an origin different than the page's origin,
 you will need to configure Cross-Origin Resource Sharing (CORS).
+To configure CORS, simply:
 
-TIP: To learn more about CORS, and why this process is necessary, see the MDN page on
-https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS[Cross-Origin Resource Sharing].
+1. {apm-server-ref-v}/configuration-rum.html[Enable RUM support] in APM Server.
+2. Adjust the {apm-server-ref-v}/configuration-rum.html#_literal_allow_origins_literal[`apm-server.rum.allow_origins`] configuration.
+By default, APM Server allows all origins.
 
-To configure CORS,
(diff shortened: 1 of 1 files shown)


docs/intro.asciidoc
```diff
@@ -69,3 +69,29 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
...
+you will need to configure Cross-Origin Resource Sharing (CORS).
+
+TIP: To learn more about CORS, and why this process is necessary, see the MDN page on
+https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS[Cross-Origin Resource Sharing].
+
+To configure CORS,
+//configuration goes here
+
+When using this setup,
+the RUM agent will send an `HTTP OPTIONS` request prior to making the original `POST` request.
```
jahtalab: Technically the RUM agent only makes the original POST request, however, since the browser checks if this request is a cross-origin request, if it is (e.g. in this case) the browser automatically makes an OPTIONS request to the server to make sure the original POST request is allowed by the server and if this check passes then the browser will send the original POST request. From the RUM agent's perspective we only make one http request with method set to POST.

The function of the OPTIONS request (also called preflight request) is to ensure that the server knows about the request's origin and that the server is prepared to accept the request with the specified method and headers.

docs/intro.asciidoc
```diff
@@ -69,3 +69,29 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
...
+This `OPTIONS` request will ensure all headers and `HTTP` methods are supported,
+and that the origin is allowed. 
+This request will fail if RUM support is not configured in the APM Server.
+See {apm-server-ref-v}/configuration-rum.html[Set up Real User Monitoring (RUM) support]
+for instructions on how to enable RUM support.
+By default, APM Server allows all origins:
+{apm-server-ref-v}/configuration-rum.html#_literal_allow_origins_literal[`apm-server.rum.allow_origins: ['*']`].
```
jahtalab: The default configuration is too broad to make the APM server work out of the box, but users are expected to limit `allow_origins` to their own origins only.

bmorelli25:
@jahtalab, thank you! I apologize for naivety on the subject, and I really, really appreciate the great feedback. I've completely reworked my original docs. This round should be much better 🤞

jahtalab:
The test failures are unrelated to these changes!

PR #236, opened 2019-04-01
fix(rum-core): hardcode agent name and version in service metadata

vigneshshanmugam:
+ Two reasons
    - Its possible to break the APM UI by simply changing the name of the agent 
    - Be consistent with other agents

+ One more issue with our current approach is that, When users use our package directly instead of our distribution, Our agent name will never be changed unless the user builds it locally using our webpack config due to the way we replace the version using string-replace-load which is a bug. 

```js
const rumjs = require('elastic-apm-rum')


// application code


// user builds the bundle using webpack, parcel or whatever toolchain

// bundle would still contains
agentVersion: '%%agent-version%%',

```

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -23,6 +23,7 @@
  *
  */
 
+import { version as agentVersion } from '../../package.json'
 const Queue = require('./queue')
 const throttle = require('./throttle')
 const { sanitizeString } = require('./utils')
@@ -65,8 +66,8 @@ class ApmServer {
       name: sanitizeString(cfg.get('serviceName'), stringLimit, true),
       version: sanitizeString(cfg.get('serviceVersion'), stringLimit),
       agent: {
-        name: sanitizeString(cfg.get('agentName'), stringLimit),
-        version: sanitizeString(cfg.get('agentVersion'), stringLimit)
+        name: 'js-base',
+        version: agentVersion
       },
       language: {
         name: 'javascript'
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -23,7 +23,6 @@
  *
  */
 
-import { version as agentVersion } from '../../package.json'
 const Queue = require('./queue')
 const throttle = require('./throttle')
 const { sanitizeString } = require('./utils')
@@ -67,7 +66,7 @@ class ApmServer {
       version: sanitizeString(cfg.get('serviceVersion'), stringLimit),
       agent: {
         name: 'js-base',
-        version: agentVersion
+        version: '4.0.1'
       },
       language: {
         name: 'javascript'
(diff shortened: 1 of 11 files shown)


packages/rum-core/src/common/apm-server.js
```diff
@@ -65,8 +65,8 @@ class ApmServer {
       name: sanitizeString(cfg.get('serviceName'), stringLimit, true),
       version: sanitizeString(cfg.get('serviceVersion'), stringLimit),
       agent: {
-        name: sanitizeString(cfg.get('agentName'), stringLimit),
-        version: sanitizeString(cfg.get('agentVersion'), stringLimit)
+        name: 'js-base',
+        version: '4.0.1'
```
vigneshshanmugam: The reason for using this instead of 
```js
const {version} = require('../../package.json')
```
is because the package.json would simply pulled in to the bundle because tree shaking wont work properly in CJS modules. We can switch to that once we migrate to ES6 modules.

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -368,16 +368,17 @@ describe('ApmServer', function() {
...
     })
 
     expect(apmServer.createServiceObject()).toEqual({
       name: 'test',
       version: '0.0.1',
       environment: 'staging',
-      agent: { name: 'rum', version: '3.0.0' },
+      agent: {
+        name: 'js-base',
+        version: jasmine.stringMatching(/^([0-9]+)\.([0-9]+)\.([0-9]+)/)
```
vigneshshanmugam: test against `package.json`

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -368,16 +369,18 @@ describe('ApmServer', function() {
     configService.setConfig({
       serviceName: 'test',
       serviceVersion: '0.0.1',
-      environment: 'staging',
-      agentName: 'rum',
-      agentVersion: '3.0.0'
+      environment: 'staging'
     })
 
+    /** To catch agent version mismatch during release */
```
jahtalab: I know we talked about this, but doesn't the tests run before the version change? meaning this test would pass anyway because the version change would happen after.

vigneshshanmugam: Yeah you are right, We cannot rely on this test for releases. 

This would work for both our bundles as well as our consumers. WDYT?
```js
import . {version} from '../package.json' ?
```

jahtalab: I think this won't work either since the build happens (as part of tests) before lerna changes the version.

I'm also not very fond of referencing package.json in our source code!

vigneshshanmugam:
@jahtalab Check this now, We now check the agent version before lerna commits the version and pushes the tags.

vigneshshanmugam:
@jahtalab Please ignore test failures in chrome 49. This is good to be merged.

Issue #159, opened 2019-01-30
Optimize bundle size


vigneshshanmugam:
Some possible optimisations that we can refactor in our code base to reduce the bundle size. 

+  **Config object** contains lots of keys with long names that gets used across most of the files which cannot be mangled by the minifier resulting in bundle size increase
+ **Accessing properties of window**, We do a lof of checks for properties in the window and we have to check if a property/function exists in the window (Ex: window.performance - atleast checked more than 10 times), These checks cannot be mangled by the minifier and most of the libraries do something smart here by wrapping in an IIFE
```js
((function(perf){
  var entries = perf.getEntries();
  var timing = perf.timing

})(window.performance)
```

In the above case, window.performance is used in only one place and rest of the code are mangled. 
+ **Factory pattern for services**, do we even need them? They add some size trying to pass around and also for performance. 

+ **Unused Code**: Transaction and stack service has unused code that is required only for testing . and can be refactored. 

+ **ES6 Modules**: To utilise some of the performance techniques like Tree shaking with the bundlers/minifiers, We have to migrate the codebase to es6 modules.

vigneshshanmugam:
### Savings: 

#### properties of window: 

I tried to experiment with fine tuning the window properties ( window.performance, window.location, window.fetch) as per the above mentioned example. 


Filename | Before | After 
-------- | --------| ---------
elastic-apm-js-base.min.js |  53521  | 53098
elastic-apm-js-base.min.js (After gzipping) | 16.3 kB | 16.1 kB

vigneshshanmugam:
Savings: 

**Factory Pattern for services**

Filename | Before | After 
-------- | --------| ---------
elastic-apm-js-base.min.js |  54047  | **51913**
elastic-apm-js-base.min.js (After gzipping) | 16.04KB | **15.81KB**

roncohen:
interesting! thanks for investigating. I wonder if you think there's added complexity when removing the factory pattern?

vigneshshanmugam:
Final changes after removing factory pattern 

Filename | Before(master) | After 
-------- | --------| ---------
elastic-apm-js-base.min.js |  54047  | **51570**

jahtalab:
We just had a very productive meeting, we decided to keep the service factory.

Some of the reasons to keep the service factory:
* We don't want to create file level instances which creates issues with our testing
[...]

PR #237, opened 2019-04-02
perf(rum): move to ES6 modules to reduce bundle size

vigneshshanmugam:
+ part of elastic/apm-agent-rum-js#159 
+ Need to transform node test files for Integration tests before jasmine

#### Savings - gzipped
Filename | Before(master) | After 
-------- | --------| ---------
elastic-apm-js-base.min.js |  16.04  | **15.75**


packages/rum/src/index.js
```diff
@@ -23,27 +23,19 @@
...
-  default: apmBase.init.bind(apmBase),
-  init: apmBase.init.bind(apmBase),
-  ApmBase,
-  apmBase,
-  apm: apmBase
-}
+const init = () => apmBase.init.bind(apmBase)
 
-module.exports = exports
+export default apmBase
```
jahtalab: The previous default was the `init` method on `apmBase` and not the `apmBase` itself.

vigneshshanmugam: Good catch 👍

packages/rum-core/src/index.js
```diff
@@ -25,21 +25,24 @@
...
+  return serviceFactory
+}
+
+export {
+  createServiceFactory,
   ServiceFactory,
-  patching,
-  utils
+  patchAll,
+  subscription,
```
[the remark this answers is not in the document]
vigneshshanmugam: changed to patchSubscription as discussed.

packages/rum/src/index.js
```diff
@@ -23,27 +23,19 @@
...
 
-var exports = {
-  __esModule: true,
-  default: apmBase.init.bind(apmBase),
-  init: apmBase.init.bind(apmBase),
-  ApmBase,
-  apmBase,
-  apm: apmBase
-}
+const init = () => apmBase.init.bind(apmBase)
```
jahtalab: The init is defined as a function that returns a function which is the actual init function. instead it should just be the actual init function

vigneshshanmugam:
@jahtalab Check last 2 commits where I have added tests for exports and added support for transplred modules for Node >=8 version.

vigneshshanmugam:
Test failures are unrelated to the PR. It fails on Chrome 49 which is a known race condition bug.

packages/rum-core/package.json
```diff
@@ -13,15 +15,21 @@
     "directory": "packages/rum-core"
   },
   "scripts": {
+    "prepublish": "npm run build:main",
+    "build:main": "npx babel src -d lib --config-file ../../babel.node.js",
```
jahtalab: Maybe we should put this under `dist/lib` as well, WDYT?

packages/rum/package.json
```diff
@@ -2,7 +2,9 @@
   "name": "@elastic/apm-rum",
   "version": "4.0.1",
   "description": "Elastic APM JavaScript agent",
-  "main": "src/index.js",
+  "main": "dist/lib/index.js",
+  "module": "src/index.js",
+  "sideEffects": false,
```
jahtalab: Doesn't patching browser API counts as sideEffects?

packages/rum/run-script.js
```diff
@@ -113,26 +113,26 @@ function serveE2e(servingPath, port) {
...
       // The stack is not useful in this context.
       err.stack = ''
       cb(err)
     } else {
       cb()
     }
   })
-
-  jrunner.showColors(true)
-  jrunner.addReporter(JasmineRunner.ConsoleReporter(jrunner))
```
jahtalab: Just to make sure this is intentional, we don't need to add the console report manually?

packages/rum-core/package.json
vigneshshanmugam: sure, lets do it.

packages/rum-core/package.json
```diff
@@ -13,15 +15,21 @@
...
     "karma": "karma start",
     "karma:bench": "karma start karma.bench.conf.js",
     "karma:coverage": "karma start --coverage",
     "test:unit": "node ../../dev-utils/run-script.js runUnitTests true packages/rum-core",
     "test": "npm run test:unit"
   },
   "files": [
-    "src"
+    "src",
+    "lib"
```
jahtalab: I think this is not needed anymore.

.gitignore
```diff
@@ -15,3 +15,5 @@ coverage
 
 # ignore lerna-debug and npm-debug logs
 *debug.log
+
+/packages/*/lib
```
jahtalab: This is not needed anymore, right?

packages/rum-core/package.json
jahtalab: It actually needs to be the dist folder.

vigneshshanmugam: Oops yeah, changing it now

PR #239, opened 2019-04-03
fix(rum): return noop agent when config is inactive

### LINKED ISSUE #210: active: false should be a noop during initialization phase
In RUM agent, when `active: false` is set in the config, 
+ We initialize the agent with active listeners on window.load, window.onerror and also patch fetch and XHR calls. 
+ The agent just dont send the transaction as well the erros captured during active: false. 

This is not the intended behaviour of active flag and it does not align with other agents. 

`active: false` would simply a noop in the agent which means we wont be patching any browser API's and also listening for events.

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#210 
+ Extracted from elastic/apm-agent-rum-js#216 
+ Refactored few methods to reduce the code size.

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -34,7 +34,10 @@ class ApmServer {
     this._configService = configService
     this._loggingService = loggingService
     this.logMessages = {
-      invalidConfig: { message: 'Configuration is invalid!', level: 'warn' }
+      invalidConfig: {
+        message: 'RUM agent configuration is invalid!',
+        level: 'warn'
+      }
     }
 
     this.errorQueue = undefined
@@ -178,21 +181,17 @@ class ApmServer {
   }
 
   addError(error) {
-    if (this._configService.isActive()) {
-      if (!this.errorQueue) {
-        this.initErrorQueue()
-      }
-      this.throttleAddError(error)
+    if (!this.errorQueue) {
+      this.initErrorQueue()
     }
+    this.throttleAddError(error)
   }
 
   addTransaction(transaction) {
-    if (this._configService.isActive()) {
-      if (!this.transactionQueue) {
-        this.initTransactionQueue()
-      }
-      this.throttleAddTransaction(transaction)
+    if (!this.transactionQueue) {
+      this.initTransactionQueue()
     }
+    this.throttleAddTransaction(transaction)
   }
 
   warnOnce(logObject) {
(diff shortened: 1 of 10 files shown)


packages/rum-core/src/common/apm-server.js
```diff
@@ -252,29 +226,47 @@ class ApmServer {
...
-            NDJSON.stringify({ metadata: { service: filteredPayload.service } })
-          )
-          const ndjsonPayload = ndjson.join('')
-          return this._postJson(endPoint, ndjsonPayload)
-        } else {
-          this._loggingService.warn('Dropped payload due to filtering!')
-        }
-      }
-    } else {
+  _send(data, type = 'transaction') {
```
jahtalab: I don't think we should have a default value here, if `type` isn't provided we should either throw or return!

I think we should also check `data`!

packages/rum-core/src/common/config-service.js
```diff
@@ -98,18 +104,17 @@ class Config {
 
       distributedTracing: true,
       distributedTracingOrigins: [],
-      distributedTracingHeaderValueCallback: undefined,
+      distributedTracingHeaderValueCallback: getDtHeaderValue,
```
jahtalab: I think we might remove this config option once our we transition to the w3c `traceparent` header.

packages/rum-core/src/common/config-service.js
```diff
@@ -211,26 +213,16 @@ class Config {
...
-
-    setTag(key, value, this.config.context.tags)
-  }
-
-  addTags(tags) {
     var keys = Object.keys(tags)
-    keys.forEach(k => {
-      this.setTag(k, tags[k])
-    })
+    keys.map(k => setTag(k, tags[k], this.config.context.tags))
```
jahtalab: Did you mean to use `forEach` since you are not mapping the array to anything!

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -154,29 +179,6 @@ describe('ApmBase', function() {
     expect(tr.name).toBe('ZoneTransaction')
   })
 
-  it('should instrument xhr when not active', function(done) {
```
jahtalab: We still instrument XHR even if not active! patching is called in the `bootstrap.js` file

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -67,6 +67,31 @@ describe('ApmBase', function() {
     })
   })
 
+  it('should be noop when agent is not active', done => {
```
jahtalab: What will happen if users call `startTransaction` on an inactive agent. We should have a test for that as well.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: fixed as well.

PR #245, opened 2019-04-09
perf(rum): remove debug logs on production build

vigneshshanmugam:
+ part of elastic/apm-agent-rum-js#159 
+ fix webpack multiple entry bug with workaround which helped reduce bundle size even further. 
+ All the debug logs are removed by default on the production bundle to reduce the bundle size. 
+ Added docs on creating optimised builds using bundlers. 

### Improvements

Filename | Before(master) | After
-- | -- | --
elastic-apm-js-base.min.js | 16.4 | 15.72
elastic-apm-js-base.min.js(no-gzip) | 53.2 | 50.3
elastic-apm-opentracing.umd.min.js | 17.96 | 17.08
elastic-apm-opentracing.umd.min.js(no-gzip) | 59.5 | 55.9


docs/intro.asciidoc
```diff
@@ -72,7 +73,49 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
 </script>
 ----
 
-NOTE: Currently our minified JavaScript bundle is about 16KB (gzipped).
+NOTE: Currently the optimized(minified + gzipped) JavaScript bundle size is about 16KiB.
```
bmorelli25:
```suggestion
NOTE: Currently the optimized (minified + gzipped) JavaScript bundle size is about 16KiB.
```

docs/intro.asciidoc
```diff
@@ -72,7 +73,49 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
...
+
+
+[float]
+[[using-production-build]]
+=== Using Production Build
+
+By default, RUM agent logs all the debug messages to the browser console. These logs are very useful in development. However, they make the RUM agent bundle size larger so you should make sure to use the optimized production version when deploying your application.
```
bmorelli25: This sentence is a bit confusing to me: "However, they make the RUM agent bundle size larger so you should make sure to use the optimized production version when deploying your application." What about reworking the entire paragraph? Something like this reads a bit easier to me:

```suggestion
By default, the RUM agent logs all debug messages to the browser console.
While these logs are very useful in development, they increase the RUM agent bundle size.
For this reason, make sure to use the optimized production version when deploying your application.
You can find instructions for building optimized code, with different bundlers, below.
```

docs/intro.asciidoc
```diff
@@ -72,7 +73,49 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
...
+
+You can find instructions for building optimized code below for different bundlers
```
bmorelli25:
```suggestion
```
Let's wrap this sentence into the previous paragraph since it fits in well there

docs/intro.asciidoc
```diff
@@ -72,7 +73,49 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
...
+const { EnvironmentPlugin } = require('webpack')
+
+plugins: [
+  new EnvironmentPlugin({
+    NODE_ENV: 'production'
+  })
+]
+----
+
+You can learn more about this in https://webpack.js.org/plugins/environment-plugin[Webpack documentation].
```
bmorelli25:
```suggestion
Learn more in the https://webpack.js.org/plugins/environment-plugin[Webpack documentation].
```

docs/intro.asciidoc
```diff
@@ -72,7 +73,49 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
...
+[[using-production-build]]
+=== Using Production Build
+
+By default, RUM agent logs all the debug messages to the browser console. These logs are very useful in development. However, they make the RUM agent bundle size larger so you should make sure to use the optimized production version when deploying your application.
+
+You can find instructions for building optimized code below for different bundlers
+
+[float]
+==== Webpack
+For optimized webpack production build, include the Environment/Define plugin in the webpack configuration.
```
bmorelli25:
```suggestion
For an optimized webpack production build, include the Environment/Define plugin in the webpack configuration.
```
Does an "an" make sense here?

docs/intro.asciidoc
```diff
@@ -72,7 +73,49 @@ https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG
...
+  })
+]
+----
+
+You can learn more about this in https://webpack.js.org/plugins/environment-plugin[Webpack documentation].
+
+[float]
+==== Rollup
+
+For optimized rollup production build, include the replace plugin which ensures the right build environment is used.
```
bmorelli25:
```suggestion
For an optimized rollup production build, include the replace plugin to ensure the right build environment is used. 
```

Does an "an" make sense here?

docs/intro.asciidoc
vigneshshanmugam: Yeah simpler, As always thanks @bmorelli25 😄

vigneshshanmugam:
@bmorelli25 Thanks Brandon 😄

vigneshshanmugam:
@jahtalab Can you have a look?

packages/rum-core/src/common/apm-server.js
```diff
@@ -198,7 +198,7 @@ class ApmServer {
     if (logObject.level === 'warn') {
       logObject.level = 'debug'
       this._loggingService.warn(logObject.message)
-    } else {
+    } else if (process.env.NODE_ENV !== 'production') {
```
jahtalab: I think the reverse is more readable, i.e. `process.env.NODE_ENV === 'development'`.

besides that also makes sure if they don't supply the `NODE_ENV ` the still get the optimised

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -28,12 +28,10 @@ import { isUndefined, extend } from '../common/utils'
 import Subscription from '../common/subscription'
 import { captureHardNavigation } from './capture-hard-navigation'
 
+const __DEV__ = process.env.NODE_ENV !== 'production'
```
jahtalab: Can we use `__DEV__ ` everywhere, instead of using `process.env.NODE_ENV`

vigneshshanmugam: We cant do that, it can be imported from constant file but wont be dead code eliminated and replaced. so only way is to use process.env in file level.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: Not really, Only for production mode debug logs should be removed and must be enabled in dev, test and other sets of possible environments. All of the libraries follow this approach and hence its better to stick to convention.

vigneshshanmugam:
@jahtalab I am waiting for your testing before I go ahead and add the same flags to some of the recent debug messages that we have landed in master.  Will do rebase and add once you are done with your test.

jahtalab:
Decided to rebase and apply these changes to new logs that have merged to master since.

vigneshshanmugam:
@jahtalab Rebased and applied the changes.

PR #256, opened 2019-04-30
fix(rum-core): In truncate check for empty values

jahtalab:
instead of falsy values when checking required fields. 

Checking for falsy doesn't work for `0` which is a valid value for a required field but we instead try to use the placeholder.

### DIFF
--- a/packages/rum-core/src/common/truncate.js
+++ b/packages/rum-core/src/common/truncate.js
@@ -99,7 +99,12 @@ function truncate(
   required = false,
   placeholder = 'N/A'
 ) {
-  if (required && !value) {
+  /*
+    The request will fail if we set a string placeholder
+    when the apm server expects a number.
+    However, if this happens it must be a bug.
+  */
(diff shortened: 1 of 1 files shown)


packages/rum-core/src/opentracing/tracer.js
```diff
@@ -74,7 +74,7 @@ class Tracer extends otTracer {
     var span
     var currentTransaction = this.transactionService.getCurrentTransaction()
 
-    if (currentTransaction && !currentTransaction.ended) {
+    if (currentTransaction) {
```
vigneshshanmugam: Do we need to create spans if there is no active transaction?

jahtalab: I think this was part of [this refactoring](https://github.com/elastic/apm-agent-rum-js/pull/259/files#diff-074367311a4235f7774f1d58a437e7b6R66). I will move it there to make it clear.

Issue #90, opened 2018-11-08
Capture route change transactions


jahtalab:
There are multiple approaches for capturing the route change transactions:
1. Instrumenting the history API
2. Providing framework-specific instrumentation

The first solution would provide support for a wide range of frameworks (potentially with some work on the user side) but we don't have access to the route name, we only have the URL which might contain IDs. To solve this issue we can provide `onStart` and `onEnd` event listeners in the config options for transaction start and transaction end respectively, as an example consider the following: 

Please note that users don't need to provide both event listeners.

```
apm.init({
   onStart: (transaction)=>{ 
      transaction.name= 'transactions-name' 
      return transaction
   },
   onEnd: (transaction)=>{ 
      transaction.name= 'transactions-name' 
      return transaction
   }
})
```

The drawback for the second approach is that we need to provide instrumentation for each framework.

Potential pitfalls for react-route:
- Multiple routes could match with any route change, but we should only create one transaction for the whole route change (need to investigate this kind of use-cases)

Currently, I think we should take both approaches, the first approach would serve as a base for the second approach. i.e. a framework specific instrumentation can register event listeners for transactions and fix the transaction name if necessary.

Update:
The above event listeners are not useful for instrumenting react-router therefore we should provide react specific instrumentation.

tbragin:
@jahtalab Really excited about this feature coming in the JS agent. Am I reading it correctly that we are supporting SPA for the React framework only? Or is it framework-agnostic?

jahtalab:
@tbragin , at first we decided to take framework-agnostic approach but in a lot of cases usage would be harder for user, so we will have to provide framework specific integrations. We're starting with React, but Angular will follow after that.

mickeyklai:
@jahtalab Great work!
What about Vue.js?

jahtalab:
Thanks for reaching out @mickeyklai !
I've created [this issue](https://github.com/elastic/apm-agent-rum-js/issues/314) to discuss Vue.js integration.

There's also [this](https://github.com/elastic/apm-agent-rum-js/issues/292) for Angular integration.

Closing this one in favor of framework-specific issues.

freezy:
Any documentation about the "hard", framework agnostic usage? Is it even possible without significant effort?

jahtalab:
@freezy, Thanks for reaching out!

[...]

PR #259, opened 2019-05-02
feat(rum-core): capture history.pushState calls as transactions

jahtalab:
This is part of the effort to [capture route change transactions](https://github.com/elastic/apm-agent-rum-js/issues/90). 

- patch history API
- refactoring of `TransactionService.startTransaction`
- introduce transaction reusability

### DIFF
--- a/packages/rum-core/src/common/constants.js
+++ b/packages/rum-core/src/common/constants.js
@@ -37,6 +37,11 @@ const FETCH_SOURCE = 'fetch'
 const XMLHTTPREQUEST_SOURCE = 'XMLHttpRequest.send'
 
 /**
+ * History sources
+ */
+const HISTORY_PUSHSTATE = 'history.pushState'
+
+/**
  * Event listener methods
  */
 const ADD_EVENT_LISTENER_STR = 'addEventListener'
@@ -55,6 +60,11 @@ const RESOURCE_INITIATOR_TYPES = [
   'beacon',
   'iframe'
 ]
+
+/**
+ * The amount of time it is allowed for a transaction to be reused in another startTransaction
+ */
+const REUSABILITY_THRESHOLD = 10000
 
 /**
  * Maximum duration of the span that is used to decide if the span is valid - 300 secs / 5 mins
@@ -76,5 +86,7 @@ export {
   REMOVE_EVENT_LISTENER_STR,
   RESOURCE_INITIATOR_TYPES,
   SPAN_THRESHOLD,
-  KEYWORD_LIMIT
+  KEYWORD_LIMIT,
+  HISTORY_PUSHSTATE,
+  REUSABILITY_THRESHOLD
 }
--- a/packages/rum-core/src/common/patching/history-patch.js
+++ b/packages/rum-core/src/common/patching/history-patch.js
@@ -23,27 +23,29 @@
  *
  */
 
-import { patchXMLHttpRequest } from './xhr-patch'
-import { patchFetch } from './fetch-patch'
-import { patchHistory } from './history-patch'
-import Subscription from '../subscription'
-
-const patchSubscription = new Subscription()
+import { INVOKE, HISTORY_PUSHSTATE } from '../constants'
 var alreadyPatched = false
-function patchAll() {
-  if (!alreadyPatched) {
-    alreadyPatched = true
-    patchXMLHttpRequest(function(event, task) {
-      patchSubscription.applyAll(this, [event, task])
-    })
-    patchFetch(function(event, task) {
-      patchSubscription.applyAll(this, [event, task])
-    })
-    patchHistory(function(event, task) {
-      patchSubscription.applyAll(this, [event, task])
-    })
-  }
-  return patchSubscription
-}
 
-export { patchAll, patchSubscription }
+export function patchHistory(callback) {
+  if (alreadyPatched || !window.history) {
+    return
+  }
+  alreadyPatched = true
+
+  const nativePushState = history.pushState
+  if (typeof nativePushState === 'function') {
+    history.pushState = function(state, title, url) {
+      const task = {
+        source: HISTORY_PUSHSTATE,
+        data: {
+          state,
+          title,
+          url
+        }
+      }
+
+      callback(INVOKE, task)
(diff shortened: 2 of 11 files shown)


packages/rum-core/src/common/patching/history-patch.js
```diff
@@ -0,0 +1,51 @@
...
+
+export function patchHistory(callback) {
+  if (alreadyPatched || !window.history) {
+    return
+  }
+  alreadyPatched = true
+
+  const nativePushState = history.pushState
+  if (typeof nativePushState === 'function') {
+    history.pushState = function(state, title, url) {
```
vigneshshanmugam: I believe it's also called for redirects which should be ignored and not captured as part of transaction.

packages/rum-core/test/common/history-patch.spec.js
```diff
@@ -0,0 +1,65 @@
...
+    cancelFn()
+  })
+
+  beforeEach(function() {
+    events = []
+  })
+
+  it('should patch history.pushState', function() {
+    history.pushState(undefined, 'test', 'test')
+
```
vigneshshanmugam: Please add a e2e test as well with history.go and history.back methods and verifying the transactions with those names

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -122,6 +123,12 @@ class PerformanceMonitoring {
           transactionService.removeTask(task.id)
         }
       }
+
+      if (event === INVOKE && task.source === HISTORY_PUSHSTATE) {
+        transactionService.startTransaction(task.data.title, 'route-change', {
```
vigneshshanmugam: The title can be empty and not sure if it's set all the time. we have to handle this case and I am not sure if it's set all the time during popstate events.

packages/rum-core/src/common/patching/history-patch.js
```diff
@@ -0,0 +1,51 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import { INVOKE, HISTORY_PUSHSTATE } from '../constants'
+var alreadyPatched = false
+
+export function patchHistory(callback) {
```
vigneshshanmugam: It's safer to do this patch after we have captured the page load event transaction, if the user registers a state in history using 'pushState', we will accidently capture all our requests as part of route change transaction instead of page-load which will be a broken behaviour..

Would like to have rigorous E2E tests for this change.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
jahtalab: The title is not used in any of the frameworks I've seen. But we don't have to handle it here since the transaction title can be undefined and if it is then it would have 'unknown'.

packages/rum-core/src/common/patching/history-patch.js
jahtalab: I considered this to be a feature, that we will capture the first route change as a part of the page-load transaction. This will give the user a complete perspective to the initial page load.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: Wont it be confusing then? If we are relying on title for the transaction name and multiple paths might lead to `Unknown` because of title being empty on different routes?

packages/rum-core/src/common/patching/history-patch.js
vigneshshanmugam: I am really not sure if this a good feature to be honest, It would simply increase the number of transactions count on the details page plus instead of focusing on the transactions that matter the most, we are simply polluting the UI more less useful `redirected transactions`. 

For the first phase at-least, we can simply drop them and add it if there is a huge interest from users.

packages/rum-core/src/common/constants.js
```diff
@@ -56,6 +61,11 @@ const RESOURCE_INITIATOR_TYPES = [
   'iframe'
 ]
 
+/**
+ * The amount of time it is allowed for a transaction to be reused in another startTransaction
+ */
+const REUSABILITY_THRESHOLD = 10000
```
vigneshshanmugam: Any reasoning behind this particular value?

jahtalab: I'll make the value into ms.

packages/rum-core/src/common/patching/history-patch.js
jahtalab: I'll investigate this scenario more

jahtalab: @vigneshshanmugam , I will add e2e tests to #265 for this scenario

packages/rum-core/src/common/patching/history-patch.js
jahtalab: I won't start a new transaction only redefines the old one which I think should be ok!

packages/rum-core/test/common/history-patch.spec.js
jahtalab: Since the rum-core doesn't have e2e tests, will also add e2e tests to #265

packages/rum-core/src/common/patching/history-patch.js
vigneshshanmugam: Sure, sounds good 👍

packages/rum-core/src/common/patching/history-patch.js
vigneshshanmugam: Have to be careful here since if the redirect is too slow, we will discard it and create a new one. Its an edge case, but might want to think about reusability threshold.

vigneshshanmugam:
Let's add todos for the other PR for the E2E Tests(edge cases) and also docs on supported vs unsupported libs.

jahtalab:
I've added the todos to the other PR!

Issue #262, opened 2019-05-05
Offline Support


kevindoveton:
Hi, we're developing a hybrid offline first application using ionic / cordova and would love to integrate this project. We currently have our own methods of handling http requests whilst offline and were wondering if there is either plans to support offline in this project or if there is a way to override the existing http implementation with our own.

jahtalab:
Hi @kevindoveton , 

Thanks for reaching out and sorry about the slow response.

Currently we don't officially support Cordova. However, it's on our roadmap to add support for it. I will make sure that we  consider the offline support as well.

Also, thanks for creating a PR to resolve this, we need to make sure this approach fits with our current use-cases. I will update the PR as we progress.

leon-vg:
Hi @jahtalab,

I am currently integrating RUM in a hybrid app with offline support. We would like to have offline support, since we would like to log and monitor issues that happen while the user has network issues. 

Could you perhaps share your thoughts for adding offline support? Perhaps we can look into it and come up with a PR to get things started.

jahtalab:
Hi @leon-vg , 

I think the best place to start is by accepting a function in a config option that is called whenever the agent needs to make http request. Similar to how we use `_makeHttpRequest` in [`apm-server.js`](https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/src/common/apm-server.js#L122)

kevindoveton:
#264 provides the basis for the approach that @jahtalab described, we used that in conjunction with our own sync service for over 18mo

leon-vg:
Thanks @jahtalab . Sounds like a good solution and it looks like PR #264 provides the basis for it. However, I saw that that specific PR was closed in 2019 because you decided nót to take this approach. Now I am a bit confused on the best way forward. :)

jahtalab:
I think at the time we wanted to improve the experience with the approach in #264  along with support for hybrid mobile applications #493 but that hasn't materialised yet. 

There are also some challenges that needs to be resolved, e.g. `_makeHttpRequest` is used to make any type of http request to the apm server and it can't completely be replaced with a user-defined config option unless that function has the knowledge of how to communicate with APM server properly and also handle breaking changes properly. so I think we need to refactor `_makeHttpRequest` or add another method to only handle events.

Also the timestamp of the generated events will be the time that apm-server received those events so for offline events, the timestamp will be incorrect.

PR #264, opened 2019-05-09
Expose the Http Request method

kevindoveton:
Could elastic/apm-agent-rum-js#262 be solved with something as simple as exposing the `_makeHttpRequest` method through a config option?


jahtalab:
@kevindoveton, 

Thanks for creating the PR and sorry about the slow response.

We've discussed this internally and have decided that we will not take this approach for offline support. However, the original issue will stay open until we provide a solution for it. Unfortunately, we don't have timeline for offline support at the moment but I will make sure to update the issue.

PR #265, opened 2019-05-09
feat: Initial react integration

jahtalab:
This is part of the effort to [capture route change transactions](https://github.com/elastic/apm-agent-rum-js/issues/90) and provide [react router integration](https://github.com/elastic/apm-agent-rum-js/issues/301)

This PR relies on and should only be merged after elastic/apm-agent-rum-js#259 .

- Add `ApmRoute` component
- Add `withTransaction` to instrument individual components
- Sets `canReuse: true` on page-load transaction sot that it can be used to capture the first route change

Todo:
- [x] Add e2e tests for history.go and history.back


dev-utils/run-script.js
```diff
@@ -51,10 +54,27 @@ function launchSauceConnect() {
   console.log('set MODE=saucelabs to launch sauce connect')
 }
 
+function serveStatic(path = '../', port = 8000) {
```
vigneshshanmugam: Can you modify the serveE2E to account for this, duplicated code.

axw:
Just trying to understand what "canReuse" is meant for. If I understand correctly, there would be an initial page load transaction, and then you'll redefine it when the route change occurs. Is that right? What are the circumstances in which there won't be a transaction when a route change occurs? i.e. when would it start a transaction rather than redefine?

Overloading "startTransaction" feels a bit hard to understand to me, maybe just me though. Did you consider adding a new function along the lines of "startOrRedefineTransaction"? That name is a bit of a mouthful, but maybe the semantics would be a bit clearer.

jahtalab:
@axw , We have multiple situations in which either we want to extend a transaction to include further events (such as the page-load -> route change expansion) or we don't have all the information at the time when we start the transaction, e.g. when `history.pushState` happens we might not have the correct transaction name but later with router instrumentation we can redefine the transaction with the route name.

I did consider refactoring it into a separate method but it seemed to me that it makes the understanding the process of starting a transaction more complicated to understand. But I agree that startTransaction can be refactored to make it easier to read.

axw:
> I did consider refactoring it into a separate method but it seemed to me that it makes the understanding the process of starting a transaction more complicated to understand.

@jahtalab I'd like to hear more about this. Maybe you can provide a code example to illustrate? In general I tend to find overloading functions with special cases makes them more difficult to use for everyone else. The name "startTransaction" implies to me that it'll always start a transaction.

vigneshshanmugam:
@jahtalab I will start reviewing this once its rebased. Was waiting for the test refactor to go in before I start it locally and test it.

jahtalab:
@vigneshshanmugam , rebased!

packages/rum-react/src/get-with-transaction.js
```diff
@@ -0,0 +1,86 @@
...
+    Usage:
+        - Pure function: `withTransaction('name','route-change')(Component)`
+        - As a decorator: `@withTransaction('name','route-change')`
+ */
+
+import React from 'react'
+import hoistStatics from 'hoist-non-react-statics'
+
+function getWithTransaction(apm) {
+  if (!apm) {
```
vigneshshanmugam: this feels unnecessary, We should just be exposing `withTransaction`.

packages/rum-react/src/get-with-transaction.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return function(WrappedComponent) {
+      class ApmComponent extends React.Component {
+        constructor(props) {
+          super(props)
+          /*
+                        We need to start the transaction in constructor because otherwise,
+                        we won't be able to capture what happens in componentDidMount of child components.
+                    */
+          this.transaction = apm.startTransaction(name, type)
```
vigneshshanmugam: Did you test the transaction timing when the component is rendered lazily?

packages/rum-react/src/get-with-transaction.js
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        componentDidMount() {
```
vigneshshanmugam: Also add `componentDidCatch` and capture errors as part of transaction.

packages/rum-react/src/get-with-transaction.js
```diff
@@ -0,0 +1,86 @@
...
+          // todo: should we pass the transaction down (could use react context provider instead)
+          return (
+            <WrappedComponent transaction={this.transaction} {...this.props} />
+          )
+        }
+      }
+
+      ApmComponent.displayName = `withTransaction(${WrappedComponent.displayName ||
+        WrappedComponent.name})`
+      ApmComponent.WrappedComponent = WrappedComponent
```
vigneshshanmugam: Why do we need this wrapped component here?

packages/rum-react/src/index.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+import { apm } from '@elastic/apm-rum'
+import { getWithTransaction } from './get-with-transaction'
+import { getApmRoute } from './get-apm-route'
+
+const withTransaction = getWithTransaction(apm)
+const ApmRoute = getApmRoute(apm)
+
+export { getWithTransaction, withTransaction, ApmRoute }
```
vigneshshanmugam: `getWithTransaction` feels not useful IMO

packages/rum-react/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "test:e2e:supported": "npm run script runE2eTests packages/rum-react/wdio.conf.js",
+    "test:sauce": "npm run build:e2e && npm run script runSauceTests true packages/rum-react test:unit test:e2e:supported",
+    "karma": "karma start",
+    "test": "npm run test:sauce"
+  },
+  "bugs": {
+    "url": "https://github.com/elastic/apm-agent-rum-js/issues"
+  },
+  "dependencies": {
+    "hoist-non-react-statics": "^3.3.0"
```
vigneshshanmugam: `@elastic/apm-rum` and `react-router-dom` must be a dep and `react` must be in peer dep.

packages/rum-react/src/get-with-transaction.js
jahtalab: I think `if (!apm) {` is a mistake I will remove that. But the function `getWithTransaction`  is there so that we can pass an apm instance to it. The reason is that we want to make sure that each tests runs with fresh new apm instance.

packages/rum-react/src/get-with-transaction.js
jahtalab: Can you elaborate on the scenario you have in mind?

packages/rum-react/src/get-with-transaction.js
vigneshshanmugam: We should have `getWithTransaction` function as part of tests then, Having something in `src` to ease testing is not ideal.

packages/rum-react/src/get-with-transaction.js
vigneshshanmugam: React decides when the component tree should be mounted in to the DOM, So I was wondering if we are capturing the transaction time properly during these scenarios - https://github.com/reactjs/rfcs/blob/gaearon-patch-2/text/0000-lazy.md

packages/rum-react/package.json
jahtalab: Yeah, good catch about `@elastic/apm-rum` but I think `react-router-dom` should be a peer dep as well, since we expect to use the version that the user already has installed.

packages/rum-react/src/get-with-transaction.js
jahtalab: I'm not sure how can we pass the apm instance if we don't define it in the source. Would you mind elaborating?

vigneshshanmugam: Something like this. 

```js
import { withTransaction } from "../"
import { createApmBase } from "@elastic/apm-rum"

let apmInstance
beforeEach(() => apmInstance = new createApmBase())

```

jahtalab: That doesn't work, we need to pass the apm instance to withTransaction so that it uses that one.

packages/rum-react/src/get-with-transaction.js
jahtalab: Let's add this after this initial release to avoid having too many changes!

dev-utils/webdriver.js
```diff
@@ -158,6 +167,101 @@ function getWebdriveBaseConfig(path) {
...
+            validCalls &&
+            serverCalls.sendTransactions &&
+            serverCalls.sendTransactions.length >= transactionCount
+        }
+
+        if (validCalls) {
+          console.log('calls', serverCalls)
+          var promises = []
+
+          if (serverCalls.sendErrors) {
```
vigneshshanmugam: this is not required, we already check for the same value in validCalls which covers it.

dev-utils/webdriver.js
```diff
@@ -158,6 +167,101 @@ function getWebdriveBaseConfig(path) {
...
+            promises = promises.concat(
+              serverCalls.sendErrors.map(function(s) {
+                return s.returnValue
+              })
+            )
+          }
+          if (serverCalls.sendTransactions) {
```
vigneshshanmugam: same here, not required.

dev-utils/webdriver.js
```diff
@@ -158,6 +167,101 @@ function getWebdriveBaseConfig(path) {
...
+      }
+
+      checkCalls()
+      apmServerMock.subscription.subscribe(checkCalls)
+    },
+    errorCount,
+    transactionCount
+  )
+
+  expect(serverCalls).toBeTruthy()
```
vigneshshanmugam: this should be moved inside the test, not inside utility function.

dev-utils/webdriver.js
vigneshshanmugam: Okay, now i get why we need it since we use the same var for both error and trans.

Let's refactor this code a bit.. checks for errors and transactions is repetitive in lots of places.

packages/rum-react/README.md
```diff
@@ -0,0 +1,8 @@
+# `rum-react`
+
+> TODO: description
```
vigneshshanmugam: Are we waiting on the API finalisation before we start working on the docs?

packages/rum-react/package.json
vigneshshanmugam: What if the user users some other router? In that case, we cannot say we cover all react apps.

packages/rum-react/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@elastic/apm-rum-react",
+  "version": "0.0.0",
+  "description": "> TODO: description",
+  "author": "Hamid <h.jahtalab@gmail.com>",
+  "homepage": "https://www.elastic.co/guide/en/apm/agent/js-base/current/index.html",
+  "license": "MIT",
+  "main": "src/index.js",
```
vigneshshanmugam: How are we planning to distribute the library? We need both `main` and `module` here and `cjs` files in main field.

packages/rum-react/babel.config.js
```diff
@@ -0,0 +1,43 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+module.exports = function(api) {
```
vigneshshanmugam: the configs for the build files should lie in a single place and gets used by fellow packages. Lets move this in to a central place and use it for both `apm-rum` and `apm-rum-react`. This creates too many configs and makes it hard for us to track changes.

packages/rum-react/test/e2e/index.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ */
+
+import { apm } from '@elastic/apm-rum'
+import { getGlobalConfig } from '../../../../dev-utils/test-config'
+import ApmServerMock from '../../../rum-core/test/utils/apm-server-mock'
+
+const { globalConfigs } = getGlobalConfig()
+
+function createApmBase(config) {
```
vigneshshanmugam: any reason why we don't want to use the `createApmBase` from `apm-rum`? Defeats the purpose of monorepo.

packages/rum-react/test/e2e/react/webpack.config.js
```diff
@@ -31,19 +31,39 @@ module.exports = {
...
   },
   module: {
     rules: [
       {
         test: /.jsx?$/,
+        exclude: /node_modules/,
         use: {
-          loader: 'babel-loader'
+          loader: 'babel-loader',
+          options: {
```
vigneshshanmugam: not required to have this config again here.

packages/rum/test/e2e/general-usecase/app.e2e-spec.js
```diff
@@ -129,4 +79,26 @@ describe('general-usercase', function() {
...
+      },
+      5000,
+      'expected element #test-element'
+    )
+
+    const serverCalls = waitForApmServerCalls(0, 1)
+    expect(serverCalls.sendTransactions.length).toBe(1)
+    const transactionPayload = serverCalls.sendTransactions[0].args[0][0]
+    expect(transactionPayload.name).toBe('Push state title')
+    expect(transactionPayload.spans.length).toBeGreaterThanOrEqual(3)
```
vigneshshanmugam: can we document what type of spans we can see here?.

packages/rum/test/e2e/general-usecase/app.js
```diff
@@ -111,4 +111,27 @@ generateError.tmp = 'tmp'
...
 testFetch(mockBackendUrl)
 
+if (location.hash === '#test-state') {
+  const path = location.pathname
+  history.pushState(
+    { data: 'test-state' },
+    'Push state title',
+    path + '#test-state'
+  )
+  history.pushState(
```
vigneshshanmugam: is this intentional to call it multiple times?

packages/rum-react/package.json
jahtalab: Yeah, I don't think we should say that we cover all routers, but it's more likely that users will come back to use with a feature request for those routers

packages/rum-react/README.md
jahtalab: Yeah, I will create a separate PR for that.

packages/rum/test/e2e/general-usecase/app.js
jahtalab: Yeah, because there's `go(-1)` call after that.

vigneshshanmugam: Can you change the arguments to 2nd call in that case? To make sure we are actually testing the right thing.

packages/rum-core/src/performance-monitoring/transaction.js
```diff
@@ -73,9 +73,16 @@ class Transaction extends SpanBase {
   }
 
   redefine(name, type, options) {
-    this.name = name
-    this.type = type
-    this.options = extend(this.options, options)
+    if (name) {
```
vigneshshanmugam: All these checks are repetitive and unnecessary if done right - check the PR here - https://github.com/elastic/apm-agent-rum-js/pull/294/

jahtalab: Here we don't have any default values, instead we are rejecting falsy values!

vigneshshanmugam: Ideally there shouldn't be any falsy values. It will be just the defaults if its falsy. Just having this check everywhere in the codebase is not solving anything to be honest. We do default to `Unknown` and `custom` for name and type everywhere.

dev-utils/webdriver.js
jahtalab: Let's refactor this separately!

packages/rum-core/src/performance-monitoring/transaction.js
jahtalab: Not sure how defaulting to `Unknown` and `custom`  is relevant here, but here we shouldn't set falsy values to these properties therefore either this method or it's caller should check for falsy values. The alternative is to have separate methods like `setName`, `setType` etc, don't think is any better. 

I'm open to suggestions though.

packages/rum/test/e2e/general-usecase/app.e2e-spec.js
jahtalab: Added comments to clarify!

packages/rum-react/package.json
jahtalab: Added the `build:main` script

packages/rum-react/src/get-with-transaction.js
jahtalab: Please see [here](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md#componentwrappedcomponent) for an explanation on `WrappedComponent`

packages/rum-react/test/e2e/react/webpack.config.js
jahtalab: I've tested it and it doesn't work without the options.

packages/rum-react/babel.config.js
jahtalab: I've refactored rum-react, but not `apm-rum` I think our babel setup needs a review/refactoring. but I think it's out of the scope of this PR.

jahtalab:
@vigneshshanmugam Would you please have another look?

packages/rum-react/babel.config.js
jahtalab: Created an issue : https://github.com/elastic/apm-agent-rum-js/issues/308

packages/rum-react/test/e2e/react/webpack.config.js
jahtalab: https://github.com/elastic/apm-agent-rum-js/issues/308

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -61,6 +61,11 @@ class PerformanceMonitoring {
       const payload = this.createTransactionPayload(tr)
       if (payload) {
         this._apmServer.addTransaction(payload)
+      } else {
```
vigneshshanmugam: can you add DEV flag for the debug log.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -140,7 +140,7 @@ class TransactionService {
         type,
         perfOptions
       )
-      tr.redefine(name, type, perfOptions)
+      tr.redefine(name, undefined, perfOptions)
```
vigneshshanmugam: can you explain and also add a comment on why we do set the argument to `undefined`.

packages/rum-react/karma.conf.js
```diff
@@ -0,0 +1,74 @@
...
+    devtool: 'source-map',
+    mode: 'development',
+    module: {
+      rules: [
+        {
+          test: /.jsx?$/,
+          exclude: /node_modules/,
+          use: {
+            loader: 'babel-loader',
+            options: {
```
vigneshshanmugam: you can pass configFile and pass the same one for rum-react/babel.config.js.
Example https://github.com/elastic/apm-agent-rum-js/blob/0ca1e5f97fd4204d6e93d099a9e10b01217bf920/dev-utils/karma.js#L70

packages/rum-react/package-lock.json
```diff
@@ -0,0 +1,21 @@
+{
```
vigneshshanmugam: lock files for packages can be removed.

packages/rum-react/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@elastic/apm-rum-react",
+  "version": "0.0.0",
+  "description": "Elastic APM RUM React integration",
```
vigneshshanmugam: Just to keep it consistent with our apm-rum,  We can rename it as `Elastic APM JavaScript agent for React Applications`?  WDYT?

packages/rum-react/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "registry": "https://registry.npmjs.org"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/elastic/apm-agent-rum-js.git"
+  },
+  "scripts": {
+    "prepublishOnly": "npm run build",
+    "script": "node ../../dev-utils/run-script.js",
+    "build": "webpack && npm run build:main",
```
vigneshshanmugam: can we keep it consistent with apm-rum? 

```
"prepublishOnly": "npm run build && npm run build:main",
"build": "webpack",
```

packages/rum-react/src/get-with-transaction.js
vigneshshanmugam: Have you checked about this scenario?

packages/rum-react/test/e2e/react/main-component.jsx
```diff
@@ -25,33 +25,36 @@
 
 import React from 'react'
 
-import { apm } from './rum'
-
 class MainComponent extends React.Component {
   constructor(props, state) {
     super(props, state)
+    console.log(props)
```
vigneshshanmugam: nit:  remove console.log

packages/rum-react/test/e2e/react/webpack.config.js
vigneshshanmugam: Yeah sounds good,  But lets do this suggestion for now.  https://github.com/elastic/apm-agent-rum-js/pull/265/files#r300721927

packages/rum/test/e2e/general-usecase/app.e2e-spec.js
```diff
@@ -129,4 +79,29 @@ describe('general-usercase', function() {
...
+    )
+
+    const serverCalls = waitForApmServerCalls(0, 1)
+    expect(serverCalls.sendTransactions.length).toBe(1)
+    const transactionPayload = serverCalls.sendTransactions[0].args[0][0]
+    expect(transactionPayload.name).toBe('Push state title')
+    /**
+     * The actual spans are tested as part of the previous test.
+     */
+    expect(transactionPayload.spans.length).toBeGreaterThanOrEqual(3)
```
vigneshshanmugam: Does this mean we will see the transaction in the UI as page-load transaction type and also see Resource and Nav timing spans as part of the History API route change?

jahtalab: Yes, that's correct! I think it makes sense to look at the first route-change as an extension of the page-load therefore showing the continuity of the application.

packages/rum-react/package.json
jahtalab: I think we should change the other ones to have only one `build` script.

vigneshshanmugam: Sure,  I am fine with that change.

packages/rum-react/karma.conf.js
jahtalab: Let's discuss this in #308, I'm still not sure how much benefit we will gain from this. Specially since in each of the babel files we need to change the config.

packages/rum-react/package.json
jahtalab: I've change it a little bit, I think we should include "Real User Monitoring" in some form in the description. We should fix apm-rum as well.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: I've added some comments in the code!

jahtalab:
@vigneshshanmugam, I've addressed the review comments and tests should be passing now, would you please have another look?

packages/rum-react/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "files": [
+    "src",
+    "dist"
+  ],
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/elastic/apm-agent-rum-js.git"
```
vigneshshanmugam: @jahtalab Lets also add directory field here to find relevant readme when publishing, Missed in the review.

PR #270, opened 2019-05-23
feat(rum): deprecate addTags in favor of addLabels

### LINKED ISSUE #215: Rename addTags to addLabels
- [x] remove `addTags` in major version.
- [x] Update docs

Related https://github.com/elastic/apm/issues/42

vigneshshanmugam:
+ first step of elastic/apm-agent-rum-js#215 
+ Opentracing `addTags` still works and under the hood calls span.addLabels for adding labels.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -68,7 +68,7 @@ TIP: Before using custom context, ensure you understand the different types of
 [[apm-add-tags]]
 ==== `apm.addTags()`
 
-deprecated, Please use <<apm-add-labels,`apm.addLabels()`>> instead
+deprecated[4.1.x,Tags have been replaced with labels], Please use <<apm-add-labels,`apm.addLabels()`>> instead
 
 [float]
 [[apm-add-labels]]
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -68,16 +68,22 @@ TIP: Before using custom context, ensure you understand the different types of
 [[apm-add-tags]]
 ==== `apm.addTags()`
 
+deprecated, Please use <<apm-add-labels,`apm.addLabels()`>> instead
+
+[float]
+[[apm-add-labels]]
+==== `apm.addLabels()`
+
 [source,js]
 ----
-apm.addTags({ [name]: value })
+apm.addLabels({ [name]: value })
 ----
 
-Set tags on transactions and errors.
+Set labels on transactions and errors.
 
-Tags are key/value pairs that are indexed by Elasticsearch and therefore searchable (as opposed to data set via `setCustomContext()`). You can set multiple tags.
+Labels are key/value pairs that are indexed by Elasticsearch and therefore searchable (as opposed to data set via <<apm-set-custom-context,`apm.setCustomContext()`>>). You can set multiple labels.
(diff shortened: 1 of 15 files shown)


packages/rum-core/src/common/utils.js
```diff
@@ -135,14 +135,14 @@ function isPlatformSupported() {
...
- * Convert values of the tag to be string to be compatible
+ * Convert values of the Tag/Label to be string to be compatible
  * with the apm server prior to <6.7 version
  *
  * TODO: Remove string conversion in the next major release since
  * support for boolean and number in the APM server has landed in 6.7
  * https://github.com/elastic/apm-server/pull/1712/
  */
-function setTag(key, value, obj) {
+function setLabel(key, value, obj) {
```
vigneshshanmugam: the java agent has added support for boolean/number values in the labels payload along with deprecating `addTags`. Should we also do that as part of this PR? @jahtalab /cc @felixbarny

docs/agent-api.asciidoc
```diff
@@ -68,16 +68,22 @@ TIP: Before using custom context, ensure you understand the different types of
 [[apm-add-tags]]
 ==== `apm.addTags()`
 
+deprecated, Please use <<apm-add-labels,`apm.addLabels()`>> instead
```
bmorelli25: I think we should add a little more clarification here. We also have deprecated tags, the format is: `deprecated[x.x.x]`. Something like:
> deprecated[1.2] Tags have been replaced with labels. Please use <<apm-add-labels,`apm.addLabels()`>> instead.

vigneshshanmugam: Nice, Will change it accordingly :)

packages/rum-core/src/common/utils.js
jahtalab: I think adding support for boolean/number is a breaking change since it those values were not supported by APM server prior to 6.7 release. IMO we should do them in a major version.

jahtalab: I've created [this issue](https://github.com/elastic/apm-agent-rum-js/issues/272) to address this!

vigneshshanmugam: Cool, let's do it as part of 5.0.x then

packages/rum-core/src/performance-monitoring/span-base.js
```diff
@@ -48,16 +48,14 @@ class SpanBase {
     }
   }
 
-  addTags(tags) {
+  addLabels(labels) {
```
jahtalab: Maybe we should deprecated `addTags` in this minor version and remove it in the next major. WDYT?

eliaskardel:
image

I am running the latest version of all components elastic search, Kibana, APM Server, RUM JS agent. 

The addLabel() function doesnt give me anything on the other side. I switched on debug in the agent and saw my labels in the developer console. However they seem to not end up in eleastic search.

eliaskardel:
const transaction = apm.startTransaction('transaction1', 'custom')
                const httpSpan = transaction.startSpan('span2', 'custom')
                httpSpan.addLabels({'test_label' : 'test_label'})
                httpSpan.end()
                transaction.end()

PR #271, opened 2019-05-27
chore(rum): enable extended debugging and console logs for e2e tests

### LINKED ISSUE #266: Capture console logs when running tests
Failures such as [this](https://github.com/elastic/apm-agent-rum-js/issues/240) are quite difficult to debug since it only happens occasionally on saucelabs (not locally) and to make things more complicated saucelabs doesn't capture console logs. 

We can implement some of these solutions to make debugging a bit easier:
- We can print console logs into the DOM, they would show up in the video of the running test captured by saucelabs. This even works in legacy browser.
- We can capture logs and store them in elasticsearch along with build and test session identifiers so that they can be search discovered quickly. We should be careful not to interfere with the running test.

vigneshshanmugam:
+ fix #266 
+ We already print console logs when test fails, but to support older drivers/browsers, now extendedDebugging is enabled. Incase the browser logs are not printed, We can do a curl request to the below API to print to figure out the logs. 
```sh
curl https://<username>:<password>@eds.saucelabs.com/<testId>/console.json
```

+ set logging preferences for browser logs
+ do not truncate logs when process exits.

### DIFF
--- a/dev-utils/test-config.js
+++ b/dev-utils/test-config.js
@@ -85,7 +85,8 @@ function getBrowserList() {
   return [
     {
       browserName: 'chrome',
-      version: '49'
+      version: '49',
+      extendedDebugging: true
     },
     {
       browserName: 'chrome',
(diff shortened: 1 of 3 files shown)


vigneshshanmugam:
@jahtalab Yes its a WIP, i added the tag in progress :)

PR #273, opened 2019-05-27
fix(rum-core): capture all spans as part of page-load transaction

### LINKED ISSUE #240: figure out why e2e tests fails occasionally on chrome 49
The E2E general-usecase tests has some race condition when running on chrome 49. Probably its a browser bug, but we need to find out why it happens occasionally. 

Example travis job. - https://travis-ci.org/elastic/apm-agent-rum-js/jobs/512143864

```
[test:e2e:supported] [chrome 49 #0-0] Running: chrome (v49)
[test:e2e:supported] [chrome 49 #0-0]
[test:e2e:supported] [chrome 49 #0-0] general-usercase
[test:e2e:supported] [chrome 49 #0-0]    ✖ should run the general usecase
[test:e2e:supported] [chrome 49 #0-0]
[test:e2e:supported] [chrome 49 #0-0] 1 failing (3.4s)
[test:e2e:supported] [chrome 49 #0-0]
[test:e2e:supported] [chrome 49 #0-0] 1) general-usercase should run the general usecase
[test:e2e:supported] [chrome 49 #0-0] Expected 2 to be greater than 2.
[test:e2e:supported] [chrome 49 #0-0] Error: Expected 2 to be greater than 2.
[test:e2e:supported] [chrome 49 #0-0]     at <Jasmine>
[test:e2e:supported] [chrome 49 #0-0]     at UserContext.<anonymous> (/home/travis/build/elastic/apm-agent-rum-js/packages/rum/test/e2e/general-usecase/app.e2e-spec.js:105:45)
[test:e2e:supported] [chrome 49 #0-0]     at <Jasmine>
[test:e2e:supported] [chrome 49 #0-0] Expected undefined to be defined.
[test:e2e:supported] [chrome 49 #0-0] Error: Expected undefined to be defined.
[test:e2e:supported] [chrome 49 #0-0]     at <Jasmine>
[test:e2e:supported] [chrome 49 #0-0]     at UserContext.<anonymous> (/home/travis/build/elastic/apm-agent-rum-js/packages/rum/test/e2e/general-usecase/app.e2e-spec.js:109:18)
[test:e2e:supported] [chrome 49 #0-0]     at <Jasmine>
[test:e2e:supported] [chrome 49 #0-0]
[test:e2e:supported] [chrome 49 #0-0] Check out job at https://app.saucelabs.com/tests/bdb6fcb9d0744d3d92d6bb0e214f3e5d

```

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#240 

+ For page load transactions, we have to adjust the transaction start/end depending on the navigation & resource timing spans.Deciding it as part of transaction.end would  would exclude the navigation and resource timing spans since we create the nav and res spans later after the load event has happened and we might consider them invalid since transaction.end would be lesser than those spans. 

+ I am not sure how we missed this, But this is a huge issue if the user has any custom/external spans that is finished before the page-load event fires.


packages/rum-core/src/performance-monitoring/transaction.js
```diff
@@ -132,8 +134,18 @@ class Transaction extends SpanBase {
...
+     * For page load transactions, we have to adjust the transaction start/end
+     * depending on the navigation & resource timing spans.
+     *
+     * Doing it here would exclude the navigation and resource timing spans since
+     * transaction.end might be lesser than those spans end and
+     * it would be considered as invalid span
+     */
+    if (this.type !== PAGE_LOAD) {
+      this._adjustStartToEarliestSpan()
+      this._adjustEndToLatestSpan()
```
jahtalab: We still call these two in the `capture-navigation-timing.js`, not sure why this would solve the issue.
In the capture-navigation-timing we just add the spans and then adjust the start/end of the transaction to be compatible with the spans that were added for resource and navigation.

vigneshshanmugam: The main point is to do it only in https://github.com/elastic/apm-agent-rum-js/blob/c162a1252f034c2ec59d5c5ce349412857d817aa/packages/rum-core/src/performance-monitoring/capture-hard-navigation.js#L217-L218 for the page-load transaction after we have calculated the nav timing spans. 

Doing it on the `transacton.js` prevents the nav/res timing spans from getting added which is the issue that we are fixing.

vigneshshanmugam:
Docs build failure is unrelated.

jahtalab:
@vigneshshanmugam , I'm still not sure why this bug only existed on chrome 49 (sorry if I'm overlooking something here).

vigneshshanmugam:
Sorry if I did not explain it properly in the description/discussion, The but exists in all browsers. The number of spans that should exist in page load is 8. We we're seeing 4 in other browsers l, whereas in chrome 49 we see only 2 sometimes..Most of the ajax/nav timing spans fails to get captured bcoz of the bug which is fixed now. The spans is not still 8 because sometimes the spans like dns/load might be 0 since the due to network cache and other behaviours. So the test is reasonably adjusted to check for more than 4 instead of the previous value 2. 

If you want more details, you can print the console logs before and after this fix. Please watch the number of spans getting added to the transactions.

Let me know if you need more info.

vigneshshanmugam:
To give more context, Please check the below console logs for before and after the fix. 

## Tested on Chrome 74. 

### Before 

Screenshot 2019-06-05 at 12 52 24

### After

Screenshot 2019-06-05 at 12 58 22

jahtalab:
Thanks @vigneshshanmugam! I was thinking about how we can refactor this part to avoid this kind of bugs! At the moment it seems we’re introducing a special case into the transaction class which until now didn't have any specific code about transaction type.

jahtalab:
Had a meeting with @vigneshshanmugam and we decided to refactor adjusting the transaction to it's spans into the PerformanceMonitoring class!

vigneshshanmugam:
@jahtalab The refactor is done and I have moved the tests around closer to the functionality of the class. Also added extra guard for the general usecase test to catch bugs.

packages/rum-core/test/performance-monitoring/transaction.spec.js
```diff
@@ -168,31 +125,6 @@ describe('transaction.Transaction', function() {
     expect(span).toBe(undefined)
   })
 
-  it('should not produce negative durations while adjusting to the spans', function() {
```
jahtalab: Shouldn't we move this test to performance monitoring as well?

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -261,13 +261,68 @@ describe('PerformanceMonitoring', function() {
...
+    expect(span.type).toContain('.truncated')
+  })
+
   it('should create correct payload', function() {
     var tr = new Transaction('transaction1', 'transaction1type', {
       transactionSampleRate: 1
     })
     var span = tr.startSpan('span1', 'span1type')
     span.end()
-    span._end += 10
```
jahtalab: I think this was added to avoid having the span duration to be 0 since we start and end the span quickly.

vigneshshanmugam: Yeah I had the same impression, It makes the span truncated since transaction end is lesser. But I can call performance.adjustTime and pass the test. We pay some time for creating the span so I thought its reasonable to remove it .WDYT?

packages/rum-core/test/performance-monitoring/transaction.spec.js
vigneshshanmugam: Oh yeah might have missed, Let me add it

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
jahtalab: But `performanceMonitoring.createTransactionPayload` calls the adjustTime!

Also, it's ok to change the tr._end after `expect(tr._end).toBeDefined()` to make sure that it includes the span!

vigneshshanmugam: Did the changes.

PR #274, opened 2019-05-29
fix(rum-core): remove sensitive info from span context

### LINKED ISSUE #252: Redact sensitive information from the http url
Currently, the username and password can be leaked into the APM data when we capture http urls, e.g. http://username:password@@example.com.

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#252 
+ Removes the username:password fields from URL on XHR calls

### DIFF
--- a/packages/rum-core/src/common/url.js
+++ b/packages/rum-core/src/common/url.js
@@ -45,6 +45,14 @@
  *
  */
 
+/**
+ * Order of the RULES are very important
+ *
+ * RULE[0] -> for checking the index of the character on the URL
+ * RULE[1] -> key to store the associated value present after the RULE[0]
+ * RULE[2] -> Extract from the front till the last index
+ * RULE[3] -> Left over values of the URL
+ */
 const RULES = [
   ['#', 'hash'],
   ['?', 'query'],
@@ -74,6 +82,14 @@ class Url {
         if (~index) {
           const instLength = instruction[2]
           if (instLength) {
+            /**
+             * we need to figure out the explicit index where the auth portion
+             * in the host ends before parsing the rest of the URL as host.
+             *
+             * ex: http://a@b@c.com/d
+             * auth -> a@b
+             * host -> c.com
+             */
             let newIndex = address.lastIndexOf(parse)
             index = Math.max(index, newIndex)
             this[key] = address.slice(0, index)
--- a/packages/rum-core/src/common/url.js
+++ b/packages/rum-core/src/common/url.js
@@ -115,6 +115,8 @@ class Url {
       if (instruction[3]) this[key] = this[key].toLowerCase()
     }
 
+    this.relative = relative
+
     this.protocol = protocol || location.protocol || ''
 
     this.origin =
--- a/packages/rum-core/src/common/url.js
+++ b/packages/rum-core/src/common/url.js
@@ -49,7 +49,8 @@ const RULES = [
   ['#', 'hash'],
   ['?', 'query'],
   ['/', 'path'],
-  [NaN, 'host', 1] //
+  ['@', 'auth', 1],
+  [NaN, 'host', undefined, 1] //
 ]
 const PROTOCOL_REGEX = /^([a-z][a-z0-9.+-]*:)?(\/\/)?([\S\s]*)/i
 
@@ -71,8 +72,16 @@ class Url {
       if (typeof parse === 'string') {
         index = address.indexOf(parse)
         if (~index) {
-          this[key] = address.slice(index)
-          address = address.slice(0, index)
+          const instLength = instruction[2]
+          if (instLength) {
+            let newIndex = address.lastIndexOf(parse)
+            index = Math.max(index, newIndex)
+            this[key] = address.slice(0, index)
+            address = address.slice(index + instLength)
+          } else {
+            this[key] = address.slice(index)
+            address = address.slice(0, index)
+          }
         }
       } else {
         /** NaN condition */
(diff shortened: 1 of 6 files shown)


packages/rum-core/test/common/utils.spec.js
```diff
@@ -291,4 +291,15 @@ describe('lib/utils', function() {
     )
     expect(utils.removeInvalidChars('invalid"')).toEqual('invalid_')
   })
+
+  it('should remove sensitive auth from url', () => {
+    expect(utils.removeAuthFromUrl('https://a:b@c.com/d')).toEqual(
+      'https://c.com/d'
+    )
+    expect(utils.removeAuthFromUrl('http://e:f@example.com/g')).toEqual(
```
axw: perhaps in this one you could put "e:f@" in the path, to show that only the userinfo is removed?

vigneshshanmugam: Is it common to have user info in the path? We already remove the query params which might have sensitive info..

axw: No sorry if I was unclear, my point was that if there's by coincidence the same string in the path, it shouldn't be removed. That's a weird/pathological case, a more realistic one would be that there's a `<some string>@<some other string>` in the path that does not match the userinfo, which should not be modified.

I see you're handling that in your regex, I just think it's a good idea to have a test to make sure that doesn't break in the future.

vigneshshanmugam: Got it now. thanks for the explanation 👍 Will add a test for the scenario.

packages/rum-core/src/common/utils.js
```diff
@@ -324,6 +324,22 @@ function removeInvalidChars(key) {
   return key.replace(/[.*"]/g, '_')
 }
 
+/**
+ * Removes the username:password portion from Urls
+ *
+ * Ex: https://a:b@example.com/c  -> https://example.com/c
```
jahtalab: Maybe we should replace the redacted parts with `[REDACTED]`, e.g. 

`https://[REDACTED]:[REDACTED]@example.com/c`

I think the node agent does this, I'm not sure if they use the exact format of `[REDACTED]` though.

vigneshshanmugam: Browsers convert the request of this form `https://a:b@example.com/c  -> https://example.com/c` while using XHR. Its better to align with browser and get rid of the userpass information than folllowing Node.

vigneshshanmugam:
@jahtalab 

1. We are not doing URL parsing as apposed to the library which is doing exactly that and it should take care of the edge cases like the one you mentioned.
2. Browsers while constructing the request does take care of all the url issues that you mentioned in your review comment. We don't need to cover all those edge cases, It's simply not required. You can try with a XMLHttpRequest in the console and can see how its getting converted for us.

axw:
@jahtalab thanks for the links, I'll add those to the meta-issue. We should probably implement some common test cases.

@vigneshshanmugam In the case of `"http://a@b?@c"` won't your code replace `"a@b?@"` with `""`, thus ending up with `"http://c"`? Why don't we need to handle this case?

vigneshshanmugam:
@axw Yes It does indeed convert it, But isn't it wrong? It would be `https://b/?@c`. Reason why i was suggested we do not need to do it, is because of how common these patterns exist in the user application. Probably i am wrong, But I thought of keeping the functionality simple and go for most common use cases instead of handling everything.

axw:
@vigneshshanmugam IMO, if it's sensitive data, then we should handle everything.

The regex you have defined is greedy. You could use the lazy/non-greedy variant `+?` to resolve that, but then `"http://a@b@c/"` wouldn't be handled correctly.

jahtalab:
@vigneshshanmugam , We read the URL raw from the `open` method call which we patch therefore any fix that browsers do, is not useful for us.

I think we should use the polyfill test-cases to see which edge cases we don't cover. Then, if a we decide not to handle an edge-case, we should document it in the code (maybe even using the test-case for that)

packages/rum-core/src/common/utils.js
jahtalab: I think this is different compared to what browsers do. We're using the URL here to show as the span name. We should show something as close to what they used in their code as possible. In other words, It is important to know whether the username and password were redacted or they simply were not there in the first place. If the user only sees `https://example.com/c` they can't know if there was any username and password.

vigneshshanmugam: Done

vigneshshanmugam:
@jahtalab I have updated our custom Url parser to handle all those cases. have a look again and let me know your thoughts.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -87,19 +88,21 @@ class PerformanceMonitoring {
           var taskId = transactionService.addTask()
 
           if (span) {
```
jahtalab: I think we should redact the url used for the span name as well.

packages/rum-core/src/common/url.js
```diff
@@ -82,12 +91,12 @@ class Url {
        * Default values for all keys from location if url is relative
        */
       this[key] =
-        this[key] || (relative && instruction[2] ? location[key] || '' : '')
+        this[key] || (relative && instruction[3] ? location[key] || '' : '')
```
jahtalab: I think these kind of changes are might be hard to follow, if we ever add another rule again, but maybe we can refactor at another time.

vigneshshanmugam: I will add more documentation to the code to make it easier to follow.

vigneshshanmugam:
@jahtalab good catch on the span name part. I will update the PR.

PR #276, opened 2019-05-29
feat(rum-core): add user timing spans to the page-load transaction

### LINKED ISSUE #149: support user timing measures as part of page load transaction
+ Since we already capture important information of the page load using nav timing and resource timing API, its would be a nice feature to include User Timing (https://developer.mozilla.org/en-US/docs/Web/API/User_Timing_API#Performance_measures), that are captured by the developers through code and we can display them in the waterfall UI along with the other metrics.

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#149


nik9000:
The docs-build failure is my fault. I broke a thing. I've reverted the offending commit which should un-break it.

@elasticmachine, run docs build.

nik9000:
My favorite is how to write a `'` inside of a single quoted string in bash: `'"'"'`. Take that, perl!

nik9000:
> My favorite is how to write a `'` inside of a single quoted string in bash: `'"'"'`. Take that, perl!

Woops, wrong issue.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -168,6 +172,27 @@ function createResourceTimingSpans(entries, filterUrls) {
...
+function createUserTimingSpans(entries) {
+  const userTimingSpans = []
+
+  for (let i = 0; i < entries.length; i++) {
+    const { name, startTime, duration } = entries[i]
+
+    if (duration <= USER_TIMING_THRESHOLD) {
+      continue
+    }
+    const kind = 'user.measure'
```
vigneshshanmugam: Trying to figure out if this type and subType info for User timing makes sense, Thoughts? @jahtalab

vigneshshanmugam: Since its part of the application code, I am learning towards keeping type as `app.custom`  or `custom.app`. All the custom instrumentations in other agents use something similar with type set to custom.

jahtalab: Decided to set it to `app` for now, we can add subType later if needed!

packages/rum-core/src/common/constants.js
```diff
@@ -60,6 +60,10 @@ const RESOURCE_INITIATOR_TYPES = [
  * Maximum duration of the span that is used to decide if the span is valid - 300 secs / 5 mins
  */
 const SPAN_THRESHOLD = 5 * 60 * 1000
+/**
+ * Check only for long tasks that are more than 60ms
+ */
+const USER_TIMING_THRESHOLD = 60
```
jahtalab: Can there be a lot of small tasks that eventually slows down the app? we have a method `groupSmallContinuouslySimilarSpans` in PerformanceMonitoring that groups small spans into one span if they repeat continuously!

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -168,6 +172,27 @@ function createResourceTimingSpans(entries, filterUrls) {
...
+    const end = startTime + duration
+    const span = new Span(name, kind)
+    span._start = startTime
+    span.ended = true
```
jahtalab: Maybe call the `span.end` before this to make sure any changes in `span.end` would be reflected here as well.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -208,11 +233,18 @@ function captureHardNavigation(transaction) {
...
+        span
+      ) {
         if (isValidSpan(transaction, span)) {
           transaction.spans.push(span)
         }
       })
+
+      const userEntries = perf.getEntriesByType('measure')
+      createUserTimingSpans(userEntries).forEach(
+        span => span._end <= transaction._end && transaction.spans.push(span)
```
jahtalab: Let's apply the same validation done in #277!

packages/rum-core/test/performance-monitoring/capture-hard-navigation.spec.js
```diff
@@ -239,6 +241,25 @@ describe('navigationTiming', function() {
     expect(spans.map(mapSpan)).toEqual(spanSnapshot)
   })
 
+  it('should createUserTimingSpans', function() {
+    const spans = createUserTimingSpans(userTimingEntries)
+    expect(spans.length).toEqual(2)
+    expect(spans).toEqual([
+      jasmine.objectContaining({
+        name: 'measure_1',
```
jahtalab: Maybe add the type here as well!

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -289,11 +290,13 @@ describe('PerformanceMonitoring', function() {
...
     window.performance.getEntriesByType = function(type) {
-      expect(['resource', 'paint']).toContain(type)
+      expect(['resource', 'paint', 'measure']).toContain(type)
       if (type === 'resource') {
         return resourceEntries
+      } else if (type === 'paint') {
+        return paintEntries
       }
-      return paintEntries
+      return userTimingEntries
```
jahtalab: Maybe wrap this in an `else if` as well and return `[]` by default!

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -245,23 +243,12 @@ describe('TransactionService', function() {
   })
 
   it('should capture resources from navigation timing', function(done) {
-    var _getEntriesByType = window.performance.getEntriesByType
-
-    window.performance.getEntriesByType = function(type) {
```
jahtalab: Why we these need to be removed? since TransactionService make use of capture navigation timing

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -313,14 +300,6 @@ describe('TransactionService', function() {
       pageLoadSampled: true
     })
 
-    window.performance.getEntriesByType = function(type) {
```
jahtalab: Same as above!

packages/rum-core/src/common/constants.js
vigneshshanmugam: Well these are profiled code that user is doing in the application at different times while the application is loading and name of the spans are different for each perfEntry which ideally skips the similar span check.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
vigneshshanmugam: I will rebase and add once the other PR is merged.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
vigneshshanmugam: test basically doesn't do any checks on the spans, just patching the window function and returning some data for no reason is not useful IMO. Let me put this way, why are there entries useful here?

packages/rum-core/src/common/constants.js
jahtalab: Why the span names would be different each time?

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
jahtalab: The test is "should capture resources from navigation timing" in other words we want to test whether TransactionService captures navigation timings and for that to work we need the entries! Maybe we should addd extra checks here to make sure navigation timings are captured correctly.

packages/rum-core/src/common/constants.js
vigneshshanmugam: `performance.measure("foo"), performance.measure("bar")`, this is how the entries will be added by the user. We are not capturing marks, only the measures.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
vigneshshanmugam: I will add more checks to the test then to make sure we are actually capturing.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -168,6 +172,27 @@ function createResourceTimingSpans(entries, filterUrls) {
...
+  for (let i = 0; i < entries.length; i++) {
+    const { name, startTime, duration } = entries[i]
+
+    if (duration <= USER_TIMING_THRESHOLD) {
+      continue
+    }
+    const kind = 'user.measure'
+    const end = startTime + duration
+    const span = new Span(name, kind)
+    span._start = startTime
```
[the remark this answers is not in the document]
jahtalab: relevant issue: https://github.com/elastic/apm-agent-rum-js/issues/307

jahtalab:
Blocked by #277

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -152,11 +153,16 @@ describe('TransactionService', function() {
...
-      expect(['resource', 'paint']).toContain(type)
+      expect(['resource', 'paint', 'measure']).toContain(type)
       if (type === 'resource') {
         return resourceEntries
+      } else if (type === 'paint') {
+        return paintEntries
+      } else if (type === 'measure') {
+        return userTimingEntries
+      } else {
+        return []
```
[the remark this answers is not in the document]
vigneshshanmugam: Did the refactoring in this PR itself.

PR #278, opened 2019-06-03
chore(rum-core): remove functions from transaction-service

vigneshshanmugam:
+ removed dead code from transaction service `sendPageLoadMetrics` which was only used in tests . and also refactored the async code. 
+ removed functions related to how we run async code with function context passed around, Not necessary anymore with arrow functions. Helps new contributors understand the code easier. 
+ Moved transaction name and type as constants. 
+ Extracted from other clean up PR to make review easier.

### DIFF
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -105,7 +105,7 @@ class TransactionService {
   }
 
   createPerfOptions(options) {
-    var config = this._config.config
+    const config = this._config.config
     return extend(
       {
         pageLoadTraceId: config.pageLoadTraceId,
@@ -170,8 +170,8 @@ class TransactionService {
     this._logger.debug('TransactionService.startTransaction', tr)
 
     tr.onEnd = () => {
-      return Promise.resolve()
-        .then(() => {
+      return Promise.resolve().then(
+        () => {
           this._logger.debug('TransactionService transaction finished', tr)
           if (this.shouldIgnoreTransaction(tr.name)) {
             return
@@ -190,8 +190,9 @@ class TransactionService {
           } else {
             this.add(tr)
           }
-        })
-        .catch(err => this._logger.debug(err))
+        },
+        err => this._logger.debug(err)
+      )
     }
     return tr
   }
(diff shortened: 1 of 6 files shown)


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -190,48 +164,36 @@ class TransactionService {
       if (perfOptions.pageLoadSampled) {
         tr.sampled = perfOptions.pageLoadSampled
       }
-
-      if (tr.name === 'Unknown' && perfOptions.pageLoadTransactionName) {
```
jahtalab: Shouldn't we set the transaction name here? I think we should set the transaction name as early as possible, I know we also try to set the name onEnd but they are both necessary.

vigneshshanmugam: Yeah I felt there was no reason to do the same thing twice, I cannot think of any scenario where we might fail to update the page load transaction name. Can you think of a scenario where this would be valid?

jahtalab: For example if the users try to get the current transaction based before the transaction has ended, they would see the wrong name. Also if we decide to influence sampling based on the transaction name (e.g. sampling higher rate for important transactions) since we make the sampling decision at the start of transaction we should also set the name.

vigneshshanmugam: Hmmm, yeah sounds good. I will just add a comment in code mentioning the same.

vigneshshanmugam:
Rebased with latest changes.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -165,31 +137,6 @@ describe('TransactionService', function() {
...
-      expect(tr.isHardNavigation).toBe(true)
-      expect(tr.marks.agent).toBeDefined()
-      expect(tr.marks.navigationTiming).toBeDefined()
-      done()
-    })
-    var tr = transactionService.sendPageLoadMetrics('test')
-
-    transactionService = new TransactionService(logger, config)
-    var zoneTr = new Transaction('test-name', 'test-type', {
-      canReuse: true
```
jahtalab: I think we should still keep this test because of the `canReuse` but maybe refactor it to use startTransaction instead of `sendPageLoadMetrics`

vigneshshanmugam: Done.

PR #279, opened 2019-06-03
feat: add event listeners for transactions

jahtalab:
This PR adds the `onTransactionStart`  and `onTransactionEnd`. 
These events can be used to change any property of the transaction.

- [x] Add EventHandler tests


jahtalab:
@vigneshshanmugam , I made this PR as a draft to discuss the event solution for fixing transaction name. it is still WIP so no need for a detail review.

jahtalab:
@vigneshshanmugam , I have simplified the implementation! Would you please have another look?

packages/rum/src/apm-base.js
```diff
@@ -101,9 +105,17 @@ class ApmBase {
...
-    configService.addEventListener(name, fn)
+    configService.events.observe(name, fn)
+  }
+
+  onTransactionStart(fn) {
+    this.observe(ON_TRANSACTION_START, fn)
+  }
+
+  onTransactionEnd(fn) {
+    this.observe(ON_TRANSACTION_END, fn)
```
jahtalab: Decided to remove these to reduce the bundle size!

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -178,6 +179,7 @@ class TransactionService {
     if (__DEV__) {
       this._logger.debug('TransactionService.startTransaction', tr)
     }
+    this._config.events.send(ON_TRANSACTION_START, [tr])
```
vigneshshanmugam: even though I understand the idea behind starting the event here, We have to think about the usecases like this
+ What if the user/us plans to use start and end events to determine the timing of the transaction

We can call the start ASAP but it might have few params not defined but I think that is not a huge issue. 


We are ending the transaction at line https://github.com/elastic/apm-agent-rum-js/pull/279/files#diff-074367311a4235f7774f1d58a437e7b6L157, when the transaction cannot be redefined. Should call onEnd that time as well.

packages/rum-core/src/common/event-handler.js
```diff
@@ -0,0 +1,69 @@
...
+          fn.apply(undefined, args)
+        } catch (error) {
+          console.log(error, error.stack)
+        }
+      })
+    }
+  }
+
+  send(name, args) {
+    this.sendOnly(name + BEFORE_EVENT, args)
```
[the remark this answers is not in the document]
jahtalab: The `after` event is used in performance monitoring class [here](https://github.com/elastic/apm-agent-rum-js/pull/279/files#diff-2e5b6c0e19c5f2a74a484bce5be88fa4R63). The reason being that we need to send the transaction to APM server **after** all event listeners for transaction end have finished!

The before event is not used currently, it doesn't have much overhead (a function call which returns immediately since no event listeners exist for the before event) but I'm ok with removing it as well.

packages/rum/src/apm-base.js
```diff
@@ -101,6 +101,11 @@ class ApmBase {
     return !this._disable
   }
 
+  observe(name, fn) {
+    const configService = this.serviceFactory.getService('ConfigService')
+    configService.events.observe(name, fn)
```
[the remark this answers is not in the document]
jahtalab: I think the event handler should be placed in a common place, since we can use it to listen to and send any event, whether related to transactions or not (e.g for span events or errors as you mentioned). 

Taking a step back, I think we need a common place to put these top level functions, whether it should be config service or not I don't have a strong opinion on. it might also be a naming issue, i.e. if we refactor `ConfigService` to be `AgentService`, that too is a fine approach IMO.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: We need to run our "fixes" for the transaction first before running the event listeners. we can not run them afterwards, since we might overwrite the user's modifications.

packages/rum/src/apm-base.js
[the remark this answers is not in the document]
jahtalab: Decided to introduce `AgentService` in a separate PR.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: The Transaction.onEnd is called when we end the transaction, therefore broadcasting the transaction-end event.

jahtalab:
The test failure is in docs (not related to these changes)

packages/rum-core/src/common/event-handler.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  sendOnly(name, args) {
+    const obs = this.observers[name]
+    if (obs) {
+      obs.forEach(fn => {
+        try {
+          fn.apply(undefined, args)
+        } catch (error) {
+          console.log(error, error.stack)
```
vigneshshanmugam: We have the same logic in subscription which IMO shouldn't be there. We should not capture errors from user's code. If there is a broken behaviour that should be reported as thrown as error.

jahtalab: I don't an error in the event listener, should be throw! That prevents other listeners from running, and might even affect the user's application! We should report it here since it helps with debugging!

vigneshshanmugam: Decided to do this after having a separate discussion.

Issue #267, opened 2019-05-21
Use window.addEventListener instead of window.onerror


palci12:
Hi. We use apm-agent in many different applications. Some have their own onError listener in window.onerror. When apm-agent starts it overwrites previous onError listener and vice-versa. 
Isn't it better to use window.addEventListener('error', fn) instead of window.onerror? 
Maybe make it configurable?

vigneshshanmugam:
Thanks for creating the issue, Since 4 we dropped the support for IE 10 and lesser. Its safe to use addEventListener as suggested. Have opened the PR for the same.

PR #281, opened 2019-06-05
feat(rum-core): use error event instead of global onerror method

vigneshshanmugam:
+ fixes https://github.com/elastic/apm-agent-rum-js/issues/267

### DIFF
--- a/packages/rum-core/test/error-logging/error-logging.spec.js
+++ b/packages/rum-core/test/error-logging/error-logging.spec.js
@@ -174,9 +174,10 @@ describe('ErrorLogging', function() {
       .then(() => done())
   })
 
-  it('should install onerror and accept ErrorEvents', function(done) {
+  it('should install global listener for error and accept ErrorEvents', function(done) {
     var count = 0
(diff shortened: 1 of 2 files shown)


packages/rum-core/test/error-logging/error-logging.spec.js
```diff
@@ -174,44 +174,60 @@ describe('ErrorLogging', function() {
...
-    var apmOnError = window.onerror
+    const filename = 'filename'
+    const lineno = 1
+    const colno = 2
 
     try {
       throw new Error(testErrorMessage)
     } catch (error) {
-      apmOnError(testErrorMessage, 'filename', 1, 2, error)
+      errorLogging.logErrorEvent({
```
jahtalab: Maybe we can call the actual listener that we get from the patching `addEventListener`. I know the listener's code is only one line but if we want to change something in the future we can be more confident that it doesn't break the tests.

vigneshshanmugam: Sure, sounds good.

Issue #242, opened 2019-04-08
Automate updating agent version before each release


jahtalab:

vigneshshanmugam:
Personally i am in favor of doing it manually since it we might accidentally push a wrong version and release it. Plus we have a solid check which would help us figure out if we are trying to release without changing version.

PR #287, opened 2019-06-11
chore(rum): fix release script

### LINKED ISSUE #211: Fix publish script
Bugs:
- Fix the version check script
- ~~Need to add `--access public` to npm publish command (through lerna)~~
- ~~lerna should prompt for 2fa authentication OTP (currently it fails) before each package publish~~

Enhancements:
- The publish scripts should pause before committing the changes for the release, so that we can edit changelog if necessary
- We should check the required env variables before running the actual publish script (currently only `GITHUB_TOKEN`)
- ~~Check if there’s a user logged in for npm~~
- ~~run `release-github` in the same script (#275)~~

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#211 elastic/apm-agent-rum-js#242 
+ Update version automatically during releases
+ use lint-staged for linting during precommit phase.


package.json
```diff
@@ -8,21 +8,27 @@
...
     "prerelease": "npm test",
     "release": "lerna publish && npm run release-github",
-    "release-package": "node scripts/publish",
+    "release-package": "node scripts/publish-package",
     "release-github": "node scripts/release-github",
     "build-docs": "sh ./scripts/build_docs.sh apm-agent-rum-js ./docs ./build",
     "lint": "npm-run-all --silent -p lint:*",
     "lint:js": "eslint .",
     "lint:license": "node scripts/license-checker",
-    "format": "eslint . --fix",
```
vigneshshanmugam: done as part of precommit automatically so removed it.

vigneshshanmugam:
Seems like babel patch version breaks the test, I am trying to figure out the reason.

vigneshshanmugam:
Fixed it, It was due to lock files.

scripts/prompt-release.js
```diff
@@ -0,0 +1,60 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+const log = require('npmlog')
+const PromptUtils = require('@lerna/prompt')
+const childProcess = require('@lerna/child-process')
+
+!(async () => {
```
jahtalab: Are you using `!` to prevent syntax error with previous line? might be clearer to just use `;`, so it would be `;(async ()=>{})()`

package.json
```diff
@@ -8,21 +8,27 @@
...
     "test": "lerna run build:main && lerna run test --stream --scope=$SCOPE --no-prefix",
     "bundlesize": "lerna run bundlesize --stream"
   },
+  "lint-staged": {
+    "*.{js,jsx}": [
+      "eslint . --fix",
+      "git add"
```
jahtalab: I was thinking about that this might create issues, imagine that you have some modified files (not untracked) that you don't want to include in this commit, e.g. because you want to have separate commits.

vigneshshanmugam: The way it works is that
1. You add the changes that needs to go in the commit by doing `git add <blah>`
2. Then lint staged gets picked up by precommit hook and then `eslint --fix` and `git add` happens only on those staged files. Modified files that you had wont be affected. So we are safe here.

scripts/prompt-release.js
vigneshshanmugam: Both does the same job as per SPEC. Personally I would just remove `standard js` and go with the language. 

If you still think it make sense, I am fine to change it to `;`

PR #288, opened 2019-06-12
chore: refactor test scripts

jahtalab:


package-lock.json
```diff
@@ -1886,6 +1886,12 @@
       "integrity": "sha512-vw3VyFPa9mlba6NZPBZC3q2Zrnkgy5xuCVI43/tTLX6umdYrYvcFtQUKi2zH3PjFZQ9XCxNM/NMrM9uk8TPOzg==",
       "dev": true
     },
+    "@types/node": {
```
vigneshshanmugam: can you undo these changes in `package-lock.json`. Seems unrelated to the PR.

dev-utils/run-script.js
```diff
@@ -51,10 +64,88 @@ function launchSauceConnect() {
...
+    } finally {
+      servers.map(s => s.close())
+      sauceConnectProcess.close(() => {
+        exitCode && process.exit(exitCode)
+      })
+    }
+  })
+}
+
+function serveStatic(path = './', port = 8000) {
```
vigneshshanmugam: Is this server used anywhere?

dev-utils/webdriver.js
```diff
@@ -32,6 +32,10 @@ const logLevels = {
   OFF: { value: Number.MAX_VALUE }
 }
 
+const { join } = require('path')
```
vigneshshanmugam: can you move the require statements to top of the file?

packages/rum/package.json
```diff
@@ -17,7 +17,7 @@
     "build": "webpack",
     "build:dev": "webpack -w",
     "build:main": "npx babel src -d dist/lib --config-file ../../babel.node.js",
-    "build:e2e": "npm run script buildE2eBundles",
+    "build:e2e": "node ../../dev-utils/run-script.js buildE2eBundles packages/rum/test/e2e",
```
vigneshshanmugam: nit: remove the `.js` extension from the scripts.

dev-utils/webdriver.js
```diff
@@ -102,6 +106,59 @@ function isLogEntryATestFailure(entry, whitelist) {
   return result
 }
 
+function getWebdriveBaseConfig(path) {
```
vigneshshanmugam: nit: `getWebdriverConfig` feels better.

jahtalab: I guess it supposed to be the "base config" so that it can be used as a base for the actual config. Not sure though.

packages/rum/package.json
jahtalab: Not sure about removing the extension, might be more clear to have it there. Don't have a strong opinion here though.

dev-utils/run-script.js
jahtalab: Good catch, I removed it!

packages/rum/package.json
vigneshshanmugam: Just to keep it consistent with other scripts in package.json and also feels easy to read on smaller widths(laptop screen).

packages/rum/package.json
```diff
@@ -26,9 +26,9 @@
     "test:node": "npm run script runNodeTests",
     "test:integration": "npm run script runIntegrationTests",
     "test:unit": "node ../../dev-utils/run-script.js runUnitTests false packages/rum",
-    "test:e2e:supported": "npm run script runE2eTests wdio.conf.js",
-    "test:e2e:failsafe": "npm run script runE2eTests wdio-failsafe.conf.js",
-    "test:sauce": "npm run script runSauceTests",
+    "test:e2e:supported": "node ../../dev-utils/run-script.js runE2eTests packages/rum/wdio.conf.js",
+    "test:e2e:failsafe": "node ../../dev-utils/run-script.js runE2eTests packages/rum/wdio-failsafe.conf.js",
+    "test:sauce": "node ../../dev-utils/run-script.js runSauceTests true packages/rum test:unit test:e2e:supported test:e2e:failsafe",
```
vigneshshanmugam: this seems wrong. the point of sauce script tests was to run unit and e2e in parallel.

vigneshshanmugam:
```js
await runAll('test:unit', loggerOpts)
      await runAll(['test:e2e:supported', 'test:e2e:failsafe'], {
        parallel: true,
        printLabel: true,
        ...loggerOpts
      })
```

The set up should look similar to this, Right now it is not.

jahtalab: Since this method is generic we need to provide parallel tasks separately and I think that makes script more complicated. Furthermore, I think it's better to increase the parallel factor of individual tasks instead of running two task at the same time.

packages/rum/package.json
jahtalab: I think we eventually need to move everything under `rum/run-script.js` to `dev-utils/run-script.js` then we can use `npm run script` so let's clean this up in the next iteration when we remove `rum/run-script.js`

packages/rum/package.json
vigneshshanmugam: Not really, its better to at least run the e2e tests parallel. Plus the script takes so many args and its better to just separate those instead of making it more generic. 

```js

// Not exported
function runSauceTests(takstoRun, flag) {
 // stays the same
// Decide on serial parallel here based on flag
}

/ / Exported method
export function startSauceTest(packageName) {
  const tasks = []
  
  if(packageName === "rum"  {
    tasks.push(...)
  } else if(packageName === "react"){
      // do something here
  }

 return runSauceTest(tasks, flag)
}
```

Package.json would look like
```json
"test:sauce":  "node ./run-script startSauceTest package/rum"
```

Currently both the `package.json` and script seems to everything together with lots of parameters. Its better to move them to a single place. Makes it easier to understand and also reason about.

dev-utils/run-script.js
```diff
@@ -51,10 +63,71 @@ function launchSauceConnect() {
...
+function runSauceTests(serve = 'true', path = './', ...scripts) {
+  /**
+   * `console.logs` from the tests will be truncated when the process exits
+   * To avoid truncation, we flush the data from stdout before exiting the process
+   */
+  if (process.stdout.isTTY && process.stdout._handle) {
+    process.stdout._handle.setBlocking(true)
+  }
+
+  let servers = []
```
vigneshshanmugam: since you are already refactoring,  I would completely remove this `serve` flag and just start server by default.

packages/rum/package.json
jahtalab: I think the difference in performance is not significant, if any at all. 
The approach in your code is not generic, it has special cases for packages. this is not scalable for multiple packages and it's hard to keep track of.

IMO, it's better to consume all possible sessions for each individual tasks. why do you think running two tasks in parallel has any advantage?

dev-utils/run-script.js
jahtalab: The serve flag is useful for development. e.g. if you want to only run the tests but don't want each time to take bring up a new server which takes time.

packages/rum/package.json
jahtalab: Also running in parallel makes finding failures and keeping track of tests harder since the failsafe and supported results are mixed together:

image

dev-utils/run-script.js
vigneshshanmugam: I like to avoid too many moving parts in the code, We could just start the servers before and move the logical parts out.

packages/rum/package.json
vigneshshanmugam: > I think the difference in performance is not significant, if any at all.

Do we have any numbers here? 

> The approach in your code is not generic, it has special cases for packages. this is not scalable for multiple packages and it's hard to keep track of.

No ,ignore the tasks on whether its serial or parallel. Look at the number of flags that is getting passed to the package.json script. It looks totally weird. I proposed to have them in a single place inside single function. 

```json
"test:sauce":  "node ./run-script startSauceTest package/rum"
```

> Also running in parallel makes finding failures and keeping track of tests harder since the failsafe and supported results are mixed together:

its just output aggregation, you can just set `aggregateOutput`: true and it would fix it.

vigneshshanmugam: I am fine with keeping tasks in serial if tests time delta is not significant. But lets try to keep the package.json a bit clean. It really gets messy by keeping list of tasks there along with other args.

dev-utils/run-script.js
```diff
@@ -51,10 +63,71 @@ function launchSauceConnect() {
...
+    servers = startTestServers(join(PROJECT_DIR, path))
+  }
+  /**
+   * Decides the saucelabs test status
+   */
+  let exitCode = 0
+  const loggerOpts = {
+    stdout: process.stdout,
+    stderr: process.stderr,
+    parallel: false,
```
vigneshshanmugam: if you decide to run tasks in serial, let's remove this key and also printLabel, default is false anyways.

packages/rum/package.json
jahtalab: I don't see an issue with the number of arguments send to a function, the issue I think is that they don't have names so ideally it should be `node ./run-script startSauceTest --package=package/rum`. similar to how we use lerna `lerna run test --stream --scope=$SCOPE --no-prefix`

In any case, we can always iterate on this, I don't think we should fix everything in this PR.

vigneshshanmugam: We have been doing this iteration for quite some time and that was the reason why I suggested this change. What I meant is to keep the package.json clean and scripts convey what its doing instead of making everything generic. 
```json
"test: sauce": "node ./run-script startSauceTest --package=package/rum"
```
will stay the same and just the tasks will be moved closer to the script. 

```js
const TASKS_FOR_RUM = ["unit", "e2e"]
function startSauceTest(package) {
   let tasks = []
    if(package==="rum") . {
        tasks.concat(TASKS_FOR_RUM)
   }
}
```

I am fine with merging this PR as it as now.

jahtalab: FWIW, I think having special cases in a generic script is not a good idea and certainly makes the script harder to follow when we have 3 or more packages, but I'm open to other suggestions.

Thanks for the review 👍

vigneshshanmugam: My whole reasoning behind it was bloated `package.json`. If we have to run another taks, the list keeps on growing. Its better to keep them in a single place where generic script lies. We can merge the PR and revisit it later if needed.  Lets not spend more time on this 😄

PR #294, opened 2019-06-17
feat(rum-core): use page url as default name for page-load transaction

### LINKED ISSUE #255: Provide a default name for page load transactions
Since we have changed how Kibana shows most impactful transactions we might be able to use page url for the transaction name by default now. We should investigate if there are other potential issues with this approach.

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#255

### DIFF
--- a/packages/rum-core/src/performance-monitoring/span-base.js
+++ b/packages/rum-core/src/performance-monitoring/span-base.js
@@ -29,7 +29,16 @@ import { NAME_UNKNOWN, TYPE_CUSTOM } from '../common/constants'
 class SpanBase {
   // context
 
-  constructor(name = NAME_UNKNOWN, type = TYPE_CUSTOM, options = {}) {
+  constructor(name, type, options = {}) {
+    /**
+     * Check for undefined and '' string
+     */
+    if (!name) {
+      name = NAME_UNKNOWN
+    }
+    if (!type) {
+      type = TYPE_CUSTOM
+    }
     this.options = options
     this.name = name
     this.type = type
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -24,8 +24,12 @@
  */
 
 import Transaction from './transaction'
-import { extend, getPageLoadMarks } from '../common/utils'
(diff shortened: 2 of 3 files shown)


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -122,14 +126,6 @@ class TransactionService {
   startTransaction(name, type, options) {
     const perfOptions = this.createPerfOptions(options)
 
-    if (!type) {
```
vigneshshanmugam: Already taken care in `span-base` https://github.com/elastic/apm-agent-rum-js/blob/0116f68aca3eccf5bc3fafcee6d13e87de8d656b/packages/rum-core/src/performance-monitoring/span-base.js#L32

packages/rum-core/src/performance-monitoring/span-base.js
```diff
@@ -29,7 +29,16 @@ import { NAME_UNKNOWN, TYPE_CUSTOM } from '../common/constants'
 class SpanBase {
   // context
 
-  constructor(name = NAME_UNKNOWN, type = TYPE_CUSTOM, options = {}) {
+  constructor(name, type, options = {}) {
```
vigneshshanmugam: default assignments checks for undefined values. Wont work for empty strings. Since our config-service has default values of `''` this check is moved here to handle all spans and transactions.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -160,19 +155,18 @@ class TransactionService {
...
         tr.traceId = perfOptions.pageLoadTraceId
       }
       if (perfOptions.pageLoadSampled) {
         tr.sampled = perfOptions.pageLoadSampled
       }
-      /**
-       * Retriving the name before transaction ends should reflect
-       * the correctg page load transaction name
-       */
-      if (tr.name === NAME_UNKNOWN && perfOptions.pageLoadTransactionName) {
```
vigneshshanmugam: This is totally unnecessary, We send the `pageLoadTransactionName` from `apm-base` which already handles this  scenario.

jahtalab:
Decided against this change until the UI is improved to handle this, we should extract other improvements from this PR and close this one.

vigneshshanmugam:
Closing this in favor of #304

PR #299, opened 2019-06-25
chore: add chrome 74 to e2e test

vigneshshanmugam:
+ test to figure out if DT works as expected in latest browsers.

### DIFF
--- a/dev-utils/test-config.js
+++ b/dev-utils/test-config.js
@@ -97,10 +97,6 @@ function getBrowserList() {
       version: '52'
     },
     {
-      browserName: 'firefox',
-      version: '66'
-    },
-    {
       browserName: 'safari',
       platform: 'OS X 10.11',
       version: '9.0'
--- a/dev-utils/test-config.js
+++ b/dev-utils/test-config.js
@@ -103,12 +103,7 @@ function getBrowserList() {
     {
       browserName: 'safari',
       platform: 'OS X 10.11',
-      version: '9.1'
-    },
-    {
-      browserName: 'safari',
-      platform: 'OS X 10.11',
-      version: '12.1'
+      version: '9.0'
     },
     {
       browserName: 'internet explorer',
@@ -128,7 +123,7 @@ function getBrowserList() {
       platformName: 'android'
     },
     {
-      appiumVersion: '1.9.1',
+      appiumVersion: '1.13.0',
       deviceName: 'iPhone Simulator',
       deviceOrientation: 'portrait',
       platformVersion: '12.2',
--- a/dev-utils/test-config.js
+++ b/dev-utils/test-config.js
@@ -93,6 +93,10 @@ function getBrowserList() {
       version: '62'
     },
     {
+      browserName: 'chrome',
+      version: '74'
+    },
+    {
       browserName: 'firefox',
       version: '59'
     },
--- a/dev-utils/test-config.js
+++ b/dev-utils/test-config.js
@@ -81,6 +81,11 @@ function getWebpackEnv() {
   }
 }
 
+/**
+ * Supported lowest and highest versions across major browser platform
+ *
+ * The list below is based purely on the market share distribution.
+ */
 function getBrowserList() {
   return [
     {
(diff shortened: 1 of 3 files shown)


dev-utils/test-config.js
```diff
@@ -92,6 +92,10 @@ function getBrowserList() {
       browserName: 'chrome',
       version: '62'
```
jahtalab: Maybe we should just change this to 74 instead, since their market share drops significantly after a new updates comes out!

vigneshshanmugam: I started this test after one of our discuss issue on chrome 74. Let's come up with some small list that we have to run tests on. I was thinking to have 1 with least -> mid range -> current/latest. What are your thoughts?

jahtalab: I think we should keep the lowest version and the highest! IMO, the benefit of having a mid decreases as new versions become available since in most cases it will automatically update and in the case of chrome new versions are quite frequent!

vigneshshanmugam: Yeah sure. I will come update the list with new versions on each platform and will move the PR status accordingly. 

DT works fine in chrome 74. Probably you can link the PR to the discuss chat if you would like the user to have a look.

dev-utils/test-config.js
```diff
@@ -119,10 +119,10 @@ function getBrowserList() {
       platformName: 'android'
     },
     {
-      appiumVersion: '1.9.1',
+      appiumVersion: '1.13.0',
       deviceName: 'iPhone Simulator',
       deviceOrientation: 'portrait',
-      platformVersion: '11.1',
+      platformVersion: '12.2',
```
jahtalab: Why do we need to drop support for iOS 11?

packages/rum-core/test/common/fetch-patch.spec.js
```diff
@@ -25,28 +25,31 @@
...
-      events.push({
-        event,
-        task
-      })
+      if (task.source === FETCH_SOURCE) {
+        events.push({
+          event,
+          task
+        })
+      }
```
jahtalab: I think this is part of the tests, in other words we add all of the events to the array to make sure we only get the ones that are relevant to the test.

We should improve the tests to account for this.

packages/rum-core/test/common/xhr-patch.spec.js
```diff
@@ -29,20 +29,27 @@ import {
...
-      events.push({
-        event,
-        task
-      })
+      if (task.source === XMLHTTPREQUEST_SOURCE) {
+        events.push({
+          event,
+          task
+        })
+      }
```
jahtalab: Same here.

dev-utils/test-config.js
vigneshshanmugam: We are not dropping, 12.2 has the major share so i just updated it since we already test for safari 9.1 in mac. least + current.

packages/rum-core/test/common/fetch-patch.spec.js
vigneshshanmugam: Since we use a global `events` variable to account for this, Our tests are getting flaky and breaking sometime. Check this 
```
LOG: Object{event: 'schedule', task: Object{source: 'XMLHttpRequest.send', state: 'invoke', type: 'macroTask', ignore: undefined, data: Object{target: ..., method: ..., sync: ..., url: ..., args: ..., aborted: ...}}},
LOG: Object{event: 'invoke', task: Object{source: 'fetch', state: 'invoke', type: 'macroTask', data: Object{target: ..., method: ..., sync: ..., url: ..., args: ..., aborted: ..., error: ...}}},
LOG: Object{event: 'invoke', task: Object{source: 'XMLHttpRequest.send', state: 'invoke', type: 'macroTask', ignore: undefined, data: Object{target: ..., method: ..., sync: ..., url: ..., args: ..., aborted: ...}}}
```

jahtalab: There seems to be a bug in one of our tests in `fetch-patch.spec.js` that ends the test before the async operation is done

Issue #26, opened 2018-06-07
Add benchmarks


jahtalab:
- [x] Micro benchmarks
     - [x] Add benchmarks for Transaction & Span creation overhead. 
- [x] Document benchmarks (Plus the Overhead)
- [x] Store benchmarks in ES
- [x] Create Kibana dashboards
- [x] Document the package size
- [x] Macro benchmarks on how the agent is affecting the user experience elastic/apm-agent-rum-js#302 
- [x] Run the benchmarks on Jenkins CI for publishing to ES
- [x] Run the benchmarks on bare metal server elastic/apm-agent-rum-js#456 

Postponed: 
- [ ] Run the benchmarks on each PR and add fields in the elasticsearch document - https://github.com/elastic/observability-dev/issues/299 -> https://github.com/elastic/apm-agent-rum-js/issues/568
- [x] Add memory benchmark for the browser - memory leaks -> https://github.com/elastic/apm-agent-rum-js/issues/514
- [ ] Consider moving the benchmark runner to the top to avoid reading/storing files multiple times

jahtalab:
Preliminary results:

```
[
  {
    "name": "PerformanceMonitoring.createTransactionPayload",
    "mean": 0.000011806532270798866,
    "hz": 84698.87491632943
  },
  {
    "name": "PerformanceMonitoring.sendTransactions-no-json",
    "mean": 0.000023883519882611687,
    "hz": 41869.875333076285
  },
  {
    "name": "PerformanceMonitoring.sendTransactions",
    "mean": 0.00023317307692307694,
    "hz": 4288.659793814432
  }
]
```

vigneshshanmugam:
Blocked on Jenkins Pipeline.

jahtalab:
We still need to make the Kibana dashboards and also the scripts to send benchmark payload to ES. 
So it's currently not blocked.

vigneshshanmugam:
Some thoughts on what are the dimensions that we have to come up on the RUM agent when it comes to the performance of the application and also the user experience. 

+ We have to document the different loading patterns and explain the pros/cons of the approach because this decides how our script parse and execution affects the page-load. We have to find how we **affect the browser from firing onload event**. 

+ **Parse time** of our agent script. Once the agent is downloaded, we patch some of the apis(XHR, Fetch) and also create global APM object. 

[...]

PR #302, opened 2019-06-26
chore(benchmark): measure impact of rum agent during page load 

vigneshshanmugam:
+ Initial draft of rum benchmarks as described here https://github.com/elastic/apm-agent-rum-js/issues/26#issuecomment-504027644

## How its run

scenarios(configurable - basic and heavy) 
total number of runs (configurable - 3) 
where the below metrics are calculated and analysed for mean and p90 on each run. 
 + CPU profile of the whole application helps us to capture time spent running the functions of JS agent (It included basically span creation, init cost and almost everything)
 + time spent running all the other code as well on the page
 + Initialisation cost of the agent `elasticApm.init` 
 + Page load time with the agent
 + Size of the payload - varies with each scenario

Test results looks like this,  which will be uploaded the ES cluster for creating dashboard. 

```json
{
  "process": {
    "version": "v10.15.2",
    "arch": "x64",
    "platform": "darwin"
  },
  "meta": {
    "commit": "b84056a",
    "commitMessage": "cache burst apm url and measure tr and spans",
    "branch": "rum-benchmarks",
    "agentName": "rum-js"
  },
  "metrics": {
    "span creation overhead": {
      "suite": "TransactionService",
      "name": "span creation overhead",
      "browser": "Chrome 76.0.3809 (Mac OS X 10.14.6)",
      "hz": 55250.705707234236,
      "unit": "ops/sec"
    },
    "custom transaction overhead": {
      "suite": "TransactionService",
      "name": "custom transaction overhead",
      "browser": "Chrome 76.0.3809 (Mac OS X 10.14.6)",
      "hz": 45371.76614794443,
      "unit": "ops/sec"
    },
    "page-load transaction overhead": {
      "suite": "TransactionService",
      "name": "page-load transaction overhead",
      "browser": "Chrome 76.0.3809 (Mac OS X 10.14.6)",
      "hz": 42251.03754628884,
      "unit": "ops/sec"
    },
    "basic": {
      "browser": "HeadlessChrome/77.0.3844.0",
      "url": "http://localhost:9000/basic",
      "runs": 3,
      "scenario": "basic",
      "page-load-time.mean.ms": 22.584999989097316,
      "page-load-time.p90.ms": 33.73000002466142,
      "rum-init-time.mean.ms": 2.614999966075023,
      "rum-init-time.p90.ms": 2.6549999602138996,
      "total-cpu-time.mean.ms": 19585.215000000007,
      "total-cpu-time.p90.ms": 20066.355999999996,
      "rum-cpu-time.mean.ms": 10222.147333333334,
      "rum-cpu-time.p90.ms": 10804.962,
      "payload-size.mean.bytes": 1878.3333333333333,
      "payload-size.p90.bytes": 1882,
      "transactions.mean.count": 1,
      "transactions.p90.count": 1,
      "spans.mean.count": 3,
      "spans.p90.count": 3,
      "bundle-size.mean.bytes": 50615.333333333336,
      "bundle-size.p90.bytes": 50618
    },
    "heavy": {
      "browser": "HeadlessChrome/77.0.3844.0",
      "url": "http://localhost:9000/heavy",
      "runs": 3,
      "scenario": "heavy",
      "page-load-time.mean.ms": 63.656666704143085,
      "page-load-time.p90.ms": 67.37000006251037,
      "rum-init-time.mean.ms": 2.801666695935031,
      "rum-init-time.p90.ms": 2.9600000707432628,
      "total-cpu-time.mean.ms": 64140.3556666667,
      "total-cpu-time.p90.ms": 64494.37799999988,
      "rum-cpu-time.mean.ms": 11802.013333333334,
      "rum-cpu-time.p90.ms": 12333.596,
      "payload-size.mean.bytes": 14716,
      "payload-size.p90.bytes": 14892,
      "transactions.mean.count": 1,
      "transactions.p90.count": 1,
      "spans.mean.count": 33.333333333333336,
      "spans.p90.count": 34,
      "bundle-size.mean.bytes": 50615.333333333336,
      "bundle-size.p90.bytes": 50618
    }
  },
  "@timestamp": 1565709284566
}
```


vigneshshanmugam:
picking this up again.

axw:
@vigneshshanmugam I've only had a brief look, but looks good. Profiling to calculate the CPU overhead is neat.

Do you intend to also run a test to calculate the base load time, without the agent, so that we can calculate the overhead?

I suggest that you collapse the metrics into a single doc per scenario. To do that, you would need to combine several of the fields together so you end up with something like:

```
{
    "scenario": "basic",
    "browser": "HeadlessChrome/77.0.3844.0",
    "url": "http://localhost:9000/basic",
    "page-load-time.mean.ms": 25.920000043697655,
    "page-load-time.p90.ms": 44.615000020712614,
    "bundle-size.mean.bytes": 50615.333333333336,
    "bundle-size.p90.bytes": 50618,
}
```

A timestamp for when the recording was made would also be useful, for plotting values over time.

vigneshshanmugam:
@axw Thanks for the review. 

>Do you intend to also run a test to calculate the base load time, without the agent, so that we can calculate the overhead?

I was actually thinking about it but decided to not do it since the overhead might not be visible by comparing the delta of with/without the agent since its mostly network time. And also the basic test covers only this scenario, Only the agent script gets downloaded without any other resources and heavy would include lots of images/css which would help identify the cost of the agent as well payload size. 

> I suggest that you collapse the metrics into a single doc per scenario. To do that, you would need to combine several of the fields together so you end up with something like:

I was following the DS with our previous benchmark metrics https://github.com/elastic/apm-agent-rum-js/pull/325#issuecomment-511875047. Happy to hear your thoughts and change it as well. 

> A timestamp for when the recording was made would also be useful, for plotting values over time.

Yup, we have it as part when the index document is created which is separated from benchmark runs. https://github.com/elastic/apm-agent-rum-js/blob/master/scripts/benchmarks.js#L134

axw:
> I was actually thinking about it but decided to not do it since the overhead might not be visible by comparing the delta of with/without the agent since its mostly network time. And also the basic test covers only this scenario, Only the agent script gets downloaded without any other resources and heavy would include lots of images/css which would help identify the cost of the agent as well payload size.

OK, fair enough.

> I was following the DS with our previous benchmark metrics #325 (comment). Happy to hear your thoughts and change it as well.

I see. In that case you end up with a document like this?

```
"metrics": {
    "rum-cpu-time": {
        "scenario": "heavy",
        "name": "rum-cpu-time",
        "mean": 6300.9490000000005,
        "browser": "HeadlessChrome/77.0.3844.0",
        "url": "http://localhost:9000/heavy",
        "p90": 6470.112,
        "unit": "ms"
    }
}
```

In which case you can query `metrics.rum-cpu-time.mean`. That works too, I was just missing the context. It seems a bit wasteful in terms of repeating fields (scenario, name, browser, url). Maybe something to revisit later.

> Yup, we have it as part when the index document is created which is separated from benchmark runs. https://github.com/elastic/apm-agent-rum-js/blob/master/scripts/benchmarks.js#L134

:+1:

vigneshshanmugam:
> In which case you can query metrics.rum-cpu-time.mean. That works too, I was just missing the context. It seems a bit wasteful in terms of repeating fields (scenario, name, browser, url). Maybe something to revisit later.

Yeah true for now it looks wasteful, But we had plans to run these benchmarks for different browsers which wont work if we sort them based on scenarios. Also while uploading the document,  we have filters in place which would only choose the required fields. https://github.com/elastic/apm-agent-rum-js/blob/21b72a7c391588354a16f3208cb678abc490f999/scripts/benchmarks.js#L55-L60

I will modify this script to remove the unnecessary items for this type.

packages/rum-core/package.json
```diff
@@ -17,11 +17,11 @@
   "scripts": {
     "prepublishOnly": "npm run build:main",
     "build:main": "npx babel src -d dist/lib",
-    "karma:bench": "NODE_ENV=production karma start karma.bench.conf.js",
     "karma:coverage": "karma start --coverage --singleRun",
     "test:unit": "node ../../dev-utils/run-script.js runUnitTests packages/rum-core true",
     "test": "npm run test:unit",
-    "tdd": "karma start --auto-watch --restartOnFileChange"
+    "tdd": "karma start --auto-watch --restartOnFileChange",
```
jahtalab: is this "Test Driven Development"? something like `karma:dev` might be more aligned with our script naming convention.

packages/rum/test/benchmarks/config.js
```diff
@@ -0,0 +1,39 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+const port = 9000
+module.exports = {
+  scenarios: ['basic', 'heavy'],
+  runs: 3,
```
jahtalab: Is 3 enough for the number of runs! Do you have a rule of thumb for what the most efficient number would be, considering the trade-off between speed and accuracy?

packages/rum/test/benchmarks/config.js
```diff
@@ -0,0 +1,39 @@
...
+  port,
+  elasticApmUrl: `http://localhost:${port}/elastic-apm-rum.js`,
+  chrome: {
+    samplingInterval: 200,
```
jahtalab: What is the limiting factor for setting this lower?

packages/rum/test/benchmarks/profiler.js
```diff
@@ -0,0 +1,103 @@
...
+     * Tune the CPU sampler to get control the
+     * number of samples generated
+     */
+    await client.send('Profiler.setSamplingInterval', {
+      interval: chrome.samplingInterval
+    })
+
+    /**
+     * Result metrics that will be filled at various
+     * time throught the page lifecycle
```
jahtalab: nit: 'throughout'

packages/rum/test/benchmarks/profiler.js
```diff
@@ -0,0 +1,103 @@
...
+        /**
+         * Resolve the promise once we measure the size
+         * of the payload to APM Server
+         */
+        resolve(metrics)
+      }
+    })
+
+    client.on('Page.loadEventFired', async function() {
+      const result = await client.send('Profiler.stop')
```
jahtalab: Should we stop the profile a few cycles after the load event?

packages/rum/test/benchmarks/server.js
```diff
@@ -0,0 +1,77 @@
...
+
+module.exports = function startServer() {
+  return new Promise(resolve => {
+    const app = express()
+
+    app.get('/elastic-apm-rum.js', (req, res) => {
+      createReadStream(
+        path.join(dist, 'bundles/elastic-apm-rum.umd.min.js'),
+        'utf-8'
+      ).pipe(res)
```
jahtalab: We can set proper caching headers to make the results predictable.

packages/rum/test/benchmarks/server.js
```diff
@@ -0,0 +1,77 @@
...
+const express = require('express')
+const path = require('path')
+const { elasticApmUrl, port } = require('./config')
+const { createReadStream } = require('fs')
+
+const pages = path.join(__dirname, 'pages')
+const dist = path.join(__dirname, '../../dist')
+
+function generateImageUrls() {
+  const number = 30
```
jahtalab: Maybe refactor the number into the function arguments!

packages/rum/test/benchmarks/config.js
vigneshshanmugam: by defaults 1000ms, since the basic version loads even faster,  to take more samples in this timeframe had to tune it. Right now the number of samples is  >50 for each run which seems good enough. I will add comments to the config to make it clear.

packages/rum/test/benchmarks/config.js
vigneshshanmugam: I just went with the lighthouse default config runs. Any odd number between 3 and 9 would work.

packages/rum-core/package.json
vigneshshanmugam: Yeah, We can change it in another PR. `karma:dev` is also fine IMO.

packages/rum/test/benchmarks/profiler.js
vigneshshanmugam: We could either done it here or wait until we capture the payload size which would capture post load event for few more milliseconds. What is your opinion?

packages/rum/test/benchmarks/server.js
vigneshshanmugam: Hmmm good idea, Let me cache burst this script to make it clear.

packages/rum/test/benchmarks/server.js
vigneshshanmugam: Sure 👍

packages/rum/test/benchmarks/profiler.js
vigneshshanmugam: Moved it till payload is captured.

vigneshshanmugam:
@jahtalab Thanks for the review. 

> As we discovered yesterday the cpu numbers for heavy and basic don't look right, would you please update the PR description if you have already fixed that issue?

Basically because of launch browser command, had to create separate process to avoid cache and it fixed the issue. 

> It might be a good idea to include the total number of transactions and spans generated during the benchmark.

Done, Also changed the data structure to match with our needs and as suggested by @axw since we had lots of common fields. Thanks Andrew, it definitely looks cleaner. 

> I was thinking about the current basic scenario, maybe we should change it so it doesn't include the rum agent at all.

I already commented about it in the previous conversation with Andrew https://github.com/elastic/apm-agent-rum-js/pull/302#issuecomment-520330515

> How should we handle the changes that we make in the benchmark parameters in the future to make sure we compare the correct benchmark results

I guess it's a learning process, we can definitely include all the variable parameters in the result of each scenario to account for those changes and filter in the dashboard to get clear view of things.

jahtalab:
Decided to:
- making markings for parse and execution time of the RUM agent js file
- adding benchmark parameters to the result docs, e.g. `parameters.runs: 3`

But will do these in a separate PR.

PR #304, opened 2019-06-26
chore: use default name and type for transaction and spans

vigneshshanmugam:
+ Set the default name and type for all transaction and span and remove the check in other places on the codebase. 
+ Extracted from other PR elastic/apm-agent-rum-js#294

### DIFF
--- a/packages/rum-core/src/performance-monitoring/span-base.js
+++ b/packages/rum-core/src/performance-monitoring/span-base.js
@@ -29,7 +29,16 @@ import { NAME_UNKNOWN, TYPE_CUSTOM } from '../common/constants'
 class SpanBase {
   // context
 
-  constructor(name = NAME_UNKNOWN, type = TYPE_CUSTOM, options = {}) {
+  constructor(name, type, options = {}) {
+    /**
+     * Check for undefined and empty string
+     */
+    if (!name) {
+      name = NAME_UNKNOWN
+    }
+    if (!type) {
+      type = TYPE_CUSTOM
+    }
     this.options = options
     this.name = name
     this.type = type
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -25,7 +25,7 @@
 
 import Transaction from './transaction'
 import { extend, getPageLoadMarks } from '../common/utils'
-import { PAGE_LOAD, NAME_UNKNOWN, TYPE_CUSTOM } from '../common/constants'
+import { PAGE_LOAD, NAME_UNKNOWN } from '../common/constants'
 import Subscription from '../common/subscription'
 import { captureHardNavigation } from './capture-hard-navigation'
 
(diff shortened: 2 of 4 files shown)


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -111,7 +111,6 @@ class TransactionService {
         pageLoadTraceId: config.pageLoadTraceId,
         pageLoadSampled: config.pageLoadSampled,
         pageLoadSpanId: config.pageLoadSpanId,
-        pageLoadTransactionName: config.pageLoadTransactionName,
```
jahtalab: Why we don't need this here? We should still set the correct transaction name at the start of the transaction.
I think we discuss this previously, the transaction should have the correct name as soon as possible, e.g. in case we want to decide sampling based on the name.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -164,13 +155,6 @@ class TransactionService {
       if (perfOptions.pageLoadSampled) {
         tr.sampled = perfOptions.pageLoadSampled
       }
-      /**
-       * Retriving the name before transaction ends should reflect
```
jahtalab: The same here, we need to set the correct name at the start of the transaction.

vigneshshanmugam: Its already set as part of apm-base https://github.com/elastic/apm-agent-rum-js/blob/b073f7f7a4c2ffdc248b8c5d6eff5a13c5eb0afd/packages/rum/src/apm-base.js#L75.  I really have no clue why we need this like 3 places if apm-base is the once that sets the listener and registers for capturing the page-load transaction.

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: explained in the below comment.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: I think we should remove the one in the apm-base, to avoid having the logic split in multiple places and keep the page load logic in transaction service!

packages/rum-core/src/performance-monitoring/span-base.js
```diff
@@ -29,7 +29,16 @@ import { NAME_UNKNOWN, TYPE_CUSTOM } from '../common/constants'
 class SpanBase {
   // context
 
-  constructor(name = NAME_UNKNOWN, type = TYPE_CUSTOM, options = {}) {
+  constructor(name, type, options = {}) {
+    /**
+     * Check for undefined and empty string
```
jahtalab: We should have caught this once we did this change recently, would you please add tests for falsy values to make sure default values are always used!

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: Sure, in that case lets' have it here.

jahtalab: Would you please add a comment to make it clear for future!

vigneshshanmugam: Sure thing.

PR #306, opened 2019-06-28
chore(rum-core): fix benchmark karma runner script [skip ci]

vigneshshanmugam:
+ dependencies were missing from `package.json` and script was specPattern was wrong. 

+ create reports dir manually since json reporter does not take of creating it for us.


packages/rum-core/test/benchmarks/001-performance-monitoring.bench.js
```diff
@@ -49,26 +37,24 @@ suite('PerformanceMonitoring', function() {
...
   }
+  const tr = generateTestTransaction(10)
+
+  benchmark('createTransactionPayload', function() {
+    performanceMonitoring.createTransactionPayload(tr)
+  })
 
   benchmark('sendTransactions-no-json', function() {
     apmServer._postJson = ResolvedPromise
-    var tr = generateTransaction()
```
vigneshshanmugam: test should not get influenced by creating transaction and span object. I will create separate benchmarks for transaction and span creation time.

vigneshshanmugam: It will happen in separate PR.

.gitignore
```diff
@@ -12,6 +12,7 @@ html_docs
 build
 reports
 coverage
+**/benchmarks/*.json
```
jahtalab: We wouldn't need this if we separate the reports from the actual benchmarks!

packages/rum-core/karma.bench.conf.js
```diff
@@ -48,13 +51,17 @@ module.exports = function(config) {
       'karma-benchmark-json-reporter'
     ],
     preprocessors: {
-      specPattern: ['webpack', 'sourcemap']
+      [specPattern]: ['webpack', 'sourcemap']
     },
     benchmarkJsonReporter: {
-      pathToJson: 'reports/benchmark-results.json',
+      pathToJson: `${BENCHMARKS_DIR}/benchmark-results.json`,
```
jahtalab: I think we should separate the reports from the actual benchmark code.

packages/rum-core/karma.bench.conf.js
```diff
@@ -35,7 +38,7 @@ module.exports = function(config) {
     JSON.stringify(customConfig, null, 2)
   )
   config.set(customConfig)
-  const specPattern = 'test/**/*.bench.js'
+  const specPattern = `${BENCHMARKS_DIR}/*.bench.js`
```
jahtalab: `${BENCHMARKS_DIR}/**/*.bench.js` would be better since it allows us to restructure the benchmarks directory.

packages/rum-core/test/benchmarks/001-performance-monitoring.bench.js
jahtalab: Would you please add a todo item to the original issue?

vigneshshanmugam: Sure

.gitignore
vigneshshanmugam: Reason i pushed it here is because there is no reports directory in core and we have to create it manually then before the bechmark script runs. Json reporter does create folders, only creates files in existing dir.

packages/rum-core/karma.bench.conf.js
vigneshshanmugam: Sure,

.gitignore
vigneshshanmugam: Do you want to manually create the folder and use it or this you are okay with the above change?

jahtalab: We decided to put the generated benchmark results in the reports folder.

Issue #298, opened 2019-06-21
CONTRIBUTING docs incorrect/incomplete


axw:
I'm trying to follow the CONTRIBUTING docs to run the tests, running the steps verbatim.

The docs say to run:

```
$ git clone git@github.com:elastic/apm-agent-rum-js.git && cd apm-agent-rum-js
$ npm install
$ npm run bootstrap
```

The `npm run bootstrap` command fails. @jahtalab says that this should be changed to `npx lerna bootstrap`, and that we should be able to get rid of the explicit `npm install` too, as [lerna bootstrap](https://github.com/lerna/lerna/tree/master/commands/bootstrap) takes care of that.

So next I ran `npx lerna bootstrap`, followed by `npm test`. This resulted in a bunch of failed integration tests, apparently because there's I had APM Server running on localhost. That bit is missing from the docs.

I started up a stack with apm-integration-testing, and re-ran `npm test`. Now I'm getting one integration test failure:

```
Failures:
1) integration-test should run integration test
  Message:
    Failed: Integration test failed, error: null, response: 500, body: Chromium revision is not downloaded. Run "npm install" or "yarn install"
  Stack:
    Error: Failed: Integration test failed, error: null, response: 500, body: Chromium revision is not downloaded. Run "npm install" or "yarn install"
        at Request.fail [as _callback] (/home/andrew/projects/apm-agent-rum-js/packages/rum/test/integration/integration-test.spec.js:38:11)
        at Request.self.callback (/home/andrew/projects/apm-agent-rum-js/node_modules/request/request.js:185:22)
        at emitTwo (events.js:126:13)
        at Request.emit (events.js:214:7)
        at Request.<anonymous> (/home/andrew/projects/apm-agent-rum-js/node_modules/request/request.js:1161:10)
        at emitOne (events.js:116:13)
        at Request.emit (events.js:211:7)
        at IncomingMessage.<anonymous> (/home/andrew/projects/apm-agent-rum-js/node_modules/request/request.js:1083:12)
        at Object.onceWrapper (events.js:313:30)
        at emitNone (events.js:111:20)
3 specs, 1 failure
Finished in 0.055 seconds
Integration tests failed: Jasmine tests failed
```

That appears to be related to Puppeteer, and I'm not sure how to resolve that.

axw:
[...]

PR #312, opened 2019-07-08
chore: optimise the testing workflow based on the mode set

vigneshshanmugam:
+ `MODE=saucelabs` decides whether to run sauce tests or only local tests. 
+ updated `contributing.md` and `release.md` with relevant information.
+ Moved params around to reduce the default params passed in the package scripts.  
+ Fixes elastic/apm-agent-rum-js#298


package.json
```diff
@@ -47,13 +46,13 @@
...
     "@lerna/child-process": "^3.14.2",
     "@lerna/prompt": "^3.13.0",
-    "@wdio/cli": "^5.7.2",
-    "@wdio/dot-reporter": "^5.7.2",
-    "@wdio/jasmine-framework": "^5.7.2",
-    "@wdio/local-runner": "^5.7.2",
-    "@wdio/sauce-service": "^5.7.2",
-    "@wdio/spec-reporter": "^5.7.2",
-    "@wdio/sync": "^5.7.2",
+    "@wdio/cli": "^5.10.10",
```
vigneshshanmugam: `wdio` packages are updated because of a bug where there were errors thrown because `fibers.node` was not found. So had to update everything.

package.json
```diff
@@ -19,8 +19,7 @@
     "lint:license": "node scripts/license-checker",
     "test": "lerna run build:main && lerna run test --stream --scope=$SCOPE --no-prefix",
     "bundlesize": "lerna run bundlesize --stream",
-    "serve": "node ./dev-utils/run-script.js startTestServers",
-    "karma": "karma start"
```
vigneshshanmugam: removed it from here, because each package.json files already contain these and we can use npx lerna run if required.

CONTRIBUTING.md
```diff
@@ -78,35 +78,42 @@ The project structure follows a monorepo approach, all officially maintained mod
...
+
+##### Integration tests
+```sh
+$ npx lerna run --scope @elastic/apm-rum test:integration
 ```
 
-To run unit/integration/end-to-end tests for packages:
+##### E2E tests (Sauce labs)
+
+E2E tests are run on saucelabs, We have to environment variables before running these tests
```
axw:
```suggestion
E2E tests are run on saucelabs, We have to set some environment variables before running these tests
```

CONTRIBUTING.md
```diff
@@ -78,35 +78,42 @@ The project structure follows a monorepo approach, all officially maintained mod
...
 
 ```sh
-// Unit tests
-$ npx lerna run --scope @elastic/apm-rum test:unit
+$ MODE=saucelabs SAUCE_USERNAME=<username> SAUCE_ACCESS_KEY=<access-key> npx lerna run --scope @elastic/apm-rum test:sauce
+```
 
-// Integration tests
-$ npx lerna run --scope @elastic/apm-rum test:integration
+To run all the tests, Set the environment variables required for the E2E and run the below command
```
axw:
```suggestion
To run all the tests, set the environment variables required for the E2E and run the below command
```

packages/rum/run-script.js
```diff
@@ -23,93 +23,9 @@
...
 const JasmineRunner = require('jasmine')
-const express = require('express')
-const serveIndex = require('serve-index')
 const testUtils = require('../../dev-utils/test-utils')
-const { runIntegrationTest } = require('./test/e2e/integration-test')
-const { generateNotice } = require('../../dev-utils/dep-info')
-
-const PROJECT_DIR = path.join(__dirname, './')
-
-function createBackendAgentServer() {
```
vigneshshanmugam: @jahtalab somehow we missed to remove this while refactoring and this code was repeated twice.

package.json
vigneshshanmugam: Also the test output looks really good with logs on top and status at bottom.

packages/rum/package.json
```diff
@@ -25,11 +25,11 @@
     "serve": "npm run script serveE2e ./ 8000",
```
jahtalab: This can be removed as well. It doesn't work with the current refactoring since `serveE2e` is removed.

dev-utils/run-script.js
```diff
@@ -24,42 +24,38 @@
...
 
-function launchSauceConnect() {
+/**
+ * Checks if the MODE is set to "saucelabs" and decides where to run the
+ * corresponding callback function
+ */
+function launchSauceConnect(callback = () => {}) {
   if (sauceLabs) {
-    return runSauceConnect(sauceConnectOpts, () => {
-      console.log('Launched SauceConnect!')
```
jahtalab: I think we should keep this log, it's useful to know at which point sauce connect was launched.

dev-utils/run-script.js
```diff
@@ -24,42 +24,38 @@
...
-    })
+    return runSauceConnect(callback)
   }
-  console.log('set MODE=saucelabs to launch sauce connect')
+  console.info('Skipping sauce tests, MODE is not set to saucelabs')
```
jahtalab: As far as I can see `launchSauceConnect` doesn't deal with running tests, isn't that correct?

Maybe the log should be moved to `runSauceTests`

dev-utils/run-script.js
```diff
@@ -24,42 +24,38 @@
...
 }
 
-function launchSauceConnect() {
+/**
+ * Checks if the MODE is set to "saucelabs" and decides where to run the
+ * corresponding callback function
+ */
+function launchSauceConnect(callback = () => {}) {
   if (sauceLabs) {
-    return runSauceConnect(sauceConnectOpts, () => {
```
jahtalab: I think this is a better place to pass the sauceConnect options.

dev-utils/test-utils.js
```diff
@@ -160,7 +161,8 @@ function startSelenium(callback, manualStop) {
   )
 }
 
-function runSauceConnect(config, callback) {
+function runSauceConnect(callback) {
+  const config = getSauceConnectOptions()
```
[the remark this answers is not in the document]
vigneshshanmugam: just that the config option is not dynamic and dont see it changing at any point in time.

packages/rum/package.json
vigneshshanmugam: good catch 👍

dev-utils/run-script.js
vigneshshanmugam: sauce connect already logs it when connected, so we were double logging the same stuff hence removed it . Check this example log

```
Starting sc with args: -u XXXXXXXX -k XXXXXXXX --no-ssl-bump-domains all --tunnel-identifier 752.3 --readyfile /tmp/sc-launcher-readyfile
Creating tunnel with Sauce Labs
Detected sc ready
Testing tunnel ready
Sauce connect ready
```

dev-utils/test-servers.js
```diff
@@ -26,6 +26,9 @@
 const express = require('express')
 const serveIndex = require('serve-index')
 const { join } = require('path')
+const {
+  runIntegrationTest
+} = require('@elastic/apm-rum/test/e2e/integration-test')
```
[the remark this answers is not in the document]
vigneshshanmugam: Let me move those to the top folder.

dev-utils/run-script.js
vigneshshanmugam: launchSauceConnect determines if the test should be run on saucelabs or not, so ideally the log should be present here.

dev-utils/test-utils.js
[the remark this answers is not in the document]
jahtalab: IMO it's better to keep this function generic. I don't see any benefit in converting a parameter into one that accesses global config.

dev-utils/run-script.js
jahtalab: I'm not sure I follow, how exactly `launchSauceConnect` decides the tests should be run on saucelabs or not?

vigneshshanmugam: checking the `saucelabs` flag using mode. https://github.com/elastic/apm-agent-rum-js/blob/1ee93d8cbb4b757e1f432de7f12c5ea4b0db1c98/dev-utils/run-script.js#L54

dev-utils/test-utils.js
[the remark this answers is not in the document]
vigneshshanmugam: I can revert it though. So move the sauceOpts to launchSauceConnect?

dev-utils/run-script.js
jahtalab: But we also have this one which checks for `sauceLabs`:
https://github.com/elastic/apm-agent-rum-js/blob/1ee93d8cbb4b757e1f432de7f12c5ea4b0db1c98/dev-utils/run-script.js#L78

dev-utils/test-utils.js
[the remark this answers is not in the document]
jahtalab: Yeah, I think that's better! Thanks!

dev-utils/run-script.js
vigneshshanmugam: Ah that one is for deciding between to run Unit tests on local or unit, e2e in saucelabs. 

The reason why I had a inner condition is because we cannot determine what to run depending on local/ci environment without checking the mode. We can do two things

1. The change present right now
2. Check for saucelabs flag on each function (`runUnitTests` and `runSauceTests`) and call launchSauceConnect accordingly

WDYT?

jahtalab: I think option 2. makes sense, but let's discuss it separately.

dev-utils/test-servers.js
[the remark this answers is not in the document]
jahtalab: I think you forgot about moving `runIntegrationTest` to the top folder dev-utils. Or am I missing something?

PR #321, opened 2019-07-11
chore(docs): add react integration documentation

jahtalab:


docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+
+The React integration package provides two approaches to instrumenting your application:
+
+===== Instrumenting application routes
+
+To instrument the application routes, you can use `ApmRoute` component provided in the package. 
+`ApmRoute` creates a transaction that has the path of the `Route` as its name and has `route-change`
+as its type.
+
+NOTE: Currently `ApmRoute` only supports applications using `react-router` library.
```
vigneshshanmugam: link to the react-router lib would be good addition.

docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
+[[react-integration]]
+=== React integration
+
+This document covers how to use React integration with Real User Monitoring JavaScript agent.
```
vigneshshanmugam:
```suggestion
This section covers how to use Real User Monitoring JavaScript agent on web applications using React library.
```

docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+----
+
+Then, you should replace `Route` components from the `react-router` library
+with `ApmRoute`. You can use `ApmRoute` in any of the routes that you would like to monitor,
+ therefore, you don't have to change all of your routes:
+
+
+[source,js]
+----
+class App extends React.Component {
```
vigneshshanmugam: nit: change indent size to 2 instead of 4

docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+import { withTransaction } from '@elastic/apm-rum-react'
+----
+
+
+Then, you can use `withTransaction` to decorate your React components,
+either using the decorator notation:
+
+[source,js]
+----
+@withTransaction('AboutComponent', 'component')
```
vigneshshanmugam: Let's remove the decorator pattern here, Decorators is still at stage 2 and we should not be encouraging people to use that. If we do that, we have to also tell users how to enable that as well. Lets keep it simple and only add function example.

docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+==== Installing Elastic APM React package
+
+Install the `@elastic/apm-rum-react` package as a dependency to your application:
+
+[source,bash]
+----
+npm install @elastic/apm-rum-react --save
+----
+
+==== Instrumenting your React application
```
bmorelli25: Let's add float tags to these headers. They'll render correctly now, but if we add anchors later on they will break 

```suggestion
[float]
==== Instrumenting your React application
```

docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+
+The React integration package provides two approaches to instrumenting your application:
+
+===== Instrumenting application routes
```
bmorelli25:
```suggestion
[float]
===== Instrumenting application routes
```

docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+                />
+                <ApmRoute path="/home" component={HomeComponent} />
+                <Route path="/about" component={AboutComponent} />
+            </div>
+        )
+    }
+}
+----
+
+===== Instrumenting individual React components
```
bmorelli25:
```suggestion
[float]
===== Instrumenting individual React components
```

docs/react-integration.asciidoc
```diff
@@ -0,0 +1,97 @@
...
+
+[source,js]
+----
+class AboutComponent extends React.Component { }
+export default withTransaction('AboutComponent', 'component')(AboutComponent)
+----
+
+
+NOTE: `withTransaction` accepts two parameters, "transaction name" and "transaction type". 
+If these parameters are not provided the defaults documented in <<transaction-api, Transaction API>> will be used.
```
bmorelli25:
```suggestion
If these parameters are not provided, the defaults documented in <<transaction-api, Transaction API>> will be used.
```

docs/index.asciidoc
```diff
@@ -27,4 +27,6 @@ include::./distributed-tracing-guide.asciidoc[]
 
 include::./opentracing.asciidoc[]
 
-include::./release-notes.asciidoc[]
\ No newline at end of file
+include::./release-notes.asciidoc[]
+
+include::./framework-integrations.asciidoc[]
```
bmorelli25: I would like to see this moved elsewhere in the docs. I don't think at the end makes sense. Perhaps after configuraiton?

docs/react-integration.asciidoc
jahtalab: I don't think of it as encouraging people to use it to be honest, it's simply a choice users can make. I've seen plenty of setups that are already using decorators, similar to how most of the syntax in a JSX file is either not standard at all or needs transpiling. I will add a note about on usage thought.

jahtalab: We decide to remove the decorator usage for now, we can revisit once we support typescript.

jahtalab:
Thanks @vigneshshanmugam and @bmorelli25 , I've address the comments!

PR #322, opened 2019-07-12
chore(coverage): adding code coverage for all packages

vigneshshanmugam:
+ testing code coverage since we have Jenkins pipeline already integrated now. 
+ fixes #153 

## Highlights
- Publish to `https://codecov.io/gh/elastic/apm-agent-rum-js`
- https://github.com/elastic/apm-agent-rum-js/pull/322/commits/fa5a844063726348f24b6f8a59952f72258fefcb enable the coverage in the Jenkins pipeline.
- It does provide a new stage to validate the code coverage and push to codecov.io using the stack `7.0.0` and without any goal (as it will run for all the packages) rather than the whole configuration matrix of stacks and scopes.
- It does extend the puppeteer docker node to run the npm goal which has been specified in the `GOAL` env variable. Therefore, the same dockerised infra can be used for running `npm run test` and `npm run coverage`.

## How to test locally

```
## You need docker and docker-compose
export STACK_VERSION=7.0.0
export APM_SERVER_URL='http://apm-server:8200'
export APM_SERVER_PORT='8200'
export GOAL="coverage"
export JENKINS_URL="https://jenkins.elastic.co/"
.ci/scripts/test.sh
```

PR #323, opened 2019-07-12
ci(jenkins): enable junit reporting to be consumed by the pipeline

v1v:
## Highlights
- Enable JUnit reporting in the Jenkins pipeline.

## Tests
- See [this](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Fapm-agent-rum-mbp%2FPR-323/detail/PR-323/2/tests/)

image

## How to test locally

```
## You need docker and docker-compose
export STACK_VERSION=7.0.0
export APM_SERVER_URL='http://apm-server:8200'
export APM_SERVER_PORT='8200'
export GOAL="test"
export JENKINS_URL="https://jenkins.elastic.co/"
.ci/scripts/test.sh
```

### DIFF
--- a/dev-utils/karma.js
+++ b/dev-utils/karma.js
@@ -127,12 +127,6 @@ function prepareConfig(defaultConfig) {
       ') Elastic Stack ' +
       process.env.STACK_VERSION
 
-    defaultConfig.plugins.push('karma-junit-reporter')
-    defaultConfig.junitReporter = {
-      outputDir: 'reports'
-    }
-    defaultConfig.reporters.push('junit')
-
     defaultConfig.plugins.push('karma-chrome-launcher')
     defaultConfig.browsers = ['ChromeHeadlessNoSandbox']
     defaultConfig.customLaunchers = {
@@ -155,9 +149,10 @@ function prepareConfig(defaultConfig) {
    *  Add coverage reports and plugins required for all environments
    */
   if (defaultConfig.coverage) {
+    defaultConfig.plugins.push('karma-junit-reporter')
     defaultConfig.plugins.push('karma-coverage')
     defaultConfig.reporters.push('coverage')
-
+    defaultConfig.reporters.push('junit')
     const babelPlugins = defaultConfig.webpack.module.rules[0].options.plugins
     babelPlugins.push('istanbul')
 
@@ -165,6 +160,16 @@ function prepareConfig(defaultConfig) {
       includeAllSources: true,
       reporters: [{ type: 'lcov' }, { type: 'text-summary' }],
       dir: 'coverage/'
+    }
+    defaultConfig.junitReporter = {
+      outputDir: 'reports',
+      outputFile: undefined,
+      suite: '',
+      useBrowserName: true,
+      nameFormatter: undefined,
+      classNameFormatter: undefined,
+      properties: {},
+      xmlVersion: null
     }
   }
 
--- a/dev-utils/karma.js
+++ b/dev-utils/karma.js
@@ -129,14 +129,7 @@ function prepareConfig(defaultConfig) {
 
     defaultConfig.plugins.push('karma-junit-reporter')
     defaultConfig.junitReporter = {
-      outputDir: 'reports',
-      outputFile: undefined,
-      suite: '',
-      useBrowserName: true,
-      nameFormatter: undefined,
-      classNameFormatter: undefined,
-      properties: {},
-      xmlVersion: null
+      outputDir: 'reports'
     }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -82,6 +82,7 @@
     "karma-failed-reporter": "^0.0.3",
     "karma-firefox-launcher": "^1.0.1",
     "karma-jasmine": "^1.1.0",
+    "karma-junit-reporter": "^1.2.0",
```
vigneshshanmugam: when you npm install, you would see `package-lock.json` getting updated. Can you check in that as well.

dev-utils/karma.js
```diff
@@ -130,6 +130,18 @@ function prepareConfig(defaultConfig) {
     defaultConfig.plugins.push('karma-firefox-launcher')
     defaultConfig.browsers.push('Firefox')
   } else if (isJenkins && !isSauce) {
+    defaultConfig.plugins.push('karma-junit-reporter')
+    defaultConfig.junitReporter = {
```
vigneshshanmugam: shall we merge this PR after #322, Did a small refactoring to keep it simple.

v1v: Yep, it sounds much better to do it sequentially

package.json
v1v: done. https://github.com/elastic/apm-agent-rum-js/pull/323/commits/ebef14e8efb37b3f80a74e610c04f4d702442581

vigneshshanmugam:
@v1v Can you rebase with the master branch, I merged the other PR.

dev-utils/karma.js
```diff
@@ -130,6 +130,18 @@ function prepareConfig(defaultConfig) {
...
+      outputDir: 'reports',
+      outputFile: undefined,
```
vigneshshanmugam: let's remove all the default configuration and pass only the options necessary.

dev-utils/karma.js
```diff
@@ -127,6 +127,12 @@ function prepareConfig(defaultConfig) {
       ') Elastic Stack ' +
       process.env.STACK_VERSION
 
+    defaultConfig.plugins.push('karma-junit-reporter')
```
vigneshshanmugam: guess this line is not required. isn't it a reporter and not a plugin? https://github.com/karma-runner/karma-junit-reporter#configuration

v1v: As far as I see it's required when running the tests locally, as I've just tried without the line `defaultConfig.plugins.push('karma-junit-reporter')` and even moving the `defaultConfig.reporters.push('junit')` line at the same position. See the below error:

```
node-puppeteer         | 15 07 2019 12:08:00.521:ERROR [reporter]: Can not load reporter "junit", it is not registered!
node-puppeteer         |   Perhaps you are missing some plugin?
```


I guess, we are now in a good position to merge it.

vigneshshanmugam: sorry my bad didnt review it properly. we have to add the check if coverage is enabled before doing this. Otherwise reports will be created for normal tests as well. 

```js
if (defaultConfig.coverage) {
 // code goes here
}
```

vigneshshanmugam: you are right 👍

v1v: Uhm, I'm a bit confused now. It seems there were above 1000 tests and since I rebased with the master branch the number of reported tests have dramatically reduced to less than 100.

- Previously - see [here](https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/PR-323/2/testReport/)
- Now - see [here](https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/PR-323/5/testReport/)

is that the expected behavior?

v1v: @vigneshshanmugam 
>> Otherwise reports will be created for normal tests as well.
Not sure what you meant, the suggested approach of adding the junit output under the `if (defaultConfig.coverage) {` doesn't enable the junit reporting though. Therefore the pipeline won't be able to show the tests which have been executed in the `tests` tab view. I guess the coverage should be independent of the junit reporting. We only want to generate the junit reporting within the jenkins context.

So I'll rollback my last changes to be able to report to jenkins if no issues, besides, I need to clarify https://github.com/elastic/apm-agent-rum-js/pull/323#discussion_r303442179 as I want to confirm if the number of executed tests should be above 100 or not

vigneshshanmugam: We have decided to keep the junit tests for tests running on Jenkins.

PR #326, opened 2019-07-15
ci(jenkins): ITs for forked repos with different name is not supported

v1v:
This is caused by https://github.com/jenkinsci/github-branch-source-plugin/pull/235

## Highlights
- ITs for the rum-js agent don't require the repo name as it does fetch the PRs too
  - See https://github.com/elastic/apm-integration-testing/blob/edb0048bb0d3c79e8c8c9349bde7030762ddc7f6/docker/rum/Dockerfile#L23-L26


v1v:
jenkins run the tests

Issue #301, opened 2019-06-26
React router integration


jahtalab:
We need to provide an integration with react router to capture `route-change` transactions.

- [x] Document the API and provide examples of usage
- [x] fix Readme
- [x] Release 0.1.0
- [x] Implement the react-router integration
- [x] Implement the e2e tests
- [x] Implement unit tests

ersel:
Hi @jahtalab , thanks for implementing this integration! I am just giving it a go on a project which uses `react-router`. Is there any reason why `render` prop is not supported?

https://medium.com/@ilovezcd/difference-between-render-and-component-prop-on-react-router-v4-368ca7fedbec

Some of our routes are using `render` instead of `component` and `ApmRoute` fails with a cryptic message saying can not read property display name of undefined.

jahtalab:
Thanks @ersel for using the react-integration! Using the `render` property is not supported with `ApmRoute`, although, it should not throw an error. I have created #328 to address the error as well as update our docs. 

My suggestion is to instrument the route's top level component with `withTransaction`. 
Let me know if this works for you.

ersel:
I have changed my render props to be components for now. Can you explain why `render` prop is not supported?

jahtalab:
@ersel , The reason is mainly due to the fact that we need to integrate with the react component's life-cycle methods, therefore we need to have the component reference. Furthermore, the `render`, `children` properties can be used in complex situation therefore, it's better to use `withTransaction` to instrument the individual components.

Did you try to use `withTransaction` instead of using the component prop?

PR #328, opened 2019-07-15
fix(rum-react): check component in withTransaction

jahtalab:
Add doc note about Route.render usage
Add ApmRoute tests and improve code coverage

### DIFF
--- a/docs/react-integration.asciidoc
+++ b/docs/react-integration.asciidoc
@@ -66,8 +66,8 @@ class App extends React.Component {
 }
 ----
 
-NOTE: `ApmRoute` only instruments component property if provided, in other cases, e.g. using `render` or `children` properties, 
-ApmRoute will only render the Route without instrumenting it, 
+NOTE: `ApmRoute` only instruments the route if component property is provided, in other cases, e.g. using `render` or `children` properties, 
+ApmRoute will only renders the route without instrumenting it, 
 please instrument the individual component using `withTransaction` in these cases instead.
 
 
--- a/docs/react-integration.asciidoc
+++ b/docs/react-integration.asciidoc
@@ -66,6 +66,11 @@ class App extends React.Component {
 }
 ----
 
+NOTE: `ApmRoute` only instruments component property if provided, in other cases, e.g. using `render` or `children` properties, 
+ApmRoute will only render the Route without instrumenting it, 
+please instrument the individual component using `withTransaction` in these cases instead.
+
+
 [float]
 ===== Instrumenting individual React components
 
--- a/packages/rum-react/src/get-apm-route.js
+++ b/packages/rum-react/src/get-apm-route.js
@@ -29,14 +29,20 @@ import { getWithTransaction } from './get-with-transaction'
 
 function getApmRoute(apm) {
   const withTransaction = getWithTransaction(apm)
+
   return class ApmRoute extends React.Component {
     constructor(props) {
       super(props)
       const { path, component: Component } = this.props
       this.ApmComponent = withTransaction(path, 'route-change')(Component)
     }
+
     render() {
-      return <Route {...this.props} component={this.ApmComponent} />
+      if (this.ApmComponent) {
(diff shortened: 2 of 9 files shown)


packages/rum-react/src/get-apm-route.js
```diff
@@ -29,14 +29,20 @@ import { getWithTransaction } from './get-with-transaction'
...
       super(props)
       const { path, component: Component } = this.props
       this.ApmComponent = withTransaction(path, 'route-change')(Component)
     }
+
     render() {
-      return <Route {...this.props} component={this.ApmComponent} />
+      if (this.ApmComponent) {
+        return <Route {...this.props} component={this.ApmComponent} />
+      } else {
```
ersel: Would it make sense to put a `console.warn` here to tell the developer that this route won't be monitored out of the box?

jahtalab: Thanks for point it out! I've the added

packages/rum-react/test/specs/get-apm-route.spec.js
```diff
@@ -0,0 +1,91 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+import 'airbnb-js-shims/target/es2015'
```
vigneshshanmugam: not sure why do we need these shims? Where are we using the new apis which is shimmed as listed here https://github.com/airbnb/js-shims#included-shims

jahtalab: The APIs are used in enzyme, you can see the initial commit's failure in travis.

docs/react-integration.asciidoc
```diff
@@ -66,6 +66,11 @@ class App extends React.Component {
 }
 ----
 
+NOTE: `ApmRoute` only instruments component property if provided, in other cases, e.g. using `render` or `children` properties,
```
vigneshshanmugam: From reading through the doc first, it was not really clear - Probably we can also add what do you mean by component property - https://reacttraining.com/react-router/web/api/Route/component

To rephrase, 
```
`ApmRoute` instruments the route change event only when the route contains a [component property](https://reacttraining.com/react-router/web/api/Route/component). In other cases eg: when using render or children propeties, `ApmRoute` will not instrument the route-change event. 

please instrument the individual component using `withTransaction` in these cases instead.
```

packages/rum-react/test/specs/get-apm-route.spec.js
vigneshshanmugam: sounds good,  can you add a comment for future reference.

docs/react-integration.asciidoc
jahtalab: I've rephrased it a little bit! But let's keep the external links to a minimum, since they might change. I think in this context it's clear what component property is referring to, given that there's an example above it as well.

jahtalab:
Thanks @vigneshshanmugam , I've made some changes, please have another look.

Re. supporting `render` and `children` I've made a comment in #301. In any case let's take the discussion there and move forward with this one.

vigneshshanmugam:
@jahtalab sure 👍

PR #330, opened 2019-07-16
feat(rum-core): improver error message on payload failure

### LINKED ISSUE #310: improve error message when payload is not sent to apm server
+ When the payload is not sent to the APM server due to network error, 4xx, 5xx. The Error message from the server is discarded, Instead we log a custom error with status code. We should also included the server response message on the custom error to make it more meaningful. 

https://github.com/elastic/apm-agent-rum-js/blob/0ca1e5f97fd4204d6e93d099a9e10b01217bf920/packages/rum-core/src/common/apm-server.js#L100-L105

Helps solves issue like this one - https://discuss.elastic.co/t/microsoft-edge-apm-js-failed-sending-transactions-and-not-send-data-by-ie12/188795

vigneshshanmugam:
+ fixes #310

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -80,16 +80,36 @@ class ApmServer {
     })
   }
 
-  _makeHttpRequest(method, url, payload, headers = {}) {
+  _constructError(reason) {
+    const { url, status, responseText } = reason
+    let message = url + ' HTTP status: ' + status
+    if (__DEV__ && responseText) {
+      try {
+        const serverErrors = []
+        const response = JSON.parse(responseText)
+        if (response.errors && response.errors.length > 0) {
+          response.errors.forEach(err => serverErrors.push(err.message))
+          message += ' ' + serverErrors.join(',')
+        }
+      } catch (e) {
+        this._loggingService.debug('Error parsing response from APM server', e)
+      }
+    }
+    return new Error(message)
+  }
+
+  _makeHttpRequest(method, url, payload, headers) {
     return new Promise(function(resolve, reject) {
       var xhr = new window.XMLHttpRequest()
       xhr[XHR_IGNORE] = true
       xhr.open(method, url, true)
       xhr.timeout = 10000
 
-      for (var header in headers) {
-        if (headers.hasOwnProperty(header)) {
-          xhr.setRequestHeader(header, headers[header])
+      if (headers) {
+        for (var header in headers) {
+          if (headers.hasOwnProperty(header)) {
+            xhr.setRequestHeader(header, headers[header])
+          }
         }
       }
 
@@ -98,26 +118,7 @@ class ApmServer {
           const { status, responseText } = xhr
           // An http 4xx or 5xx error. Signal an error.
           if (status === 0 || (status > 399 && status < 600)) {
-            let message = []
-            if (responseText) {
-              try {
-                const response = JSON.parse(responseText)
-                if (response.errors && response.errors.length > 0) {
-                  response.errors.forEach(err => message.push(err.message))
-                }
-              } catch (e) {
-                if (__DEV__) {
-                  loggingService.debug(
-                    'Error parsing response from APM server',
-                    e
-                  )
-                }
-              }
-            }
-            const err = new Error(
-              url + ' HTTP status: ' + status + ' ' + message.join(',')
-            )
-            reject(err)
+            reject({ url, status, responseText })
           } else {
             resolve(responseText)
           }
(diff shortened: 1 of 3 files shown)


packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -179,6 +184,26 @@ describe('ApmServer', function() {
...
+          id: '21312',
+          span_count: 0,
+          duration: 100,
+          type: 'app'
+        }
+      ])
+      .catch(err => {
+        const trimmedMessage = err.message.replace(/\/n/g, '').trim()
+        expect(trimmedMessage).toEqual(
+          'http://localhost:8001/intake/v2/rum/events HTTP status: 400 error validating JSON document against schema: I[#] S[#] doesn\'t validate with "transaction#"\n  I[#] S[#/allOf/1] allOf failed\n    I[#] S[#/allOf/1/required] missing properties: "trace_id"'
```
vigneshshanmugam: we can also remove the  new line indentation and print in single line. WDYT @jahtalab ?

vigneshshanmugam:
APM server error messages are different in some versions which led to test failure. Will find some solutions to that

packages/rum-core/test/common/apm-server.spec.js
jahtalab: I don't have a strong opinion! I think if it is readable in the console, then there's no need to change the APM server response.

vigneshshanmugam: cool, Actually its pretty neat with indentation in console. So ll keep it as it is.

vigneshshanmugam:
@jahtalab good for review now.

packages/rum-core/test/common/fetch-patch.spec.js
```diff
@@ -117,7 +117,7 @@ describe('fetchPatch', function() {
     })
     it('should reset fetchInProgress global state', function(done) {
       expect(globalState.fetchInProgress).toBe(false)
-      window.fetch('http://localhost:54321/').then(done, done)
+      window.fetch('http://localhost:54321/').then(done, () => done())
```
vigneshshanmugam: to avoid jasmine printing deprecation warnings. done(err) was causing issues.

packages/rum-core/src/common/apm-server.js
```diff
@@ -80,39 +80,51 @@ class ApmServer {
...
       xhr.onreadystatechange = function() {
         if (xhr.readyState === 4) {
-          var status = xhr.status
+          const { status, responseText } = xhr
+          // An http 4xx or 5xx error. Signal an error.
           if (status === 0 || (status > 399 && status < 600)) {
-            // An http 4xx or 5xx error. Signal an error.
-            var err = new Error(url + ' HTTP status: ' + status)
-            err.xhr = xhr
+            let message = []
```
jahtalab: I think this should be moved to the calling method where e.g. [here](https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/src/common/apm-server.js#L135). 

is a generic method and should only resolve or reject the promise.

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -89,10 +94,6 @@ describe('ApmServer', function() {
 
   it('should report http errors', function(done) {
     var apmServer = new ApmServer(configService, loggingService)
-    configService.setConfig({
-      serverUrl: 'http://localhost:54321',
-      serviceName: 'test-service'
-    })
```
jahtalab: Why is this removed? This test is for the case when the apm-server is unreachable!

packages/rum-core/test/common/fetch-patch.spec.js
jahtalab: shouldn't it be `.then(done, fail)` instead.

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -179,8 +180,32 @@ describe('ApmServer', function() {
...
+          'validating JSON document against schema: I[#] S[#] doesn\'t validate with "transaction#"'
+        )
+        expect(err.message).toContain('missing properties: "trace_id"')
+      })
+      .then(done)
+  })
+
   it('should report http errors for queued errors', function(done) {
-    spyOn(loggingService, 'warn').and.callThrough()
+    spyOn(loggingService, 'warn')
```
jahtalab: Is there any reason to remove the `callThrough`? I think at least it ensures there are no errors thrown in the actual method.

packages/rum-core/test/common/apm-server.spec.js
jahtalab: The same with two other instances below.

packages/rum-core/test/common/apm-server.spec.js
vigneshshanmugam: You would see the logs everywhere when the test runs. If there is error then you would see them as args in spy so not really different.

vigneshshanmugam:
@jahtalab JSON.parse is a one time operation and our response is not that big when in terms of error. But I like your suggestion to keep the descriptive error only for debug purposes. Let me iterate this PR.

packages/rum-core/test/common/fetch-patch.spec.js
vigneshshanmugam: Not really,  that url never exists so its an error

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: Not really, the only job of the method is to construct the error based on status code and we cannot replicate the error creation logic multiple times since we have error handling logic for transaction/error.

jahtalab: We don't have to duplicate the error creation we can put it in a function and call that, but we should call that function in the caller of `_makeHttpRequest`

packages/rum-core/test/common/apm-server.spec.js
vigneshshanmugam: Forgot to revert those, was testing IE bug. I will address those with other review comments.

packages/rum-core/src/common/apm-server.js
```diff
@@ -97,22 +115,17 @@ class ApmServer {
...
+            resolve(responseText)
           }
         }
       }
 
-      xhr.onerror = function(err) {
-        reject(err)
-      }
-
+      xhr.onerror = err => reject(err)
```
jahtalab: Should we also reject here with the same structure as above?

PR #337, opened 2019-07-18
fix(rum-core): check ignoreTransactions config value

jahtalab:


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -219,15 +219,16 @@ class TransactionService {
 
   shouldIgnoreTransaction(transactionName) {
     var ignoreList = this._config.get('ignoreTransactions')
-
-    for (var i = 0; i < ignoreList.length; i++) {
-      var element = ignoreList[i]
-      if (typeof element.test === 'function') {
-        if (element.test(transactionName)) {
+    if (ignoreList) {
```
vigneshshanmugam: can we do ignoreList.length > 0 instead since we expect array value for this config?

jahtalab: The config might not be an array actually, specially if the agent is not initialised! Also the config is provide by the user so it might be anything.

jahtalab: I've added `ignoreList.length` to the if check as well.

PR #348, opened 2019-07-25
chore(travis): remove Build Docs

jahtalab:


jahtalab:
I think we should keep that script for local testing, we should fix that script!

PR #350, opened 2019-07-26
feat(rum-core): reuse transaction if it has atleast one active span

### LINKED ISSUE #341: discard transaction without spans before redefining it
Assume there are multiple routes in an application and we use route component for all of those

```jsx
<ApmRoute path="/a" component={ComponentA}/>
<ApmRoute path="/b" component={ComponentB}/>
```

+ When we click on link to `a`, agent will start capturing the transaction with name `/a` and type `route-change`. 
+ If the user aborts that particular request and clicks on link `b`, We will reuse the previous transaction and redefine the name of the transaction with `/b` if the transaction is within the reusability threshold.

We can add an extra check to discard the previous transaction if there are no active spans present on it. Reusing the transaction without any spans would simply make the transaction duration behave incorrectly.

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#341

### DIFF
--- a/packages/rum-core/src/performance-monitoring/transaction.js
+++ b/packages/rum-core/src/performance-monitoring/transaction.js
@@ -68,14 +68,14 @@ class Transaction extends SpanBase {
    * To avoid stale transaction capturing everything, the active
    * transaction can be reused only on the follow conditions
    * - current duration of transaction be must be less than the threshold
-   * - should contain atleast one active span
+   * - should contain atleast one active or completed span
    */
   canReuse(threshold = REUSABILITY_THRESHOLD) {
     return (
       !!this.options.canReuse &&
       !this.ended &&
       performance.now() - this._start < threshold &&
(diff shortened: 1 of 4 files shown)


vigneshshanmugam:
@jahtalab Valid point - I already thought about that, But IMO, it's still better than showing stale transaction with without any spans in most cases.

jahtalab:
I think we should investigate more to see if this can be solved within ApmRoute. I agree the current conditions for reusing is not perfect but reusing a transaction is meant to be used in scenarios like redirect. I think even the history patch would not be reused with this solution.

Also, the original issue seem an edge case to me, since we do detectFinish on component mount as well as the unmount!

vigneshshanmugam:
Had a discussion with @jahtalab, We have decided to move this functionality to rum-react package and decide the transaction reusability in the ApmRoute component.

PR #351, opened 2019-07-26
feat(rum-core): provide debug logs when transaction was discarded

### LINKED ISSUE #300: Provide debug logs when transaction is discarded
Providing appropriate debug logs when we discard a transaction (when we don't send it to APM server). an example of this is when the transaction has 0 spans.

vigneshshanmugam:
+ fixes #300

### DIFF
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -186,16 +186,24 @@ class PerformanceMonitoring {
   }
 
   filterTransaction(tr) {
-    var performanceMonitoring = this
-    var transactionDurationThreshold = this._configService.get(
+    const transactionDurationThreshold = this._configService.get(
       'transactionDurationThreshold'
     )
-    var duration = tr.duration()
-    if (
-      !duration ||
-      duration > transactionDurationThreshold ||
-      !tr.spans.length
-    ) {
+    const duration = tr.duration()
+    if (!duration || duration > transactionDurationThreshold) {
+      this._logginService.debug(
+        `Transaction was discarded! transaction duration: ${duration} contains falsy value or it is greater than transactionDurationThreshold: ${transactionDurationThreshold} config`
+      )
+
+      return false
+    }
+
+    if (tr.spans.length === 0) {
+      if (__DEV__) {
+        this._logginService.debug(
+          `Transaction was discarded! transaction does not include any spans`
+        )
+      }
       return false
     }
 
@@ -207,19 +215,17 @@ class PerformanceMonitoring {
       tr.resetSpans()
     }
 
-    var browserResponsivenessInterval = this._configService.get(
+    const browserResponsivenessInterval = this._configService.get(
       'browserResponsivenessInterval'
     )
-    var checkBrowserResponsiveness = this._configService.get(
+    const checkBrowserResponsiveness = this._configService.get(
       'checkBrowserResponsiveness'
     )
 
     if (checkBrowserResponsiveness && !tr.isHardNavigation) {
-      var buffer = performanceMonitoring._configService.get(
-        'browserResponsivenessBuffer'
-      )
+      const buffer = this._configService.get('browserResponsivenessBuffer')
 
-      var wasBrowserResponsive = performanceMonitoring.checkBrowserResponsiveness(
+      const wasBrowserResponsive = this.checkBrowserResponsiveness(
         tr,
         browserResponsivenessInterval,
         buffer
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -186,16 +186,24 @@ class PerformanceMonitoring {
...
-    var duration = tr.duration()
-    if (
-      !duration ||
-      duration > transactionDurationThreshold ||
-      !tr.spans.length
-    ) {
+    const duration = tr.duration()
+    if (!duration || duration > transactionDurationThreshold) {
+      this._logginService.debug(
+        `Transaction was discarded! transaction duration: ${duration} contains falsy value or it is greater than transactionDurationThreshold: ${transactionDurationThreshold} config`
```
axw: What are the conditions under which a transaction's duration is falsy? I think there are two:

 - duration is zero
 - transaction either wasn't started or wasn't ended (or neither; also, not sure if it's actually possible to obtain a transaction without starting it?)

Would it be easy enough to restate the message in those terms? It saves the user understanding the implementation details.

axw: Nit: "transaction" should be capitalised, since it follows an exclamation mark.

EDIT: ah I see you're being consistent with existing logging. Perhaps defer this and fix at once another time.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -186,16 +186,24 @@ class PerformanceMonitoring {
...
+      )
+
+      return false
+    }
+
+    if (tr.spans.length === 0) {
+      if (__DEV__) {
```
axw: Why only if `__DEV__`?

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -204,7 +205,32 @@ describe('PerformanceMonitoring', function() {
...
+    })
+    spyOn(logger, 'debug').and.callThrough()
+    const transaction1 = new Transaction()
+    transaction1.end()
+    transaction1._start = 0
+    transaction1._end = 201
+    expect(transaction1.duration()).toBe(201)
+    expect(performanceMonitoring.filterTransaction(transaction1)).toBe(false)
+    expect(logger.debug).toHaveBeenCalledWith(
+      'Transaction was discarded! transaction duration: 201 contains falsy value or it is greater than transactionDurationThreshold: 200 config'
```
axw: > Transaction was discarded! transaction duration: 201 contains falsy value or it is greater than transactionDurationThreshold: 200 config

Is reads pretty awkwardly. I think it'd be easier to read as:

> Transaction was discarded! transaction duration (201) is greater than the transactionDurationThreshold configuration (200)

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: production bundle size will get bigger, so we remove those debug logs and save bytes on our final bundle size. Optimization

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
vigneshshanmugam: Hmmm, can split those two if statements and can do this instead 👍

vigneshshanmugam: Done

packages/rum-core/src/performance-monitoring/performance-monitoring.js
axw: Yeah I was wondering why here and not elsewhere. I suppose this one is a bit more unusual.

axw: But now I see you've added `__DEV__` in the other spot too, so I retract my question ;)

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -186,16 +186,39 @@ class PerformanceMonitoring {
...
-    ) {
+    const duration = tr.duration()
+
+    if (!duration) {
+      if (__DEV__) {
+        let message = 'Transaction was discarded! '
+        if (duration === 0) {
+          message += `Transaction duration is 0`
+        } else {
+          message += `Transaction wasn't ended`
```
axw: Nit: probably better not to use a contraction. We don't in other log messages.

PR #360, opened 2019-07-31
feat(rum):add instrument flag to toggle instrumentations

### LINKED ISSUE #192: Add config option to disable auto instrumentation
This option needs to disable automatically creating spans and transactions but the user should be able to create transactions and spans manually (or through the OpenTracing bridge). See https://github.com/elastic/apm/issues/49 for more detail.

relevant discuss issues:
* https://discuss.elastic.co/t/custom-transaction-for-single-page-applications-route-changes/190585/

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#192 
+ Constants fields for different sources are normalized and matches with events to avoid bloating the code.  
+ Separated the tasks in to individual events for AJAX, Fetch and History calls.

### DIFF
--- a/packages/rum-core/src/common/service-factory.js
+++ b/packages/rum-core/src/common/service-factory.js
@@ -26,7 +26,7 @@
 import ApmServer from './apm-server'
 import ConfigService from './config-service'
 import LoggingService from './logging-service'
-import { ON_CONFIG_CHANGE } from './constants'
+import { CONFIG_CHANGE } from './constants'
 
 class ServiceFactory {
   constructor() {
@@ -67,7 +67,7 @@ class ServiceFactory {
     }
 
     setLogLevel(loggingService, configService)
-    configService.events.observe(ON_CONFIG_CHANGE, function() {
+    configService.events.observe(CONFIG_CHANGE, function() {
       setLogLevel(loggingService, configService)
     })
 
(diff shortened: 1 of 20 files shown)


packages/rum-core/src/common/service-factory.js
```diff
@@ -52,20 +49,12 @@ class ServiceFactory {
         serviceFactory.getService('LoggingService')
       )
     })
-
-    this.registerServiceInstance('PatchUtils', patchUtils)
```
vigneshshanmugam: was never used, Let's add only if its required. We expose most of the modules from @elastic/apm-rum-core, So if we need those, We can directly . import and use . it.

packages/rum-core/src/common/service-factory.js
```diff
@@ -52,20 +49,12 @@ class ServiceFactory {
...
-    this.registerServiceInstance('Utils', utils)
   }
-  init() {
-    if (this.initialized) {
```
axw: This doesn't seem to be related to the rest of the changes - is it necessary? This change seems like something that could have subtle ramifications, so I'd prefer if it could be done separately if possible.

vigneshshanmugam: Sure will separate it out. Thanks for the review @axw

PR #366, opened 2019-08-06
fix(rum): do not polyfill the global Promise variable

### LINKED ISSUE #365: remove patching of global promise from bootstrap
+ Since we patch the global Promise API from `es6-promise` on unsupported browsers, users of `bluebird` are running in to issues. 

```
window.Promise = require('bluebird')
```

Repro - https://github.com/ronkorland/apm-issue-example

vigneshshanmugam:
+ fixes #365 
+ Use the promise library as a `ponyfill` to support older browsers instead of patching the global `window.Promise` variable. 

**Patching the global environment is not ideal for libraries since the users might be interested in patching with their library of choice**

### DIFF
--- a/packages/rum/test/specs/index.spec.js
+++ b/packages/rum/test/specs/index.spec.js
@@ -30,6 +30,7 @@ import { getGlobalConfig } from '../../../../dev-utils/test-config'
 
 describe('index', function() {
   const globalConfig = getGlobalConfig()
+  const { serverUrl, serviceName } = globalConfig.agentConfig
   var originalTimeout
 
   beforeEach(function() {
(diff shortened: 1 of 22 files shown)


packages/rum/test/specs/index.spec.js
```diff
@@ -97,4 +98,29 @@ describe('index', function() {
       }
     }
   })
+
+  it('should not throw erorr on global Promise patching', () => {
```
axw:
```suggestion
  it('should not throw error on global Promise patching', () => {
```

vigneshshanmugam:
Thanks @axw for the review

PR #382, opened 2019-08-16
fix(rum): log unsupported message only on browser environment

### LINKED ISSUE #363: NextJS + React (server-side rendered) Issues
We're running into an issue with APM-RUM. It's pretty straight forward, I'm sure I must be missing something, personally, but any help would be appreciated.

Stated in reductionist terms:

```
_document.js is server-side only and is used to change the initial server-side rendered document markup.

_app.js is client-side only and is used to initialize pages.
```

When I run the app, I get the following log in my shell/terminal:
`APM: Platform is not supported!`

The following code is nested within the code block for the `render()` method within the _app.js page of our NextJS codebase.

```
    if (isClient()) {
      const apm = initApm({
        serviceName: 'NEXT-REACT',
        serverUrl: 'REDACTED_ADDRESS:443',
        metricsInterval: 1,
        transactionSampleRate: 1,
      });
    }
```

To be even more detailed - 

We are using a custom server.js file with this nextJS setup, and in that setup, we also use APM (not RUM) on that server. The server serves the NextJS Server-Side rendered application.

Any help or guidance would be much appreciated! Thank you :)

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#363

### DIFF
--- a/packages/rum/src/bootstrap.js
+++ b/packages/rum/src/bootstrap.js
@@ -37,7 +37,11 @@ export default function bootstrap() {
   if (isPlatformSupported()) {
     patchAll()
     enabled = true
-  } else {
+  } else if (typeof window !== 'undefined') {
+    /**
+     * Print this error message only on the browser console
+     * on unsupported browser versions
+     */
     console.log('APM: Platform is not supported!')
   }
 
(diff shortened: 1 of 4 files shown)


vigneshshanmugam:
@hamid added test for node as well.

PR #391, opened 2019-08-22
ci(jenkins): avoid using the any agent to skip running in the master-worker

v1v:
## Highlights
- When working with ephemeral workers the only agent which might be always up is the `master` one, besides the static workers if any. So far, we don't have any after moving to windows ephemeral workers.
- This particular change will help to scale up when the build queue is long but will increase a bit the wait time.
- In other words, there are three major concepts regarding the overall time:
  - Waste of time for resources during the build
  - Waste of time for resources in the queue
  - Real build time.
- This particular fix will reduce the time for the first and second one.
- GIT_SHA rather than GIT_BASE_COMMIT as the last one is caused when doing a merge within a PR.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,7 @@ it is need as field to store the results of the tests.
 @Field def rumTasksGen
 
 pipeline {
-  agent { label 'linux && immutable' }
+  agent none
   environment {
     REPO = 'apm-agent-rum-js'
     BASE_DIR = "src/github.com/elastic/${env.REPO}"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -11,6 +11,12 @@ it is need as field to store the results of the tests.
 */
 @Field def rumTasksGen
 
+/**
+This is the git commit sha which it's required to be used in different stages.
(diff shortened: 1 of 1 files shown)


vigneshshanmugam:
@v1v Seems like the test is running for more than 2 hours, is this a sideeffect of this change?

v1v:
> @v1v Seems like the test is running for more than 2 hours, is this a sideeffect of this change?

Not really, for some reason that particular build got stuck when archiving the artifacts in the master instance. See [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Fapm-agent-rum-mbp/detail/PR-391/1/pipeline/121#log-183). It might be related to the number of concurrent builds which were triggered at the same time from other projects.

I guess, I did stress a bit the CI with a dozen of different builds (one per apm agent/server and so on) which might affect the overall times, but that's not a common case.

I just triggered a new build, so let's see how it behaves.

Besides, this particular change does remove any bottlenecks with the master-worker which it seems it was the one causing some slowness.

v1v:
Travis GitHub check is not required for this particular PR. So I'll merge it now

PR #392, opened 2019-08-26
fix(rum-react): respect active flag in react integration

### LINKED ISSUE #386: framework integrations should respect active flag
+ React router integration `<ApmRoute>` should respect active flag from the agent config and should disable the route integration completely. 

Faced this issue when doing the Kibana integration when both the `init` call and `<ApmRoute>` are at different places and It would be good to do a noop when the agent is inactive and we don't need to deal with doing different bundles when agent is activated vs not. 

```
Disabled - ApmRoute =  Route

Enabled - ApmRote = ApmRoute + Route
```

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#386 
+ blocks the kibana integration with RUM.

### DIFF
--- a/packages/rum-react/src/get-with-transaction.js
+++ b/packages/rum-react/src/get-with-transaction.js
@@ -47,8 +47,16 @@ function isReactClassComponent(Component) {
 function getWithTransaction(apm) {
   return function withTransaction(name, type) {
     return function(Component) {
+      const configService = apm.serviceFactory.getService('ConfigService')
+      const loggingService = apm.serviceFactory.getService('LoggingService')
+      if (!configService.isActive()) {
+        loggingService.warn(
+          `RUM agent is inactive, route-change transaction is not instrumented`
+        )
+        return Component
+      }
+
       if (!Component) {
-        const loggingService = apm.serviceFactory.getService('LoggingService')
         loggingService.warn(
           `${name} is not instrumented since component property is not provided`
         )
--- a/packages/rum-react/src/get-with-transaction.js
+++ b/packages/rum-react/src/get-with-transaction.js
@@ -48,15 +48,12 @@ function getWithTransaction(apm) {
   return function withTransaction(name, type) {
     return function(Component) {
       const configService = apm.serviceFactory.getService('ConfigService')
-      const loggingService = apm.serviceFactory.getService('LoggingService')
       if (!configService.isActive()) {
-        loggingService.warn(
-          `RUM agent is inactive, route-change transaction is not instrumented`
-        )
         return Component
       }
 
       if (!Component) {
+        const loggingService = apm.serviceFactory.getService('LoggingService')
         loggingService.warn(
           `${name} is not instrumented since component property is not provided`
         )
--- a/packages/rum-react/test/specs/get-apm-route.spec.js
+++ b/packages/rum-react/test/specs/get-apm-route.spec.js
@@ -38,22 +38,29 @@ import { MemoryRouter as Router, Route } from 'react-router-dom'
 import { ApmBase } from '@elastic/apm-rum'
 import { createServiceFactory } from '@elastic/apm-rum-core'
 import { getApmRoute } from '../../src/get-apm-route'
+import { getGlobalConfig } from '../../../../dev-utils/test-config'
 
 function Component(props) {
   return <h1>Testing, {props.name}</h1>
 }
 
 describe('ApmRoute', function() {
-  let serviceFactory, configService
+  const { serverUrl, serviceName } = getGlobalConfig().agentConfig
+  let serviceFactory, apmBase
 
   beforeEach(() => {
     serviceFactory = createServiceFactory()
-    configService = serviceFactory.getService('ConfigService')
-    configService.init({ active: true })
+    apmBase = new ApmBase(serviceFactory, false)
+    apmBase.init({
+      active: true,
+      serverUrl,
+      serviceName,
+      disableInstrumentations: ['page-load', 'error']
+    })
   })
 
   it('should work Route component', function() {
-    const ApmRoute = getApmRoute(new ApmBase(serviceFactory))
+    const ApmRoute = getApmRoute(apmBase)
 
     const rendered = mount(
       <div>
@@ -75,7 +82,6 @@ describe('ApmRoute', function() {
   })
 
   it('should work with Route render and log warning', function() {
-    const apmBase = new ApmBase(serviceFactory)
     const loggingService = serviceFactory.getService('LoggingService')
     const ApmRoute = getApmRoute(apmBase)
 
--- a/packages/rum-react/test/specs/get-apm-route.spec.js
+++ b/packages/rum-react/test/specs/get-apm-route.spec.js
@@ -44,8 +44,16 @@ function Component(props) {
 }
 
 describe('ApmRoute', function() {
+  let serviceFactory, configService
+
+  beforeEach(() => {
+    serviceFactory = createServiceFactory()
+    configService = serviceFactory.getService('ConfigService')
+    configService.init({ active: true })
+  })
+
   it('should work Route component', function() {
-    const ApmRoute = getApmRoute(new ApmBase(createServiceFactory()))
+    const ApmRoute = getApmRoute(new ApmBase(serviceFactory))
 
     const rendered = mount(
       <div>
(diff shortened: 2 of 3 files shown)


packages/rum-react/src/get-with-transaction.js
```diff
@@ -47,8 +47,16 @@ function isReactClassComponent(Component) {
 function getWithTransaction(apm) {
   return function withTransaction(name, type) {
     return function(Component) {
+      const configService = apm.serviceFactory.getService('ConfigService')
+      const loggingService = apm.serviceFactory.getService('LoggingService')
+      if (!configService.isActive()) {
+        loggingService.warn(
+          `RUM agent is inactive, route-change transaction is not instrumented`
```
jahtalab: The warning here is confusing since the `withTransaction` could be used on its own and not necessarily for a `route-change`!

vigneshshanmugam: We could remove the logging part here completely since we already log agent is inactive from during init phase, WDYT?

jahtalab: Yeah, let's remove this completely 👍 !

vigneshshanmugam: done.

packages/rum-react/test/specs/get-apm-route.spec.js
```diff
@@ -44,8 +44,16 @@ function Component(props) {
 }
 
 describe('ApmRoute', function() {
+  let serviceFactory, configService
+
+  beforeEach(() => {
+    serviceFactory = createServiceFactory()
+    configService = serviceFactory.getService('ConfigService')
+    configService.init({ active: true })
```
jahtalab: Is this required? I mean `active: true` since just calling the init should be enough, active is true by default.
If the tests fail without this, then we have introduced a bug!

packages/rum-react/test/specs/get-with-transaction.spec.js
```diff
@@ -51,26 +52,29 @@ function TestComponent(apm) {
...
 
 describe('withTransaction', function() {
+  const { serverUrl, serviceName } = getGlobalConfig().agentConfig
+  let serviceFactory, configService
+
+  beforeEach(() => {
+    serviceFactory = createServiceFactory()
+    configService = serviceFactory.getService('ConfigService')
+    configService.init({
+      active: true,
```
jahtalab: Same here!

packages/rum-react/test/specs/get-with-transaction.spec.js
```diff
@@ -91,4 +93,14 @@ describe('withTransaction', function() {
...
   })
+
+  it('should not instrument the route when rum is inactive', () => {
+    const transactionService = serviceFactory.getService('TransactionService')
+    spyOn(transactionService, 'startTransaction')
+
+    configService.setConfig({ active: false })
+    TestComponent(new ApmBase(serviceFactory, true))
+
+    expect(transactionService.startTransaction).not.toHaveBeenCalled()
```
jahtalab: Maybe instead of this we should check if the Component is the same as the WrappedComponent!

packages/rum-react/test/specs/get-apm-route.spec.js
vigneshshanmugam: I removed the `apm.init` in favor of this and yes we require this to be set since now we expect the isActive function to return true to start  instrumenting the route https://github.com/elastic/apm-agent-rum-js/blob/9f429dbf6d39e7208ca926140eeec839c5be7a8b/packages/rum-react/src/get-with-transaction.js#L51-L53

packages/rum-react/test/specs/get-with-transaction.spec.js
vigneshshanmugam: answered above.

packages/rum-react/test/specs/get-with-transaction.spec.js
vigneshshanmugam: Sure, We can have both of those checks.

packages/rum-react/test/specs/get-apm-route.spec.js
jahtalab: What is the advantage of this approach vs just using `apm.init`? IMO, apm.init makes more sense since we're passing the apm instance into `get-with-transaction`

vigneshshanmugam: Two things
1. We need to call init on all the tests since we expect the config.isActive to be true to make it work. 
2. `apm.init` does lots of things under the hood like registering error listeners, config validation, perf monitoring which is not required for these specific tests.

jahtalab: For No. 1 we still need to call init on ConfigService, I don't see a difference here. 

I think since ApmBase is a dependency to `get-with-transaction` therefore it needs to be initialised for the test to be actually realistic. e.g. the things that ApmBase does in init might actually affect these tests!

vigneshshanmugam: Are you saying that we have to call apm.init for all of the tests instead of config.init? 

We can do that but we have to stop error listened and onload listeners from initializing otherwise we would have side effects

jahtalab: We don't have to call apm.init for all the tests but I'm ok with calling the init for all of them as well if that makes the tests easier to write.

For the side effects we probably can use flags to disable page load and error handling!

vigneshshanmugam: Yeah that would be equal to what I was going for, Let's do that then.

vigneshshanmugam: I have changed it now, Please have a look once @jahtalab

packages/rum-react/test/specs/get-with-transaction.spec.js
```diff
@@ -79,16 +84,33 @@ describe('withTransaction', function() {
...
+    function Component() {
+      return <h2>Component</h2>
+    }
+    const withTransaction = getWithTransaction(apmBase)
+
+    const WrappedComponent = withTransaction('test-transaction', 'test-type')(
+      Component
+    )
+    expect(WrappedComponent).toEqual(Component)
+    expect(transactionService.startTransaction).not.toHaveBeenCalled()
```
jahtalab: `transactionService.startTransaction` won't be called unless the component is rendered, so this probably always passes regardless of the test!

packages/rum-react/test/specs/get-with-transaction.spec.js
```diff
@@ -51,26 +52,30 @@ function TestComponent(apm) {
...
+      serverUrl,
+      serviceName,
+      disableInstrumentations: ['page-load', 'error']
+    })
+  })
+
   it('should work if apm is disabled or not initialized', function() {
-    TestComponent(new ApmBase(createServiceFactory(), true))
-    TestComponent(new ApmBase(createServiceFactory(), false))
+    TestComponent(new ApmBase(serviceFactory, true))
```
jahtalab: This might be not related to this PR, but I'm not sure why this test passes. We probably should investigate why when we pass disable = true to ApmBase, it still calls startTransaction.

Do you happen to know?
Either way we can address that separately if you prefer.

vigneshshanmugam: I guess it does not matter, component should rendered if APM is enabled/disabled and we don't check if startTransaction is called or not. So I am not sure what is the issue here?

packages/rum-react/test/specs/get-with-transaction.spec.js
vigneshshanmugam: Yes I added the component to get rendered as well in previous line for double checking. 
https://github.com/elastic/apm-agent-rum-js/blob/046b6400a835f539f79fd60ce8a84a77688fd4bc/packages/rum-react/test/specs/get-with-transaction.spec.js#L103

if you comment out the config.isActive check in the source, this test would fail which is expected.

packages/rum-react/test/specs/get-with-transaction.spec.js
jahtalab: We probably can improve this test a bit, let's do it separately!

packages/rum-react/test/specs/get-with-transaction.spec.js
jahtalab: Ok, I thought maybe you intended to test with the component that you created in the test instead instead of the one created in `TestComponent`. Might be a bit better to group the code together so it's clear which checks are intended for those parts.

PR #398, opened 2019-08-27
ci(jenkins): revert none agent

v1v:
This reverts https://github.com/elastic/apm-agent-rum-js/pull/393 and https://github.com/elastic/apm-agent-rum-js/pull/391

## Highlights
- As long as we use the share step `gitCheckout` we might need to use the agent top-level to share the env variables between stages.
- It's not ideal but even though the changes were quite straight, its behavior was not as expected. 
- Let's keep the pipeline stable for now and we will figure out what's going on.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ it is need as field to store the results of the tests.
 
 /**
 This is the git commit sha which it's required to be used in different stages.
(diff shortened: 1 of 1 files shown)


v1v:
Travis check failed but it's not relevant for this particular PR.

Issue #400, opened 2019-08-28
Fix Saucelabs tests failures


jahtalab:
Currently we're seeing a lot of failures on saucelabs which are not related to the test itself. This affects our development process. We might need to adjust the concurrency of tests!

Happens mostly on IE 11 tests.

vigneshshanmugam:
One thing we could do is to stop the matrix tests for Framework integrations on all stacks, So for `apm-rum-react`, we would test it only with a single stack(can be 7.3.0) and we would skip older stacks since we already test `apm-rum` and `apm-rum-core`. React integration as such doesn't do anything special so its not required.

PR #403, opened 2019-08-30
chore(saucelabs): fix script timeout issues in saucelabs

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#400 
+ Increase the timeout for script execution commands to 30 seconds. 
+ handle issues with setTimeout not being supported on Firefox and MSEdge
+ Due to the bundles getting generated for React in dev mode, the size of the bundles are huge and it causes E2E tests to fail in some browsers. 
+ By compiling the bundles in prod mode, the size of the bundles gets reduced from 1.6 MB to ~300Kbs. This should reduce the time required for downloading the script which is 30 seconds by default. 
+ Run react test only on 7.0.0 stack

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,20 +42,12 @@ jobs:
         - SCOPE=@elastic/apm-rum
     - stage: test
       env:
-        - STACK_VERSION=6.5.0
-        - SCOPE=@elastic/apm-rum-react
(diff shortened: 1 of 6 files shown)


vigneshshanmugam:
Confirmed everything with the travis job,  works as expected.

PR #412, opened 2019-09-06
feat(rum-core): breakdown graphs for transaction

### LINKED ISSUE #289: Investigate possible use-cases for breakdown graphs according to the proposal
The breakdown graphs proposal is in review status, so we should investigate if we can utilise it for RUM specific breakdowns

- [ ] Remove the timestamp from the payload sent to APM server
- [x] Provide detailed breakdown docs on each span type. elastic/apm-agent-rum-js#467 
- [x] Consider post transaction-end, creation of the breakdown metrics (#438   )
- [x] Revise the tests according to https://github.com/elastic/apm-agent-rum-js/issues/289#issuecomment-534999842 (elastic/apm-agent-rum-js#438 )
- [x] ~~Disable the breakdown metrics on transaction class by default (#434)~~ -> we use the config option

* https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64/edit?ts=5cd54da0#heading=h.j9rqr5k34x5v

vigneshshanmugam:
+ part of elastic/apm-agent-rum-js#289 
+ Breakdown graphs implementation for transactions based on the transaction types. For page-load transaction, we filter the timings based on the timings exposed through Navigation timing API. 
+ for other transaction types, the span types, subtypes for the transaction are taken in to account to get the breakdown. The duration - total timings of all spans are marked as `app` time.
+ Since RUM agent does not have any custom metrics like the backend agents, we send the metric sets along with the transaction. Once RUM starts supporting custom metrics, we will have dedicated metrics queue and will have different interval for reporting these metrics. 

### page-load transaction

Screenshot 2019-09-05 at 14 49 14

## Route change and other transaction types

Screenshot 2019-09-06 at 17 09 40

### TODO
+ [x] Add tests
+ [x] docs- add breakdownMetrics config


packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,141 @@
...
+
+import { getDuration } from '../common/utils'
+import { PAGE_LOAD } from '../common/constants'
+
+/**
+ * Page load transaction breakdown timings
+ *
+ * Interested events from the Navigation timing API
+ */
+const pageLoadBreakdowns = [
```
vigneshshanmugam: @jahtalab Let me know if you have any ideas on breaking down the page-load transaction. I am trying to keep it simpler for the users and took some timing breakdown from devtools.

packages/rum-core/src/common/apm-server.js
```diff
@@ -207,25 +206,32 @@ class ApmServer {
...
 
   ndjsonErrors(errors) {
-    return errors.map(function(error) {
-      return NDJSON.stringify({ error })
-    })
+    return errors.map(error => NDJSON.stringify({ error }))
+  }
+
+  ndjsonMetricsets(metricsets) {
+    const timestamp = Date.now() * 1000
```
jahtalab: The timestamp should be captured by the Apm Server to avoid using wrong times on the clients as well as the issues with timezone.

packages/rum-core/src/common/apm-server.js
```diff
@@ -207,25 +206,32 @@ class ApmServer {
...
+    return metricsets
+      .map(metricset =>
+        NDJSON.stringify({ metricset: { timestamp, ...metricset } })
+      )
+      .join('')
   }
 
   ndjsonTransactions(transactions) {
-    var ndjsonSpan = this.ndjsonSpan
```
jahtalab: This was a small optimisation to avoid creating new new objects every time in the spans loop

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * Capture breakdown metrics for the transaction based on the
+ * transaction type
+ */
+export function captureBreakdown(transcation) {
+  const breakdowns = []
+  const trDuration = transcation.duration()
+  const { type, sampled } = transcation
+  const transactionBreakDownDetails = transactionBreakdownDetails(transcation)
```
jahtalab: Maybe a use a different name for the const compared to the function!

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,141 @@
...
+  ['domInteractive', 'loadEventEnd', 'Page Render']
+]
+
+function getValue(value) {
+  return { value }
+}
+
+function transactionBreakdownDetails({ name, type }) {
+  return { name, type }
+}
```
jahtalab: I'm not entirely sure these functions are necessary, instead of just doing the same thing in the main function.

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  const timings = window.performance.timing
+  if (type === PAGE_LOAD && timings) {
+    for (let i = 0; i < pageLoadBreakdowns.length; i++) {
+      const current = pageLoadBreakdowns[i]
+      const start = timings[current[0]]
+      const end = timings[current[1]]
+      const duration = getDuration(start, end)
+      if (duration == null) {
```
jahtalab: What about when the duration is 0?

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,141 @@
...
+        continue
+      }
+      breakdowns.push(
+        getSpanBreakdown(
+          {
+            type: current[2],
+            duration: () => duration
```
jahtalab: I'm not sure if it's necessary to create something similar to the Span class, why not just pass these as arguments?

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -232,75 +232,81 @@ function createUserTimingSpans(entries, transactionEnd) {
...
   const perf = window.performance
-  if (transaction.isHardNavigation && perf && perf.timing) {
-    const timings = perf.timing
-    if (transaction.marks && transaction.marks.custom) {
-      var customMarks = transaction.marks.custom
-      Object.keys(customMarks).forEach(key => {
-        customMarks[key] += transaction._start
-      })
-    }
+  if (!(transaction.isHardNavigation && perf && perf.timing)) {
```
jahtalab: Not sure why negating the condition here would be better that having a simple if around the logic that depends on this condition.

packages/rum-core/src/performance-monitoring/capture-hard-navigation.js
```diff
@@ -232,75 +232,81 @@ function createUserTimingSpans(entries, transactionEnd) {
...
+    /**
+     * Add transaction context information from performance navigation timing entry level 2 API
+     */
+    let navigationEntry = perf.getEntriesByType('navigation')
+    if (navigationEntry && navigationEntry.length > 0) {
+      navigationEntry = navigationEntry[0]
+      transaction.addContext({
+        response: getResponseContext(navigationEntry)
+      })
     }
```
jahtalab: The changes in the capture-hard-navigation.js seems to be only moving the code around, is that correct? or am I missing something?

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,141 @@
...
+import { getDuration } from '../common/utils'
+import { PAGE_LOAD } from '../common/constants'
+
+/**
+ * Page load transaction breakdown timings
+ *
+ * Interested events from the Navigation timing API
+ */
+const pageLoadBreakdowns = [
+  ['navigationStart', 'fetchStart', 'Request Stalled'],
```
jahtalab: For the capture navigation timing we base all the times on `fetchStart`

packages/rum-core/src/performance-monitoring/breakdown.js
jahtalab: We should report the duration 0, but we should double check if `duration == null` needs to be reported as well, we need check with UI team.

packages/rum-core/src/performance-monitoring/breakdown.js
jahtalab: Decided to remove the `request stalled`!

Also, decided to reconsider the title of metrics!

packages/rum-core/src/common/apm-server.js
jahtalab: This would be blocker for breakdown graphs, but we can merge it behind feature flag.

packages/rum-core/src/performance-monitoring/breakdown.js
vigneshshanmugam: Not required to report both 0 and 'null' values. Confirmed with Felix.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: created an issue for the server - https://github.com/elastic/apm-server/issues/2700

vigneshshanmugam: Found a way to fix this without getting affected by system clock drift. So its not a blocker anymore.

packages/rum-core/src/common/apm-server.js
```diff
@@ -207,25 +207,32 @@ class ApmServer {
...
 
   ndjsonErrors(errors) {
-    return errors.map(function(error) {
-      return NDJSON.stringify({ error })
-    })
+    return errors.map(error => NDJSON.stringify({ error }))
+  }
+
+  ndjsonMetricsets(metricsets) {
+    const timestamp = microtime()
```
jahtalab: The real issue with creating the timestamp here is that the clients time is sometimes inaccurate (sometimes it's off in years). We have seen this before, even if there are a few clients that have a wrong time it will clearly show up in the data since the APM server would generate indices based on date. Therefore, the users will see indices with dates that are clearly wrong (e.g. in the future) and managing these indices quickly gets out of hand.

jahtalab: I think this is still a blocker for the break down feature.

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,130 @@
...
+import { PAGE_LOAD } from '../common/constants'
+
+/**
+ * Page load transaction breakdown timings
+ *
+ * Interested events from the Navigation timing API
+ */
+const pageLoadBreakdowns = [
+  ['fetchStart', 'requestStart', 'Network'],
+  ['requestStart', 'responseStart', 'TTFB'],
```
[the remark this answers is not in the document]
vigneshshanmugam: I was choosing between TTFB/Server response time. Probably later is better since TTFB might confuse users. I was actually using the chrome devtools timing viewer which more users are familiar with. 
Screenshot 2019-09-13 at 10 19 32

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,130 @@
...
+  ['domLoading', 'domInteractive', 'DOM Processing'],
```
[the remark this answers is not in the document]
vigneshshanmugam: The article talks about using domInteractive as a signal to the `Time to Interactive` metric which is quite different from our measurements. Its a signal that browser has succesfully parsed the sync scripts and the DOM itself. The reason why I did not use `domComplete` is that users might have code that might be running  on `domContentLoaded` which would delay the `domComplte` time and `domInteractive` does not suffer from this pitfall.

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,130 @@
...
+import { getDuration } from '../common/utils'
+import { PAGE_LOAD } from '../common/constants'
+
+/**
+ * Page load transaction breakdown timings
+ *
+ * Interested events from the Navigation timing API
+ */
+const pageLoadBreakdowns = [
+  ['fetchStart', 'requestStart', 'Network'],
```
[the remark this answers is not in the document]
vigneshshanmugam: yes, you are right. Network = DNS + TLS + HTTP time - But I was not sure how to name that since we include the HTTP time in TTFB. 

Redirects is separate beast, There are some problems with capturing those metrics 
+ We need an opt in from previous document that redirects to the current one -> Timing Allow Origin Response header must be present. 
+ If there are multiple chains before loading the document, it will be always set to 0. 

`www.a.com` -> `b.a.com` our redirect time would be lost here. I couldnt find any good read on this other than this issue https://github.com/w3c/navigation-timing/issues/73

packages/rum-core/src/performance-monitoring/breakdown.js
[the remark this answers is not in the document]
axw: Yeah, I agonised over this one a bit. I think in devtools it's clearer since it's a timeline, whereas in our breakdown it'll be freestanding.

packages/rum-core/src/performance-monitoring/breakdown.js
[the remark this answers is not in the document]
axw: As discussed offline, perhaps change the name from `DOM Processing` to `DOM Loading` (or similar), so we don't conflict with the navigation timing terminology used in https://www.w3.org/TR/navigation-timing/#dom-performancetiming-fetchstart, and to signify that it's only for the bit up until domInteractive.

jahtalab:
Decided to disable this feature and separate the docs so that we can merge this one!

packages/rum-core/src/performance-monitoring/breakdown.js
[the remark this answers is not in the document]
vigneshshanmugam: Everything has been moved under Network for now.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: In most of the browsers I tested, some via saucelabs. new object is faster and its a small micro optimization so we dont need to worry about it much. 
https://jsperf.com/new-object-vs-override

packages/rum-core/src/common/utils.js
```diff
@@ -398,6 +398,27 @@ function getPageLoadMarks() {
...
+
+/**
+ * Get number of microseconds elapsed since January 1, 1970 00:00:00 UTC
+ *
+ * On browsers that support DOM High resolution timestamp, this function takes care of
+ * accounting for clock drift
+ */
+function microtime() {
+  let time = Date.now()
+  if (typeof performance.now === 'function') {
```
vigneshshanmugam: we still need this since we are planning to test this feature via extension. comes handy with 7.3 APM server version. Lets remove it once we enable this feature by default on 7.5

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -354,4 +354,16 @@ describe('TransactionService', function() {
...
+
+  it('should not capture breakdown metrics by default', done => {
+    config.events.observe(TRANSACTION_END, function() {
+      expect(tr1.breakdownTimings.length).toBe(0)
+      done()
+    })
+
+    const tr1 = transactionService.startTransaction('test1', 'custom')
+    const span1 = tr1.startSpan('span1', 'app')
+    span1.end()
```
jahtalab: This might generate a span with duration 0, which would make the span to be ignored by the breakdown! 

The same with transaction which could have the duration as 0

vigneshshanmugam: Not really, it would never create with duration 0 since we use performance.now under the hood and its in milliseconds. so it would always be > 0.

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -0,0 +1,129 @@
...
+     */
+    const spans = transaction.spans
+    let childTimings = 0
+    for (let i = 0; i < spans.length; i++) {
+      const span = spans[i]
+      const { type, subType } = span
+      const duration = span.duration()
+      childTimings += duration
+      breakdowns.push(
+        getSpanBreakdown(transactionDetails, type, duration, subType)
```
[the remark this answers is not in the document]
vigneshshanmugam: oops forgot to fix this part. yes we need to handle that case. Will fix it and also add tests for the same.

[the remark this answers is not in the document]
vigneshshanmugam: this is bit more complicated that I thought and also have to account or change our span type for user timing since it includes `app` and transaction breakdown has `app` as transaction type and self time of the transaction is included in that metricset.

packages/rum-core/test/performance-monitoring/breakdown.spec.js
```diff
@@ -0,0 +1,343 @@
...
+  let now
+
+  let originalNowFn
+  beforeAll(() => {
+    originalNowFn = window.performance.now
+    window.performance.now = () => now
+  })
+
+  afterAll(() => {
+    window.performance.now = originalNowFn
```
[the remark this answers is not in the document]
vigneshshanmugam: Yes, I can explain in a call though.

jahtalab:
We decided to move forward with this PR after the minor fixes and revisit if it's possible to make breakdowns into a post process!

PR #414, opened 2019-09-10
fix(rum-angular): proper transaction name for lazy loaded routes

### LINKED ISSUE #413: Incorrect transaction name for lazy loaded child routes
+ When routes are lazy loaded in the application, the path of the child routes are set to their parents. https://angular.io/guide/router#lazy-loading-route-configuration

Might be a bug with the way the path is calculated in the service https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-angular/src/apm-service.ts#L85

Found this issue from the post - https://discuss.elastic.co/t/has-anyone-been-able-to-use-elastic-apm-rum-angular-with-aot/198456

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#413 
+ We traverse the activated router tree using DFS and construct the path along the way eliminating the parent path names which is set to empty string. `firstChild` always reflects the current activated route.

### DIFF
--- a/packages/rum-angular/src/apm-service.ts
+++ b/packages/rum-angular/src/apm-service.ts
@@ -40,7 +40,7 @@ import { Promise } from 'es6-promise'
 export class ApmService {
   static apm: any
 
-  constructor(public router: Router) {}
+  constructor(public router: Router, protected route: ActivatedRoute) {}
 
   init(config) {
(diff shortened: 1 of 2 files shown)


Azuka:
Awesome, @vigneshshanmugam!

Issue #253, opened 2019-04-24
Support fetching agent configuration from the APM Server (Central config management)


jahtalab:
As discussed [here](https://github.com/elastic/apm/issues/76), we should provide support for fetching agent configuration from APM server. Currently, only the `TRANSACTION_SAMPLING_RATE` configuration is provided.

Note: Implementation details are still being worked out by the team. see the above issue for more.

- [x] Provide Docs
- [x] Investigate using `If-None-Match` and `etag`
- [x] ~~Investigate the appropriate interval to fetch the config~~ -> decided against periodic fetching of configs due to overhead.
- [x] Provide tests
- [x] Provide configuration to enable this feature
- [x] This feature should be disable by default (We should make it clear that users need to enable this in the agent)

jalvz:
@jahtalab is unlekely that etags work for you right? we probably should we tie your fetch interval with a caching interval in the apm-server to avoid too many browsers hammering Kibana, it would help a bit to not make it configurable.
Would that work for you? Be conservative, hardcode it, and wait for user feedback?

jahtalab:
Using etags should generally work, there are some restrictions regarding cross-origin requests, e.g. https is required to make it work. I will investigate more and make a proof of concept soon. Worst case scenario is that we would have to add other headers to the response from the APM server and store the etag manually.

In any case, I'd say having an internal cache in APM server is a good approach. It's too excessive to query Elasticsearch for each config fetch given that the rate of fetching is a lot higher compared to the rate of editing.

jalvz:
Right. I was thinking that the traffic will likely be dominated by the number of agents (instead by the polling intervals), so even using etags would not save us a lot...

vigneshshanmugam:
Regarding fetch interval, Did we even decide on polling for collecting the configs from server? I thought we decided to do it only for `page-load` event. Please correct me if my assumption is wrong.

> Seem like 4/5 with the caching solution mentioned could be an acceptable solution. It would mean one additional request to get the config for every first time user and then a 304 request for every subsequent page load. For many websites, that probably means an additional request for the vast majority of page loads.

https://github.com/elastic/apm-dev/issues/474#issuecomment-466392759 

This is a simplistic approach for RUM and would work for majority of the use cases like dynamic configs. For SPA with soft navigation, We could improve the solution simply by having a queue that would let us fetch when the count of transactions goes a bit higher. We are just complicating the use case by polling the apm-server and its pretty expensive on the browser.

[...]

PR #415, opened 2019-09-10
feat: Support central config management

jahtalab:
Part of elastic/apm-agent-rum-js#253

This PR might change depending on the outcome of https://github.com/elastic/apm-server/issues/2646 and https://github.com/elastic/apm-server/issues/2694 issues. We should wait for these before merging!


dev-utils/karma.js
```diff
@@ -43,7 +43,7 @@ const { tunnelIdentifier } = getSauceConnectOptions()
  * Common base config for all the mono repo packages
  */
 const baseConfig = {
-  files: [specPattern],
+  files: [require.resolve('babel-polyfill/dist/polyfill.js'), specPattern],
```
vigneshshanmugam: Any reason why we need to polyfill all shims instead of promise polyfill alone?.

dev-utils/run-script.js
```diff
@@ -89,6 +89,25 @@ function runNodeTests() {
...
 
+var cleanUps = []
+
+function exitHandler(exitCode) {
+  console.log('Running cleanups.')
+  cleanUps.forEach(f => {
+    try {
+      f(exitCode)
+    } catch (e) {
+      console.log(e)
```
vigneshshanmugam: nit: console.error

dev-utils/test-config.js
```diff
@@ -118,10 +119,38 @@ function getBrowserList() {
...
+}
+
+function isVersionInRange(version, min) {
+  var isInRange = true
+  if (version) {
+    var parsedVersion = parseVersion(version)
+    if (min) {
+      var minParsed = parseVersion(min)
+      isInRange =
+        isInRange &&
```
vigneshshanmugam: not sure why we need this `isRange &&` check

dev-utils/test-servers.js
```diff
@@ -25,8 +25,10 @@
 
 const express = require('express')
 const serveIndex = require('serve-index')
+const proxy = require('express-http-proxy')
 const { join } = require('path')
 const { runIntegrationTest } = require('./integration-test')
+const { DEFAULT_APM_SERVER_URL } = require('./test-config')
```
vigneshshanmugam: We should use the `getTestEnvironmentVariables` function and get the serverUrl instead of exporting this constant since it can be changed via env.

package.json
```diff
@@ -73,6 +73,7 @@
     "babel-loader": "^8.0.4",
     "babel-plugin-istanbul": "^5.1.4",
     "babel-plugin-transform-typescript-metadata": "^0.2.2",
+    "babel-polyfill": "^6.26.0",
```
vigneshshanmugam: should be `@babel/polyfill`

packages/rum-core/package.json
```diff
@@ -17,11 +17,13 @@
   "scripts": {
     "prepublishOnly": "npm run build:main",
     "build:main": "npx babel src -d dist/lib",
+    "karma:dev": "karma start --auto-watch --restartOnFileChange",
+    "karma:bench": "NODE_ENV=production karma start karma.bench.conf.js",
```
vigneshshanmugam: should be `bench` because of how benchmark script is run https://github.com/elastic/apm-agent-rum-js/blob/c38f175a3d9b8352c2d153318eb2ccf5c67e91c7/scripts/benchmarks.js#L90

packages/rum-core/src/common/apm-server.js
```diff
@@ -140,6 +140,29 @@ class ApmServer {
     return new Queue(onFlush, { queueLimit, flushInterval })
   }
 
+  fetchConfig(serviceName, environment) {
+    const serverUrl = this._configService.get('serverUrl')
+    var configEndpoint = `${serverUrl}/config/v1/agents`
+    if (!serviceName) {
```
vigneshshanmugam: We should not care about validation at this point since its already done when initialising the agent.

packages/rum-core/src/common/config-service.js
```diff
@@ -257,6 +258,21 @@ class Config {
       })
     }
 
+    const sampleRate = properties.transactionSampleRate
+    if (
+      typeof sampleRate !== 'undefined' &&
+      (typeof sampleRate !== 'number' ||
+        isNaN(sampleRate) ||
```
vigneshshanmugam: isNaN is not required since we already test for ranges and 'number' type. Am i missing something?

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -398,4 +401,28 @@ describe('ApmServer', function() {
     ]
     expect(result).toEqual(expected)
   })
+
+  if (isVersionInRange(testConfig.stackVersion, '7.3.0')) {
+    it('should fetch remote config', async () => {
+      var config = await apmServer.fetchConfig('nonexistent-service')
+      expect(config).toEqual({})
+      try {
```
vigneshshanmugam: nit: remove try catch.

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -398,4 +401,28 @@ describe('ApmServer', function() {
...
+      config = await apmServer.fetchConfig(
+        'nonexistent-service',
+        'nonexistent-env'
+      )
+      expect(config).toEqual({})
+
+      try {
+        config = await apmServer.fetchConfig()
+      } catch (e) {
+        expect(e).toBe('serviceName is required for fetching central config.')
```
vigneshshanmugam: should be removed here based on the validation logic.

packages/rum-core/src/common/config-service.js
vigneshshanmugam: I would simply replace this logic with `Number.isFinite(sampleRate)` check and checking the min and max range

packages/rum/src/apm-base.js
```diff
@@ -74,10 +74,46 @@ class ApmBase {
...
+    return apmServer
+      .fetchConfig(
+        configService.get('serviceName'),
+        configService.get('environment')
+      )
+      .then(config => {
+        var transactionSampleRate = config['transaction_sample_rate']
+        if (transactionSampleRate) {
+          transactionSampleRate = Number(transactionSampleRate)
+          const config = { transactionSampleRate }
```
vigneshshanmugam: this whole logic can be replaced with `this.config(config)` and the function will take care of the logging the error message and setting config if its valid.

vigneshshanmugam: Since the message here says value, we can replace this generic error logic with `value` instead of `character`. https://github.com/elastic/apm-agent-rum-js/blob/c38f175a3d9b8352c2d153318eb2ccf5c67e91c7/packages/rum/src/apm-base.js#L155

dev-utils/karma.js
jahtalab: This is needed for the async functions in the tests.

dev-utils/test-config.js
jahtalab: It's only because we might decide to add more condition before this check to expand this function.

dev-utils/test-servers.js
jahtalab: The serverUrl we get from the env variables is used for the tests to connect to, in other words this proxy is creating a server that tests will connect to. But the apm servers actual destination is a different one. We can add another env variable for that but at this point I think that's unnecessary.

packages/rum-core/src/common/apm-server.js
jahtalab: This function does not have any knowledge about validation process which happens elsewhere, but it needs to handle invalid arguments. In other words if for any reason we pass undefined as serviceName to this function it should be able to handle that instead of making an actual request to the apm server.

packages/rum/src/apm-base.js
jahtalab: `this.config` checks for `serviceName` and `serverUrl` which are required but we don't set them from the remote config.

packages/rum-core/test/common/apm-server.spec.js
jahtalab: Responded above.

jahtalab:
@vigneshshanmugam , 

Re. delaying the page load transaction for the remote config, I'm still on the fence about it, but I'm leaning toward it as well. 

Re. the caching there is a todo item in the original issue about it. It is not meant to be addressed in this PR.

packages/rum-core/src/common/config-service.js
jahtalab: `isNaN` is actually needed here because of the range check is always fails when the value is `NaN` and type its type is also a number.

packages/rum-core/src/common/apm-server.js
```diff
@@ -140,6 +147,29 @@ class ApmServer {
...
+    if (environment) {
+      configEndpoint += `&service.environment=${environment}`
+    }
+    return this._makeHttpRequest('GET', configEndpoint, { timeout: 5000 })
+      .then(response => {
+        return JSON.parse(response)
+      })
+      .catch(reason => {
+        const error = this._constructError(reason)
+        this._loggingService.warn('Failed fetching config!', error)
```
vigneshshanmugam: we should remove warning from here and only log it on `apm-base.js`. Right now it creates two warnings on the console.

packages/rum/src/apm-base.js
```diff
@@ -49,35 +49,84 @@ class ApmBase {
...
+            configService.get('sendPageLoadTransaction')
+          ) {
+            this._sendPageLoadMetrics()
+          }
+        }
+        if (configService.get('centralConfig')) {
+          /**
+           * Waiting for the remote config before sending the page load transaction
+           */
+          this.fetchCentralConfig().then(sendPageLoad)
```
vigneshshanmugam: We need tests for this case.

packages/rum-core/src/common/apm-server.js
```diff
@@ -140,6 +147,29 @@ class ApmServer {
...
+    if (!serviceName) {
+      return Promise.reject(
+        'serviceName is required for fetching central config.'
+      )
+    }
+    configEndpoint += `?service.name=${serviceName}`
+    if (environment) {
+      configEndpoint += `&service.environment=${environment}`
+    }
+    return this._makeHttpRequest('GET', configEndpoint, { timeout: 5000 })
```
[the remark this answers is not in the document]
jahtalab: The value doesn't seem to be one that we will reuse! Let's keep it here until we have more usage for value!

packages/rum/src/apm-base.js
jahtalab: Added the test to cover this!

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -340,4 +340,21 @@ describe('ApmBase', function() {
...
+
+  it('should wait for remote config before sending the page load', done => {
+    const apmBase = new ApmBase(serviceFactory, !enabled)
+    spyOn(apmBase, 'fetchCentralConfig').and.callThrough()
+    spyOn(apmBase, '_sendPageLoadMetrics').and.callFake(() => {
+      done()
+    })
+
+    apmBase.init({
+      serviceName: 'test-service',
```
vigneshshanmugam: nit: we can use the serviceName defined at the top.

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -330,15 +331,18 @@ describe('ApmBase', function() {
...
+    apmBase.fetchCentralConfig().then(() => {
+      expect(loggingService.warn).toHaveBeenCalledWith(
+        'Invalid value "NaN" for transactionSampleRate. Allowed: Number between 0 and 1.'
+      )
+      expect(configService.get('transactionSampleRate')).toBe(1)
+
+      apmServer._makeHttpRequest = createPayloadCallback('0.5')
+      apmBase.fetchCentralConfig().then(() => {
+        expect(configService.get('transactionSampleRate')).toBe(0.5)
+        done()
```
vigneshshanmugam: this test needs to account for failures. 
`done` should be handled for both success and failure cases.

jahtalab: Do you mean if the promise fails? I'm returning a Promise.resolve in the lines above, so not sure how the promise could fail!

Issue #416, opened 2019-09-12
Uncaught TypeError: Cannot convert undefined or null to object


adolt:
Hi! I found that @elastic/apm-rum will throw an error when captured a `Script Error`.

![image](https://user-images.githubusercontent.com/25565814/64771853-41540180-d582-11e9-9e40-ad2a8a9ac461.png)

I made a [issue example](https://github.com/adolt/apm-issue-example2). And I've found, it is because of missing the `null` type check [here](https://github.com/elastic/apm-agent-rum-js/blob/4c6d120afa49c77dcdb3fa0da5719fe55be270f9/packages/rum-core/src/error-logging/error-logging.js#L66) leads to `Object.keys(null)` [this line](https://github.com/elastic/apm-agent-rum-js/blob/4c6d120afa49c77dcdb3fa0da5719fe55be270f9/packages/rum-core/src/error-logging/error-logging.js#L129).

PR #418, opened 2019-09-12
fix(rum-core): handle script error events properly

vigneshshanmugam:
+ fixes https://github.com/elastic/apm-agent-rum-js/issues/416
+ When script error is thrown, error key is set to `null` which passes the typeof `object` check. We handle this case in this PR and the code portions are merged to make it cleaner.

### DIFF
--- a/packages/rum-core/src/error-logging/error-logging.js
+++ b/packages/rum-core/src/error-logging/error-logging.js
@@ -48,24 +48,26 @@ class ErrorLogging {
       culprit = lastFrame.filename
     }
 
-    const message =
-      errorEvent.message || (errorEvent.error && errorEvent.error.message)
-    let errorType = errorEvent.error ? errorEvent.error.name : ''
+    const { message, error } = errorEvent
+    let errorMessage = message
+    let errorType = ''
+    let errorContext = {}
+    if (error && typeof error === 'object') {
+      errorMessage = error.message
+      errorType = error.name
+      errorContext = this._getErrorProperties(error)
+    }
+
     if (!errorType) {
       /**
        * Try to extract type from message formatted like
        * 'ReferenceError: Can't find variable: initHighlighting'
        */
-      if (message && message.indexOf(':') > -1) {
-        errorType = message.split(':')[0]
+      if (errorMessage && errorMessage.indexOf(':') > -1) {
+        errorType = errorMessage.split(':')[0]
       }
     }
-
     const configContext = this._configService.get('context')
-    let errorContext
-    if (typeof errorEvent.error === 'object') {
-      errorContext = this._getErrorProperties(errorEvent.error)
-    }
     const browserMetadata = getPageMetadata()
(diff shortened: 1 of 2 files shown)


packages/rum-core/test/error-logging/error-logging.spec.js
```diff
@@ -220,7 +220,8 @@ describe('ErrorLogging', function() {
       colno
     })
     listener('error', {
-      message: 'Script error.' + testErrorMessage
+      message: 'Script error.' + testErrorMessage,
+      error: null
```
jahtalab: Shouldn't we also test for undefined as well as null?

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -48,32 +48,34 @@ class ErrorLogging {
       culprit = lastFrame.filename
     }
 
-    const message =
-      errorEvent.message || (errorEvent.error && errorEvent.error.message)
```
jahtalab: The logic here prioritises `errorEvent.message` over `errorEvent.error.message` if they both exists! The change reverses this priority, is that intentional? What's the reasoning?

packages/rum-core/test/error-logging/error-logging.spec.js
vigneshshanmugam: Its always set to null during the creation https://www.w3.org/TR/html50/webappapis.html#dom-errorevent-error

packages/rum-core/src/error-logging/error-logging.js
vigneshshanmugam: I didn't get you here. 

https://github.com/elastic/apm-agent-rum-js/blob/cd916b88395b879b6b472b8733d63fbe2938ef90/packages/rum-core/src/error-logging/error-logging.js#L52-L59

We set it to `errorEvent.message` first at line 52 and set it to `errorEvent.error.message` later on line 56.

packages/rum-core/test/error-logging/error-logging.spec.js
jahtalab: I vaguely remember that some browsers have/had a different behaviour even though the spec says it needs to be null!

packages/rum-core/src/error-logging/error-logging.js
jahtalab: Consider this:

```
let errorEvent = {message:'message', error: {message:'error-message'}}
```

The previous logic will result in `errorMessage` to be `message` since `errorEvent.message || (errorEvent.error && errorEvent.error.message)` only checks the first part of the OR operator and never checks the `error`. While the current changes will set `errorMessage` to `error-message`

vigneshshanmugam: Got it now, It was intentional that my impression that both messages are similar, I was wrong here. I will update and also add test case for the same.

Issue #410, opened 2019-09-06
react state update warning on unmounted component in ApmRoute component


vigneshshanmugam:
While instrumenting kibana, got this error warning in the dev mode 

```
backend.js:1 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in ChoroplethMap (created by PageLoadCharts)
    in div (created by EuiPanel)
    in EuiPanel (created by PageLoadCharts)
    in div (created by EuiFlexItem)
    in EuiFlexItem (created by PageLoadCharts)
    in div (created by EuiFlexGrid)
    in EuiFlexGrid (created by PageLoadCharts)
    in PageLoadCharts (created by TransactionCharts)
    in TransactionCharts (created by TransactionOverview)
    in ChartsSyncContextProvider (created by TransactionOverview)
    in div (created by EuiFlexItem)
    in EuiFlexItem (created by TransactionOverview)
    in div (created by EuiFlexGroup)
    in EuiFlexGroup (created by TransactionOverview)
    in TransactionOverview (created by ServiceDetailTabs)
    in ServiceDetailTabs (created by ServiceDetails)
    in div (created by ServiceDetails)
    in ServiceDetails (created by component)
    in component (created by withTransaction(component))
    in withTransaction(component) (created by Route)
    in Route (created by ApmRoute)
    in ApmRoute (created by App)

```

vigneshshanmugam:
useEffect should run only on mount of the component not on every render, this can be fixed by passing empty array as second argument to useEffect.

PR #419, opened 2019-09-12
fix(rum-core): create transaction only on component mount

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#410 
+ We need to create route change transaction only when component mounts(first render in terms of effect terminology), however the effect hook was running on component rerenders. By passing empty arr on the useEffect hook, we create the transaction only on mount

+ We also need to create transaction only once inside setState and not inside function body since it gets executed on every render and effectively changes the transaction state which was passed down to the underlying component. 

This PR keeps the functional components in sync with the Class based ones. 

## TODO
+ [x] Add tests
+ [x] figure out issues with transaction type not set to `route-change` in e2e. 
+ [x] Validate the issue with Kibana App

### DIFF
--- a/packages/rum-react/src/get-with-transaction.js
+++ b/packages/rum-react/src/get-with-transaction.js
@@ -79,15 +79,14 @@ function getWithTransaction(apm) {
         typeof React.useEffect === 'function'
       ) {
         ApmComponent = function ApmComponent(props) {
-          let transaction = null
           React.useEffect(() => {
-            transaction = apm.startTransaction(name, type, {
+            const transaction = apm.startTransaction(name, type, {
               canReuse: true
             })
             return () => transaction && transaction.detectFinish()
           }, [])
 
-          return <Component transaction={transaction} {...props} />
+          return <Component {...props} />
         }
       } else {
         ApmComponent = class ApmComponent extends React.Component {
--- a/packages/rum-react/src/get-with-transaction.js
+++ b/packages/rum-react/src/get-with-transaction.js
@@ -95,7 +95,7 @@ function getWithTransaction(apm) {
           )
 
           /**
-           * React guarentess the parent component effects are run after the child components effects
+           * React guarantees the parent component effects are run after the child components effects
            * So once all the child components effects are run, we run the detectFinish logic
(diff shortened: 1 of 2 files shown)


packages/rum-react/src/get-with-transaction.js
```diff
@@ -76,17 +76,39 @@ function getWithTransaction(apm) {
...
           React.useEffect(() => {
             transaction && transaction.detectFinish()
-            return () => transaction && transaction.detectFinish()
-          })
+            return () => {
+              /**
+               * Incase the transaction is never ended, we close the
+               * transaction when component unmounts
+               */
+              transaction && transaction.detectFinish()
```
jahtalab: should this be `transaction.end()`? since even if the transaction has scheduled tasks while the component unmount, it doesn't make sense to keep it open. For the class components we are also using `detectFinish` so, not strictly related to this PR.

packages/rum-react/src/get-with-transaction.js
```diff
@@ -76,17 +76,39 @@ function getWithTransaction(apm) {
...
+           * start the transaction only on component mounting
+           */
+          const [transaction] = React.useState(() =>
+            apm.startTransaction(name, type, {
+              canReuse: true
+            })
+          )
 
+          /**
+           * React guarentess the parent component effects are run after the child components effects
```
jahtalab: nit: "guarantees"

packages/rum-react/test/specs/get-with-transaction.spec.js
```diff
@@ -113,4 +112,45 @@ describe('withTransaction', function() {
     expect(WrappedComponent).toEqual(Component)
     expect(transactionService.startTransaction).not.toHaveBeenCalled()
   })
+
+  it('should not create new transaction on every render', () => {
```
jahtalab: Nice work with the tests 👍

packages/rum-react/src/get-with-transaction.js
vigneshshanmugam: Yeah you are  right, will change it to `.end` in both places.

Issue #334, opened 2019-07-17
Separate custom transaction life-cycle from automatic transaction


jahtalab:
Currently, transactions that are created by the user (custom transactions) and transactions that are created by the agent, live in the same space and can interfere with each other. This creates a number of issues:

* There can only be one active transaction at any give time
* Some transactions might be ended unexpectedly since another transaction is started.

To resolve these issues, it helps to think of the transactions in two different modes. One is automatic mode which means the agent controls the life-cycle of the transaction as well as creation of it's spans and the other one is manual mode in which case the agent does not create spans nor it detects the end of the transaction. we can add a property to the Transaction class that controls the mode and by default makes the transaction into manual mode.

With this approach we can address the following use-cases:
* Creating multiple transactions at the same time
* Capturing http requests as Transactions

The drawback of this approach is that custom transactions will no longer have automatic span creation (e.g. for XHR and fetch), to solve this we can give users the option to make a transaction in automatic mode.

WDYT, @vigneshshanmugam , @axw

axw:
Sounds good to me.

Is there anything stopping us also using/implementing something along the lines of https://github.com/angular/zone.js/? With something like that we could have an active transaction per zone, with `apm.startSpan` operating on the current zone's active transaction.

vigneshshanmugam:
Thanks Hamid for taking a stab at this. Seems super important and providing two modes at the same time seems like a good approach. 

> The drawback of this approach is that custom transactions will no longer have automatic span creation (e.g. for XHR and fetch), to solve this we can give users the option to make a transaction in automatic mode.

I feel like instead of doing this, We can provide events/some other mechanism to the users that would provide details of current span/marks that happen during the custom transaction. But these events only happen when we are not instrumenting the application using some other transaction(which is in automatic mode).

jahtalab:
@axw , early versions of the RUM agent used zone.js, but we decided to remove it, the main reason was the bundle size (zone.js is not very light). However, we can probably implement a simpler variation of zones in the RUM agent, although, zone management is not trivial to implement, so I would like to defer it until we have enough use-cases to justify it.

@vigneshshanmugam , I thought about having individual events for our instrumentation, but I think that is more complicated for the users. IMO, having modes makes it easy for the user to choose at a high level, it's also an easier solution to implement.

vigneshshanmugam:
[...]

PR #423, opened 2019-09-17
feat(rum-core): capture resource and user timing spans for soft navigation

### LINKED ISSUE #268: Capture resource and user timing spans for route change transactions
We should filter the resource entries to only include the entries that have are relevant for the route change transactions.

- [ ] Disable the resource timings on transaction class by default (#434)

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#268 
+ Use transaction start time as a filtering logic for capturing resource and user timing spans inside transaction timeframe. 

Should be merged after elastic/apm-agent-rum-js#424 

## TODO
+ [x] Update Unit tests
+ [x] Add E2E tests

### DIFF
--- a/packages/rum-core/src/performance-monitoring/capture-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-navigation.js
@@ -25,6 +25,7 @@
 
 import Span from './span'
 import {
+  PAGE_LOAD,
   RESOURCE_INITIATOR_TYPES,
   MAX_SPAN_DURATION,
   USER_TIMING_THRESHOLD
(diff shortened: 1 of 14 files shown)


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -94,16 +94,10 @@ class TransactionService {
   }
 
   capturePageLoadMetrics(tr) {
-    var capturePageLoad = this._config.get('capturePageLoad')
-    if (
-      capturePageLoad &&
-      !this._alreadyCapturedPageLoad &&
```
vigneshshanmugam: @jahtalab Do you remember why we need this flag?

packages/rum-core/src/performance-monitoring/capture-navigation.js
```diff
@@ -230,70 +232,89 @@ function createUserTimingSpans(entries, transactionEnd) {
...
+    const trStart = transaction._start
     /**
-     * capture resource timing information as Spans during page load transaction
+     * Capture resource timing information as spans
      */
-    if (typeof perf.getEntriesByType === 'function') {
-      const resourceEntries = perf.getEntriesByType('resource')
-
-      const ajaxUrls = []
-      for (let i = 0; i < transaction.spans; i++) {
```
vigneshshanmugam: this logic was never run since we didn't check for `.length` 🤦‍♂

jahtalab:
Decided to use the suggested flag `captureTimings` and remove `isHardNavigation`!

packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
```diff
@@ -194,28 +208,101 @@ describe('Capture hard navigation', function() {
...
+        ended: true
       })
     ])
   })
 
-  it('should captureHardNavigation', function() {
-    var tr = new Transaction('test', 'test')
-    tr.isHardNavigation = true
+  it('should capture spans for hard navigation', function() {
+    const tr = new Transaction('test', PAGE_LOAD)
```
vigneshshanmugam: for hard navigation we still need both `PAGE_LOAD` and `captureTimings` since we do not want to capture navigation timings as part of custom/route-change transaction.

vigneshshanmugam:
@jahtalab I will rebase it after the review so that I can address it along with review changes if required.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -262,7 +262,7 @@ class PerformanceMonitoring {
       'checkBrowserResponsiveness'
     )
 
-    if (checkBrowserResponsiveness && !tr.isHardNavigation) {
+    if (checkBrowserResponsiveness && !tr.captureTimings) {
```
jahtalab: Would you please add a todo here to refactor `captureTimings` here once the #334 issue is fixed.

PR #427, opened 2019-09-18
feat(rum-core): capture unhandled promise rejection as errors

### LINKED ISSUE #280: capture unhandledRejection as errors in the agent
+ Promise rejection errors bubble up to the window on `PromiseRejectionEvent`, We can hook on to those events and capture them as error events. More visibility in to the application. 
+ details - https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#280 
+ We use the prefix  'Unhandled promise rejection: ' when the reason is not an error object, since we want to capture the exception type as `ReferenceError` or something similar when error object is thrown as part of promise rejection event.


packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -110,12 +110,31 @@ class ErrorLogging {
...
+  }
+
+  logPromiseEvent(promiseRejectionEvent) {
+    const prefix = 'Unhandled promise rejection: '
+    const { reason } = promiseRejectionEvent
+
+    if (reason == null) {
+      this.logError(prefix + '<no reason defined>')
+    } else if (typeof reason === 'object') {
+      this.logError(reason)
```
jahtalab: `reason` could be any type of object, but logError only accepts string or Error .

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -110,12 +110,31 @@ class ErrorLogging {
...
+    window.addEventListener('unhandledrejection', promiseRejectionEvent =>
+      this.logPromiseEvent(promiseRejectionEvent)
+    )
+  }
+
+  logPromiseEvent(promiseRejectionEvent) {
+    const prefix = 'Unhandled promise rejection: '
+    const { reason } = promiseRejectionEvent
+
+    if (reason == null) {
```
jahtalab: Do we need to check for undefined?

packages/rum-core/test/error-logging/error-logging.spec.js
```diff
@@ -285,4 +286,52 @@ describe('ErrorLogging', function() {
...
+      done()
+    })
+    /**
+     * simulate window.PromiseRejectionEvent event since its not supported by
+     * all browsers
+     */
+    const reason = new Error(testErrorMessage)
+    const event = new CustomEvent('unhandledrejection')
+    event.reason = reason
+    window.dispatchEvent(event)
```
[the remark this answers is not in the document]
vigneshshanmugam: it's not actually different CustomEvent extends Event. This makes it simple.

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -110,12 +110,31 @@ class ErrorLogging {
...
     window.addEventListener('error', errorEvent =>
       this.logErrorEvent(errorEvent)
     )
   }
 
+  onUnhandledRejection() {
+    window.addEventListener('unhandledrejection', promiseRejectionEvent =>
+      this.logPromiseEvent(promiseRejectionEvent)
+    )
+  }
```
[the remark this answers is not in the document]
vigneshshanmugam: Yeah lets keep the registerErrorListener and both methods under these.

packages/rum-core/src/error-logging/error-logging.js
vigneshshanmugam: The only interesting things that we need from the object is `message` and `stack`. logError handles that case. I will make it clear here.

packages/rum-core/src/error-logging/error-logging.js
vigneshshanmugam: `==` checks for both `null` and `undefined`.

packages/rum-core/src/error-logging/error-logging.js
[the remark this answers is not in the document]
vigneshshanmugam: done

packages/rum-core/src/error-logging/error-logging.js
jahtalab: What about other falsy values?

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -110,10 +110,32 @@ class ErrorLogging {
...
+
+  logPromiseEvent(promiseRejectionEvent) {
+    const prefix = 'Unhandled promise rejection: '
+    const { reason } = promiseRejectionEvent
+
+    if (reason == null) {
+      this.logError(prefix + '<no reason specified>')
```
jahtalab: Do other agents have a similar message here? Maybe it makes sense to use the same format for `no reason specified`

packages/rum-core/test/error-logging/error-logging.spec.js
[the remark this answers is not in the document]
jahtalab: `PromiseRejectionEvent` and `CustomEvent`  both extend Event but that doesn't mean they are the same. Even though `PromiseRejectionEvent` might be supported by a browser they might not support it completely, or the implementation might differ. IMO, it's always better to test the actual interfaces if possible.

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -110,10 +110,32 @@ class ErrorLogging {
...
+    } else if (typeof reason.message === 'string') {
+      /**
+       * Promise is rejected with an error or error like object
+       */
+      this.logError({
+        message: prefix + reason.message,
+        stack: reason.stack ? reason.stack : null
+      })
+    } else {
+      this.logError(prefix + reason)
```
jahtalab: The reason can be an object as well in that case the error message would include `[object Object]`! We should handle the different types of `reason` properly!

packages/rum-core/test/error-logging/error-logging.spec.js
```diff
@@ -285,4 +286,61 @@ describe('ErrorLogging', function() {
...
+
+    const errorLikeObj = {
+      message: testErrorMessage,
+      foo: 'bar'
+    }
+    errorLogging.logPromiseEvent({
+      reason: errorLikeObj
+    })
+    expect(getErrors()[4].exception.message).toMatch(testErrorMessage)
+    expect(getErrors()[4].exception.stacktrace.length).toBe(0)
```
jahtalab: We should expand this test to include `reason` that are arbitrary objects or even other types! Because `reason` can be created by user code and it means it can have any value!

packages/rum-core/test/error-logging/error-logging.spec.js
```diff
@@ -285,4 +286,61 @@ describe('ErrorLogging', function() {
...
+     * simulate window.PromiseRejectionEvent event since its not supported by
+     * all browsers
+     */
+    const reason = new Error(testErrorMessage)
+    const event = new CustomEvent('unhandledrejection')
+    event.reason = reason
+    window.dispatchEvent(event)
+  })
+
+  it('should handle multiple reasons for promise rejections', () => {
```
jahtalab: Maybe something like `should handle different type of reason for promise rejection`

packages/rum-core/src/error-logging/error-logging.js
vigneshshanmugam: Checked Node.js already, there is no message. we can agree on it later if it needed.

packages/rum-core/src/error-logging/error-logging.js
vigneshshanmugam: good point, will move the logic around to make it easier.

packages/rum-core/test/utils/polyfill.js
```diff
@@ -93,3 +93,21 @@ Object.setPrototypeOf =
...
+  if (typeof window.CustomEvent === 'function') return false
+
+  function CustomEvent(event, params) {
+    params = params || { bubbles: false, cancelable: false, detail: null }
+    var evt = document.createEvent('CustomEvent')
+    evt.initCustomEvent(event, params.bubbles, params.cancelable, params.detail)
+    return evt
+  }
+
+  window.CustomEvent = CustomEvent
```
jahtalab: Can we move this to the file level instead of setting on window? The reason being what we already discussed in other PRs in regards to avoiding global polyfills to make sure our tests run in a close environment to the real browsers!

vigneshshanmugam: We wont run in to issues like those since this one is pretty specific to tests and we don't use this code on our src files.

jahtalab: We don't currently `CustomEvent` in our source but we might in the future. I think it's a good idea to keep the polyfills to a minimum and limited to the file level as much as possible!

vigneshshanmugam: moved it.

vigneshshanmugam: moved it inside file.

adolt:
Since `unhandledrejection` is not well supported, what about the unsupported browsers?

vigneshshanmugam:
Agent also listens for `onerror` events, so errors would be captured through those listeners.

PR #432, opened 2019-09-19
fix(rum): publish all packages as transpiled modules

vigneshshanmugam:
+ Publish all the packages/* transpiled by `babel-preset-env` with the least supported environment which is **Internal Explorer 11**
+ We have three entry points defined in the package.json 
```
'main': 'dist/lib/index.js' // output path for CommonJS/Node
'module':  'dist/es/index.js'  // output path for JS Modules with `import` & `export` statements
'source': 'src/index.js'  // custom entry module which might be used by supported bundlers. 
```

+ Most of the bundlers prefer `module` over `main` since it can do better tree-shaking on ES6 modules which would reduce the code size. 

+ This would allow users to not configure babel for `@elastic/*` modules which might be painful since users might need to be aware of the features/proposals that are used by the RUM library. 

## TODO
+ [x] Add tests to check if we are publishing es5 code.


vigneshshanmugam:
Integration test failure is related to https://github.com/elastic/apm-integration-testing/pull/600

packages/rum/test/e2e/index.js
```diff
@@ -24,7 +24,7 @@
  */
 
 import { polyfill } from 'es6-promise'
-import { apmBase } from '../../src'
+import { apmBase } from '@elastic/apm-rum'
```
vigneshshanmugam: The reason why this has to be changed is the same file has been used to create apm instance inside `rum-react` package https://github.com/elastic/apm-agent-rum-js/blob/28df070ed4001d1e3ee3fcb929364e5960339793/packages/rum-react/test/e2e/index.js#L26

Due to our new bundling one reference would go to `packages/rum/src/index.js` and other would go to `packages/rum/dist/es/index.js` and both would result in two apm instances which would make the test fail. By referencing it with package name we solve this issue since webpack always resolves to `module` field in package.json.

packages/rum/test/bundle/bundle.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      libraryTarget: 'umd'
+    },
+    mode: 'none',
+    plugins: [new UglifyJSPlugin({})]
+  }
+}
+
+describe('Browser bundle test', () => {
+  afterEach(() => {
+    rimraf.sync(DIST_DIR)
```
jahtalab: Instead of this can we change the output directory of the bundle to `tmp/bundle-test`? and also not pollute the dist folder with test temporary files!

vigneshshanmugam: this dist folder path is inside `test/bundle/dist`.

jahtalab: Right, Maybe using `tmp` folder would be a better option still. Since we gitignore that folder, I know this is removing the folder after each test but for example if the process unexpectedly exits!

vigneshshanmugam: Sure will change the path name

PR #433, opened 2019-09-19
refactor: use the npm run working directory for running tests

vigneshshanmugam:
+ small cleanup to remove passing around package paths to all the script commands.


jahtalab:
We decided to investigate the parameter parsing, something like `run script --path=./test`.

PR #438, opened 2019-09-25
feat(rum-core): move breakdown algorithm post transaction finish

vigneshshanmugam:
+ Introduce setting `start` and `end` of transaction and spans through options arguments
+ Also address these comments  https://github.com/elastic/apm-agent-rum-js/issues/289#issuecomment-534999842

TODO

+ [x] The algorithm needs few changes and tests fails currently at the moment.

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -239,7 +239,7 @@ class ApmServer {
   }
 
   ndjsonMetricsets(metricsets) {
-    const timestamp = Date.now()
+    const timestamp = Date.now() * 1000
     return metricsets
       .map(metricset =>
         NDJSON.stringify({ metricset: { timestamp, ...metricset } })
--- a/packages/rum-core/src/performance-monitoring/breakdown.js
+++ b/packages/rum-core/src/performance-monitoring/breakdown.js
@@ -41,12 +41,59 @@ function getValue(value) {
   return { value }
 }
 
+function calculateSelfTime(transaction) {
+  const { spans, _start, _end } = transaction
+  /**
+   * When there are no spans transaction duration accounts for
+   * the overall self time
+   */
+  if (spans.length === 0) {
+    return transaction.duration()
+  }
+  /**
+   * The below selfTime calculation logic assumes the spans are sorted
+   * based on the start time
+   */
+  spans.sort((span1, span2) => span1._start - span2._start)
+
+  let span = spans[0]
+  let spanEnd = span._end
+  let spanStart = span._start
+
+  let lastContinuousEnd = spanEnd
+  let selfTime = spanStart - _start
+
+  for (let i = 1; i < spans.length; i++) {
+    span = spans[i]
+    spanStart = span._start
+    spanEnd = span._end
+
+    /**
+     * Add the gaps between the spans to the self time
+     */
+    if (spanStart > lastContinuousEnd) {
+      selfTime += spanStart - lastContinuousEnd
+      lastContinuousEnd = spanEnd
+    } else if (spanEnd > lastContinuousEnd) {
+      lastContinuousEnd = spanEnd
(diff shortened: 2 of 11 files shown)


packages/rum-core/src/common/apm-server.js
```diff
@@ -239,7 +239,7 @@ class ApmServer {
   }
 
   ndjsonMetricsets(metricsets) {
-    const timestamp = Date.now()
+    const timestamp = Date.now() * 1000
```
vigneshshanmugam: this was missed when changing the implementation from `microtime`, has to be in microseconds for the UI to display the breakdown.

vigneshshanmugam:
@jahtalab Thanks Hamid, I have modified the algorithm a bit.

packages/rum-core/src/performance-monitoring/span-base.js
```diff
@@ -50,7 +50,7 @@ class SpanBase {
     this.traceId = options.traceId
     this.sampled = options.sampled
     this.timestamp = options.timestamp
-    this._start = now()
+    this._start = mayBeTime(options.startTime)
```
jahtalab: Not sure if we need this function here. Why not just a simple falsy check on `options.startTime`.

Also, maybe just use `options.start` instead!

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -41,12 +41,59 @@ function getValue(value) {
...
+   * based on the start time
+   */
+  spans.sort((span1, span2) => span1._start - span2._start)
+
+  let span = spans[0]
+  let spanEnd = span._end
+  let spanStart = span._start
+
+  let lastContinuousEnd = spanEnd
+  let selfTime = spanStart - _start
```
jahtalab: Could this be negative? We moved the logic for adjusting the transaction times to the earliest and latest span to `PerformanceMonitoring.adjustTransactionTime` but at that point we have already calculated breakdowns.

Maybe we should refactor `adjustTransactionTime` to run after `Transaction.end` again.
We also need test for this case

packages/rum-core/src/performance-monitoring/span-base.js
vigneshshanmugam: have to account for 0 as well and was using it on lots of places so moved it to function.

packages/rum-core/src/performance-monitoring/breakdown.js
vigneshshanmugam: Done

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -217,7 +209,24 @@ class TransactionService {
...
+         * Capture breakdown metrics once the transaction is completed
+         */
+        const breakdownMetrics = this._config.get('breakdownMetrics')
+        if (breakdownMetrics) {
+          tr.captureBreakdown()
+        }
+
+        if (__DEV__) {
+          this._logger.debug('TransactionService transaction ended', tr)
+        }
```
jahtalab: Can we move this debug log to the end?

packages/rum-core/src/performance-monitoring/span-base.js
jahtalab: I think something like `isValidTime` is more clear about what is does and can be used elsewhere as well. `mayBeTime` is really ambiguous about what it actually does!

Also I'm questioning whether we should do any validation here. The reason is that if the value of `startTime` is not valid (e.g. negative value) currently we just use `now()` instead. This might prevent us from capturing bugs or could make debugging issues harder because we use `now()` when `startTime` is invalid.

vigneshshanmugam: decided to rename the function to `getTime`

PR #439, opened 2019-09-26
feat(rum-core): use etag for fetching config

jahtalab:
Part of elastic/apm-agent-rum-js#253
store remote config in localStorage

### DIFF
--- a/dev-utils/test-servers.js
+++ b/dev-utils/test-servers.js
@@ -80,8 +80,16 @@ function startApmServerProxy(port = 8001) {
       userResHeaderDecorator(headers) {
         if (!headers['Access-Control-Allow-Origin']) {
           headers['Access-Control-Allow-Origin'] = '*'
+          headers['Access-Control-Expose-Headers'] = 'ETag'
         }
         return headers
+      },
+      proxyReqOptDecorator(proxyReqOpts, srcReq) {
+        const { ifnonematch } = srcReq.query
+        if (ifnonematch) {
+          proxyReqOpts.headers['If-None-Match'] = `"${ifnonematch}"`
+        }
+        return proxyReqOpts
       }
     })
   )
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -80,11 +80,14 @@ class ApmServer {
       headers: {
         'Content-Type': 'application/x-ndjson'
       }
-    })
+    }).then(({ responseText }) => responseText)
   }
 
   _constructError(reason) {
     const { url, status, responseText } = reason
+    if (typeof status == 'undefined') {
+      return reason
+    }
     let message = url + ' HTTP status: ' + status
     if (__DEV__ && responseText) {
       try {
@@ -127,7 +130,7 @@ class ApmServer {
           if (status === 0 || (status > 399 && status < 600)) {
             reject({ url, status, responseText })
           } else {
-            resolve(responseText)
+            resolve(xhr)
           }
         }
       }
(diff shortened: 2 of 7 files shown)


dev-utils/test-servers.js
```diff
@@ -80,8 +80,16 @@ function startApmServerProxy(port = 8001) {
...
         if (!headers['Access-Control-Allow-Origin']) {
           headers['Access-Control-Allow-Origin'] = '*'
+          headers['Access-Control-Expose-Headers'] = 'ETag'
         }
         return headers
+      },
+      proxyReqOptDecorator(proxyReqOpts, srcReq) {
+        const { ifnonematch } = srcReq.query
+        if (ifnonematch) {
+          proxyReqOpts.headers['If-None-Match'] = `"${ifnonematch}"`
```
vigneshshanmugam: doesn't it create `""value""` value instead of "value"

vigneshshanmugam: doesn't it create `""value""` value instead of "value"

packages/rum-core/src/common/apm-server.js
```diff
@@ -159,9 +162,26 @@ class ApmServer {
...
+    if (localConfig) {
+      configEndpoint += `&ifnonematch=${localConfig.etag}`
+    }
+
     return this._makeHttpRequest('GET', configEndpoint, { timeout: 5000 })
-      .then(response => {
-        return JSON.parse(response)
+      .then(xhr => {
+        const { status, responseText } = xhr
+        const etag = xhr.getResponseHeader('etag')
```
vigneshshanmugam: shouldn't we account for multiple values if the header is set more than 1 time? Its an edge case, but pretty common if APM server has Load balancer in front.

packages/rum-core/src/common/apm-server.js
```diff
@@ -81,11 +81,14 @@ class ApmServer {
       headers: {
         'Content-Type': 'application/x-ndjson'
       }
-    })
+    }).then(({ responseText }) => responseText)
   }
 
   _constructError(reason) {
     const { url, status, responseText } = reason
+    if (typeof status == 'undefined') {
```
vigneshshanmugam: can you explain a bit on why we need this special case?

dev-utils/test-servers.js
jahtalab: The `ifnonematch` query string is sent without double quotes!

packages/rum-core/src/common/apm-server.js
jahtalab: Can you please elaborate a bit on "Load balancer" scenario?

packages/rum-core/src/common/apm-server.js
jahtalab: This is when the` reason` is any thing but a XHR object, for example the `reason` could be an `Error` if there's a bug in our code that throws an exception. without this that exception is lost which makes it harder for us to debug.

packages/rum-core/src/common/apm-server.js
```diff
@@ -158,9 +164,26 @@ class ApmServer {
...
-        return JSON.parse(response)
+      .then(xhr => {
+        const { status, responseText } = xhr
+        if (status === 304) {
+          return localConfig
+        } else {
+          let remoteConfig = JSON.parse(responseText)
+          const etag = xhr.getResponseHeader('etag')
+          if (etag) {
+            remoteConfig.etag = etag.replace(/["]/g, '')
```
vigneshshanmugam: was there any discussion on storing the etag without double quotes in RUM? I see two contradicting comments here https://github.com/elastic/apm-server/pull/2747#discussion_r333453028

packages/rum-core/src/common/config-service.js
```diff
@@ -260,6 +260,21 @@ class Config {
...
+    if (config) {
+      return JSON.parse(config)
+    }
+  }
+
+  setLocalConfig(config) {
+    if (config) {
+      sessionStorage.setItem(LOCAL_CONFIG_KEY, JSON.stringify(config))
+    } else {
+      sessionStorage.removeItem(LOCAL_CONFIG_KEY)
```
vigneshshanmugam: just worried about whether this might hit us hard at somepoint. Will there be a case when server returns a falsy value and we end up deleting the old config value? If this is used only for tests, lets remove this.

packages/rum-core/src/common/apm-server.js
jalvz: i honestly would have preferred to keep the double quotes everywhere as-is to not need extra logic on both sides, but from previous discussion with @jahtalab , apm-server will expect the `ifnonematch` query parameter without double quotes (only when passed as a query param).

the Etag header returned by apm-server in the response will include double quotes.

packages/rum-core/src/common/config-service.js
jahtalab: Decided to ignore falsy values instead of removing the local config!

packages/rum-core/src/common/apm-server.js
jahtalab: The main reason was that double quotes need to be encoded in the URL (and decoded on the server).

PR #440, opened 2019-09-26
feat: Introduce managed transaction option

jahtalab:
Fixes: #334 
* By default startTransaction creates an unmanaged transaction
* Refactor TransactionService

### DIFF
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -94,6 +94,7 @@ class PerformanceMonitoring {
     return (event, task) => {
       if (task.source === HISTORY && event === INVOKE) {
         transactionService.startTransaction(task.data.title, 'route-change', {
+          managed: true,
           canReuse: true
         })
       }
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -94,22 +94,24 @@ class TransactionService {
 
   createPerfOptions(options) {
     const config = this._config.config
-    return extend(
-      {
-        pageLoadTraceId: config.pageLoadTraceId,
-        pageLoadSampled: config.pageLoadSampled,
-        pageLoadSpanId: config.pageLoadSpanId,
-        pageLoadTransactionName: config.pageLoadTransactionName,
-        transactionSampleRate: config.transactionSampleRate,
-        checkBrowserResponsiveness: config.checkBrowserResponsiveness
-      },
-      options
-    )
+    let presetOptions = { transactionSampleRate: config.transactionSampleRate }
+    let perfOptions = extend(presetOptions, options)
+    if (perfOptions.managed) {
+      perfOptions = extend(
+        {
+          pageLoadTraceId: config.pageLoadTraceId,
+          pageLoadSampled: config.pageLoadSampled,
+          pageLoadSpanId: config.pageLoadSpanId,
+          pageLoadTransactionName: config.pageLoadTransactionName,
+          checkBrowserResponsiveness: config.checkBrowserResponsiveness
+        },
+        perfOptions
+      )
+    }
+    return perfOptions
   }
 
-  startTransaction(name, type, options) {
-    const perfOptions = this.createPerfOptions(options)
-
+  startManagedTransaction(name, type, perfOptions) {
     let tr = this.getCurrentTransaction()
 
     if (!tr) {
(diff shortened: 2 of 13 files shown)


packages/rum-core/test/benchmarks/transaction-service.bench.js
```diff
@@ -35,12 +35,16 @@ suite('TransactionService', () => {
...
+    const tr = transactionService.startTransaction('/index', 'page-load', {
+      managed: true
+    })
     setImmediate(() => tr.end())
   })
 
   benchmark('custom transaction overhead', () => {
-    const tr = transactionService.startTransaction('custom', 'custom')
+    const tr = transactionService.startTransaction('custom', 'custom', {
+      managed: true
```
vigneshshanmugam: managed must be false here since we are measuring the overhead of custom transaction.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -163,51 +165,66 @@ class TransactionService {
       tr.captureTimings = true
```
vigneshshanmugam: we can set this flag without checking for route-change or page-load since its managed.

axw:
@jahtalab I like "managed" - nice one. I think we need to update both https://www.elastic.co/guide/en/apm/agent/rum-js/current/custom-transactions.html and https://www.elastic.co/guide/en/apm/agent/rum-js/current/transaction-api.html. The first one states "There can only be one active transaction at any time", so we'll need to update that, and make it clear that spans won't be automatically created for unmanaged transactions.

jahtalab:
@vigneshshanmugam , 

* I don't think the `managed` flag should be true by default. A managed transaction would reset the current transaction, it's better to explicitly set this flag to true once we set the current transaction instead of the other way around.
* At this point I see `canReuse` and `managed` separately, in other words there can be a managed transaction that should not be reused.
* I think we should keep the behaviour of `detectFinish` simple and instead the caller should check if the transaction is managed or not.
* Unmanaged transactions never exist on the `currentTransaction` therefore we never automatically add a span to them. The only way to add a span to an unmanaged transaction is by calling `startSpan` on the transaction object.


@axw , Good point regarding the docs I will add an item to the issue to address it in a separate PR.

jahtalab:
* Decided to refactor the logic in the multiple framework integrations into a single method. e.g. `startRouteChange` `startPageLoad` or `startNavigationTransaction`, so that this logic is not spread over multiple places. We will address this separately.
* We should check if `transactionService.detectFinish` is used anywhere, if not let's remove it, and we should update the tests to use `removeTask` instead. Also add a test for `transactionService.startSpan` for unmanaged transactions.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -262,11 +263,7 @@ class PerformanceMonitoring {
       'checkBrowserResponsiveness'
     )
 
-    /**
-     * TODO: Refactor captureTimings flag here once this PR
-     * is addressed - https://github.com/elastic/apm-agent-rum-js/issues/334
-     */
-    if (checkBrowserResponsiveness && !tr.captureTimings) {
+    if (checkBrowserResponsiveness && tr.options.managed) {
```
vigneshshanmugam: Our previous logic was doing this for all transaction except for page-load since we never want to miss sending the page-load. But right now if the browser is not responsive, we still dont send page-load. Can we remove this check?

vigneshshanmugam: Our previous logic was doing this for all transaction except for page-load since we never want to miss sending the page-load. But right now if the browser is not responsive, we still dont send page-load. Can we remove this check?

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -94,22 +95,24 @@ class TransactionService {
 
   createPerfOptions(options) {
```
vigneshshanmugam: nit: probably change it to createOptions? `createPerfOptions` doesn't seem right. Nothing related to the PR, just saying since you are already doing the refactoring.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -358,4 +377,43 @@ describe('TransactionService', function() {
...
+    spyOn(tr1, 'addTask').and.callThrough()
+    spyOn(tr1, 'removeTask').and.callThrough()
+    spyOn(tr1, 'end')
+    transactionService.addTask('taskId')
+    expect(tr1.addTask).toHaveBeenCalledWith('taskId')
+
+    transactionService.removeTask('taskId')
+    expect(tr1.removeTask).toHaveBeenCalledWith('taskId')
+    expect(tr1.end).toHaveBeenCalled()
+  })
```
vigneshshanmugam: Can we add another test where we add two transactions 1. managed and 2. unmanaged and call start span accordingly and check if they are getting added to the right thing. It provides some guarantees that both the transactions can coexist on the page.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -149,6 +149,7 @@ class TransactionService {
     tr.captureTimings = true
 
     if (type === PAGE_LOAD) {
+      tr.options.checkBrowserResponsiveness = false
```
vigneshshanmugam: I like this approach 👍

vigneshshanmugam:
@jahtalab Are we targeting docs changes in this PR that andrew mentioned?

jahtalab:
Just created the docs PR: #442

PR #446, opened 2019-10-02
fix(rum-core): handle relative urls without slash properly

### LINKED ISSUE #444: [Bug] Wrong origin check for relative URLs
For API calls in our application we're using relative urls: 
```
fetch('api/packages/getAll', {...}).
```

For that calls 'elastic-apm-traceparent' header is missing.

After research, we found out that origin is not correct for those calls:

Screen Shot 2019-09-28 at 23 58 44

checkSameOrigin(...) returns false for such kind of urls (not absolute).

https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/src/performance-monitoring/performance-monitoring.js#L140

Now we're using a "hacky" tmp solution:

```
initApm({
    ...
    distributedTracingOrigins: ['https://api'],
     ...
});
```

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#444 
+ Handle the pathname resolution according to browser URL implementation. 
+ Added more tests to cover this case.

### DIFF
--- a/packages/rum-core/test/common/url.spec.js
+++ b/packages/rum-core/test/common/url.spec.js
@@ -101,8 +101,45 @@ describe('Url parser', function() {
     var result = new Url('?param=value')
     expect(result).toEqual(
       jasmine.objectContaining({
-        path: '',
+        path: '/',
         query: '?param=value'
+      })
+    )
+  })
+
+  it('should parse url correctly without /', () => {
+    expect(new Url('api/foo')).toEqual(
+      jasmine.objectContaining({
+        path: '/api/foo',
+        query: '',
+        hash: '',
+        relative: true,
+        protocol: 'http:'
+      })
+    )
+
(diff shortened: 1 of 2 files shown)


packages/rum-core/test/common/url.spec.js
```diff
@@ -101,12 +101,49 @@ describe('Url parser', function() {
     var result = new Url('?param=value')
     expect(result).toEqual(
       jasmine.objectContaining({
-        path: '',
+        path: '/',
```
vigneshshanmugam: behaves according to browser URL implementation. 

```js
new URL('?param=value', 'http://localhost')
  host: "localhost"
  hostname: "localhost"
  href: "http://localhost/?param=value"
  origin: "http://localhost"
  pathname: "/"
  protocol: "http:"
  search: "?param=value"
```

PR #449, opened 2019-10-07
fix(rum-angular): use explicit angular injection for service

### LINKED ISSUE #437: 'apm-rum-angular' ApmService is not initialized with router and route in production build of Angular 7
I cloned https://stackblitz.com/edit/angular-router-basic-example-grz785 demo, upgraded angular version to 7 and apm-rum-angular to 0.1.1 to satisfy devironment environment at work.

Run ng build --prod to build project and serve the production build by http-server.  "Uncaught TypeError: Cannot read property 'events' of undefined" error message is logged in console log.

After further investigation,   undefined router and undefined route is passed to the constructor of ApmService. (https://github.com/elastic/apm-agent-rum-js/blob/28df070ed4001d1e3ee3fcb929364e5960339793/packages/rum-angular/src/apm-service.ts#L43)

this.router.events throws error since this.router is undefined.
(https://github.com/elastic/apm-agent-rum-js/blob/28df070ed4001d1e3ee3fcb929364e5960339793/packages/rum-angular/src/apm-service.ts#L65)

Test repo: https://github.com/railsstudent/apm-rum-angular-routing-failed

Please kindly help!

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#437 
+ Libraries needs to use explicit injection to make it work with all build configs.
+ Since the injection uses Reflection API like metadata and decorate, its better to use explicit check instead of doing `instanceof` since it fails with Angular CLI builds. 
+ remove the activated route deps since we can extract the firstChild from router state.  
+ Modify the E2E tests with the new behaviour. 
+ Update the angular docs to reflect the same. 

We can extend the tests in the future to include Angular AOT compiler and all those, but it requires pulling in lots of dependencies, so kept it simple for now and tested all the scenarios in the example repo which is a repo constructed with Angular CLI https://github.com/railsstudent/apm-rum-angular-routing-failed

### DIFF
--- a/docs/angular-integration.asciidoc
+++ b/docs/angular-integration.asciidoc
@@ -16,8 +16,8 @@ npm install @elastic/apm-rum-angular --save
 [float]
 ==== Instrumenting your Angular application
 
-The Angular integration package comes with a `ApmService` that will start subscribing to 
-https://angular.io/api/router/Event[Angular Router Events] once the service is initialized. 
+The Angular integration package comes with a `ApmService` which uses Angular Depedency injection pattern and 
+will start subscribing to https://angular.io/api/router/Event[Angular Router Events] once the service is initialized. 
 
 `ApmService` must be initialized from either the application module or application component since 
 the RUM agent has to start capturing all the resources and API calls as soon as possible. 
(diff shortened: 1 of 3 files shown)


axw:
@vigneshshanmugam 

description says

> remove the activated route deps since we can extract the firstChild from router state.

Looks like you reverted that. Does the description need updating, or the code?

vigneshshanmugam:
@axw I was testing e2e test, revered it again. good for review now.

PR #457, opened 2019-10-16
chore: upgrade dev dependencies and fix flaky test

vigneshshanmugam:
+ upgrades all the test/dev related dependencies only
+ fix a flaky test which relied on page load.


jahtalab:
We need more investigate for the test failures.

dev-utils/karma.js
```diff
@@ -56,6 +56,11 @@ const baseConfig = {
     'karma-webpack',
     'karma-sourcemap-loader'
   ],
+  client: {
+    jasmine: {
+      random: false
```
vigneshshanmugam: jasmine 3.0 has this breaking change of running tests in random order by default. Had to disable it to fix the XHR tests.

vigneshshanmugam:
Have to figure out why the tests are getting stuck on travis.

dev-utils/karma.js
```diff
@@ -158,7 +164,6 @@ function prepareConfig(config, packageName) {
     console.log('prepareConfig: Run in SauceLab mode')
     config.sauceLabs.build = buildId
     console.log('saucelabs.build:', buildId)
-    config.concurrency = 3
```
vigneshshanmugam: By setting it to `Infinity`, it starts the process and waits for browser socket connection to run the tests which seems to be work better without hanging to start the browsers after 3 tests passes. https://travis-ci.org/elastic/apm-agent-rum-js/jobs/622655609

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -34,41 +35,32 @@ var enabled = bootstrap()
...
-      })
+
+    apmBase.setInitialPageLoadName('new page load')
+    apmBase.observe(TRANSACTION_END, endedTr => {
+      expect(document.readyState).toBe('complete')
+      expect(tr.detectFinish).toHaveBeenCalled()
+      expect(endedTr.name).toBe('new page load')
+      expect(endedTr.type).toBe(PAGE_LOAD)
+      done()
     })
```
jahtalab: nit: can you use one of the `endedTr` or `tr`?
You can potentially check that they are actually the same objects `endedTr === tr`

dev-utils/karma.js
jahtalab: Doesn't it make the test more unstable in saucelabs since we only have a limited number of sessions there?

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -74,13 +74,9 @@ describe('ApmServer', function() {
   var apmServer
   var configService
   var loggingService
-  var originalTimeout
   var performanceMonitoring
 
   beforeEach(function() {
-    originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL
-    jasmine.DEFAULT_TIMEOUT_INTERVAL = 50000
```
jahtalab: I think these were here because we make actual connection to the APM server and that can take longer than the default timeout interval

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -605,6 +611,8 @@ describe('PerformanceMonitoring', function() {
               source: XMLHTTPREQUEST
             }
           ])
+          cancelXHRSub()
+          cancelFetchSub()
```
jahtalab: Thanks for cleaning up the subscriptions 👍

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -34,41 +35,32 @@ var enabled = bootstrap()
...
   })
 
   it('should disable all auto instrumentations when instrument is false', () => {
-    const apmBase = new ApmBase(serviceFactory, !enabled)
```
jahtalab: Not strictly related to this PR but I remember that we had a test `new ApmBase(serviceFactory, true)` that tests the agent when it's disabled in the environment. I can't find it now, do you remember why we removed that?

packages/rum/test/specs/index.spec.js
```diff
@@ -31,16 +31,6 @@ import { getGlobalConfig } from '../../../../dev-utils/test-config'
 describe('index', function() {
   const globalConfig = getGlobalConfig()
   const { serverUrl, serviceName } = globalConfig.agentConfig
-  var originalTimeout
-
-  beforeEach(function() {
-    originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL
-    jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000
```
jahtalab: same as above!

packages/rum/test/specs/apm-base.spec.js
vigneshshanmugam: It was on apm-rum-react if i remember correctly.

dev-utils/karma.js
vigneshshanmugam: Nope,Since there is a wait time and also retries involved. Also with this change, I no longer see broken tests due to timeouts in saucelabs. A nice side effect 😄

packages/rum-core/test/common/apm-server.spec.js
vigneshshanmugam: reverted and added a comment. Also, we don't do this on performance-monitoring for example so it might not be an issue at all.

packages/rum/test/specs/index.spec.js
```diff
@@ -108,6 +108,7 @@ describe('index', function() {
             },
             reason => {
               fail('Failed sending error:', reason)
+              done()
```
vigneshshanmugam: this was hanging the test on failures.

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -52,10 +52,11 @@ describe('ApmBase', function() {
 
     apmBase.setInitialPageLoadName('new page load')
     apmBase.observe(TRANSACTION_END, endedTr => {
+      expect(endedTr).toEqual(tr)
```
jahtalab: nit: `toEqual` only checks the properties but not if they are actually the same instance. In this case it's not going to matter that much though.

packages/rum/test/specs/index.spec.js
jahtalab: The docs suggest using `done.fail()`! I couldn't find anything in regards to both calling `fail` and `done`! Just want to make sure calling `done` after `fail` doesn't pass the test inadvertently!

packages/rum/test/specs/apm-base.spec.js
vigneshshanmugam: It checks deep equals which should check most of the transaction properties so its good anyways.

packages/rum/test/specs/index.spec.js
vigneshshanmugam: The test would still fail in this case, done is just to make sure we dont wait till the timeout and report the error as soon as possible.

PR #460, opened 2019-10-18
feat: vue router integration with rum agent

### LINKED ISSUE #314: Provide router integration for Vue.js
Similar to what we have done in #265, we should provide router integration for Vue.js.

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#314 
+ Exposes the API through Vue Plugins which would let the user to pass in necessary option and also allow users to access the apm instance throughout the vue app using `this.$apm` variable. 
```.js
import { ApmVuePlugin } from '@elastic/apm-rum-vue'
import VueRouter from 'vue-router'

const router = new VueRouter({})

Vue.use(ApmVuePlugin, {
   router,
   config: {
      serviceName: 'vue-app'
   }
})

```

##  TODO
+ [x] Add Unit tests
+ [x] add docs to supported technologies elastic/apm-agent-rum-js#465 
+ [x] Notice generation Notice.txt
+ [x] Add E2E testss


packages/rum-vue/src/index.js
```diff
@@ -0,0 +1,41 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import { apmBase } from '@elastic/apm-rum'
+import { routeHooks } from './route-hooks'
+
+export function ApmVuePlugin(Vue, options) {
```
jahtalab: According to  [the docs](https://vuejs.org/v2/guide/plugins.html), the plugin should expose an `install` method! this might do the same thing as well, but we probably should follow docs.

packages/rum-vue/test/specs/index.spec.js
```diff
@@ -0,0 +1,154 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import '../polyfill'
```
jahtalab: This shouldn't polyfill the whole test, can we move this to the file level!

packages/rum-vue/src/route-hooks.js
```diff
@@ -0,0 +1,50 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+export function routeHooks(router, apm) {
+  let transaction
+  router.beforeResolve((to, from, next) => {
```
jahtalab: I think we should use `beforeEach` here instead, `beforeResolve` is called after the route is "confirmed", which includes checking other `beforeEach` hooks and since each of these hooks could resolve asynchronously the whole route change could take longer because of these.

packages/rum-vue/src/route-hooks.js
```diff
@@ -0,0 +1,50 @@
...
+      path = matched[matched.length - 1].path || path
+    }
+    transaction = apm.startTransaction(path, 'route-change', {
+      managed: true,
+      canReuse: true
+    })
+    next()
+  })
+
+  router.afterEach(() => {
```
jahtalab: We can also use `beforeRouteEnter` and pass a callback to the `next` function which will be called after the DOM is updated. See [the 12th step in navigation flow](https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards)

packages/rum-vue/src/index.js
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Initialize the APM with the config
+   */
+  const apmInstance = apm.init(config)
+
+  routeHooks(router, apm)
+  /**
+   * Provide it via $apm to be accessed in all Vue Components
+   */
+  Vue.prototype.$apm = apmInstance
```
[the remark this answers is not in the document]
vigneshshanmugam: Yup, found it here https://vuejs.org/v2/guide/plugins.html and also https://github.com/vuejs/vue-router/blob/dev/src/install.js#L42

packages/rum-vue/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "directory": "packages/rum-angular"
+  },
+  "bugs": {
+    "url": "https://github.com/elastic/apm-agent-rum-js/issues"
+  },
+  "dependencies": {
+    "@elastic/apm-rum": "file:../rum"
+  },
+  "peerDependencies": {
+    "vue-router": "^3.0.0"
```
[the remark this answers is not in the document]
vigneshshanmugam: Nope, we don't depend on any internal exports of Vue.

packages/rum-vue/src/route-hooks.js
```diff
@@ -0,0 +1,50 @@
...
+  let transaction
+  router.beforeResolve((to, from, next) => {
+    /**
+     * Get the last matched route record which acts as stack when
+     * route changes are pushed and popped out of the stack
+     */
+    const matched = to.matched || []
+    let path = to.path
+    if (matched.length > 0) {
+      path = matched[matched.length - 1].path || path
```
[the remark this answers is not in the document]
vigneshshanmugam: to.path doesn't have slug pattern.  so if the route has path `/a/:id`, and the current route goes to `/a/1` - to path would reflect `/a/1` wheres last match would exactly match the slug `/a/:id`

packages/rum-vue/src/index.js
vigneshshanmugam: I mean we can rename the method to Install if its better, but its just a convention. 
Also the install method would make sense if we are doing directives, mixins and so on. Its unnecessary for our plugin

packages/rum-vue/src/route-hooks.js
vigneshshanmugam: We should not do `beforeEach` since the user might cancel the current navigation and can install route guards before that. Like checking for authentication and so on. `beforeResolve` accounts for those cases as well and starts once navigation is confirmed with all checks. Its the safest.

packages/rum-vue/src/route-hooks.js
vigneshshanmugam: It's a component level hook and wont work for us, We don't expose any component in our package.

packages/rum-vue/test/specs/index.spec.js
vigneshshanmugam: `Polyfill.js` includes these two lines 

```
import 'core-js/es/object/assign'
import 'es6-promise/auto'
```

which is required for the test to pass. It's already on file level.

packages/rum-vue/src/route-hooks.js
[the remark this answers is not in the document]
jahtalab: Decided to add a comment in the code!

packages/rum-vue/test/e2e/app.e2e-spec.js
```diff
@@ -0,0 +1,59 @@
...
+        /**
+         * route to /fetch
+         */
+        $('#fetch').click()
+        const fetchResult = $('#content')
+        return fetchResult.getText().indexOf('loaded data.json') !== -1
+      },
+      5000,
+      'expected data.json to be loaded'
+    )
```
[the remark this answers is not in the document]
jahtalab: Decided to keep it like this for now, we probably should refactor it later on though.

packages/rum-vue/src/route-hooks.js
vigneshshanmugam: Vue router calls `beforeEach` and `beforeResolve` before our history.pushstate works so we might miss some fetches that happen in components `beforeEach`. I am moving our logic to `beforeEach` which should fix the problem. In the worst case, if the component beforeEach logic takes more than 5 seconds which is our reuse interval. we might end up creating new transaction to capture the route-change. Its not a huge deal though, since the users would call `next(nextRoute)` in an async way on each hook, so we should be good here. WDYT @jahtalab ?

vigneshshanmugam:
@jahtalab Did some checks, Vue router exposes the `init` method on the class since it exposes a mixin and its called when mixin is created https://github.com/vuejs/vue-router/blob/ea8cb474f869a5a12a095fcb5989c45c68971d14/src/install.js#L26

IMO, we had to call the apm initialisation during the plugin install phase
+ General makes it easier for the integration since users does not need to worry about initialising it later
+ We need the initialised agent before using it on the route hooks. If users are initialising later with configs that we dont expect like `active: false` we might be instrumenting the code which feels wrong. 

There are also other cases where things can go wrong, so its better to stick to the current approach and change later on if its really required by the user.

packages/rum-vue/src/route-hooks.js
```diff
@@ -0,0 +1,64 @@
...
+export function routeHooks(router, apm) {
+  let transaction
+
+  function scheduleTransactionFinish() {
+    if (transaction) {
+      Promise.resolve().then(() => transaction.detectFinish())
+    }
+  }
+
+  router.beforeEach((to, from, next) => {
```
jahtalab: I think we might need both `beforeEach` and `beforeResolve`.  The reason is that during `beforeEach` callbacks they might decide to move to another route instead of the one that initiated the original event. For example if the user is not logged in instead of going to the actual route it would go to the login route.

We probably need test for this scenario as well.

packages/rum-vue/src/route-hooks.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+import { Promise } from 'es6-promise'
+
+export function routeHooks(router, apm) {
+  let transaction
+
+  function scheduleTransactionFinish() {
+    if (transaction) {
+      Promise.resolve().then(() => transaction.detectFinish())
```
jahtalab: This might result in a race condition in which the `transaction` variable has been change by other parts of the function.

packages/rum-vue/test/e2e/app.e2e-spec.js
```diff
@@ -0,0 +1,75 @@
...
+     * `app` span type will not be captured in safari 9 since
+     * User Timing API is not supported.
+     */
+    const { name } = getBrowserInfo()
+    if (name.indexOf('safari') >= 0) {
+      expect(foundSpans.length).toEqual(1)
+    } else {
+      expect(foundSpans.length).toEqual(2)
+    }
+  })
```
jahtalab: Can we use something else instead of user timing API (like creating a span by calling `$apm.startSpan()`)?

This adds technical debt that we need to clean up later on.

packages/rum-vue/src/route-hooks.js
```diff
@@ -0,0 +1,64 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import { Promise } from 'es6-promise'
```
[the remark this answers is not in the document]
jahtalab: Decided to refactor this logic into `scheduleMacroTask` and expose it from apm core.

packages/rum-vue/src/route-hooks.js
jahtalab: Decided to keep the current approach and test if it's actually a problem, separately

packages/rum-vue/src/route-hooks.js
```diff
@@ -0,0 +1,64 @@
...
+    })
+    next()
+  })
+
+  router.afterEach(scheduleTransactionFinish)
+  /**
+   * hanbdle when the navigation is cancelled in `beforeEach` hook of components
+   * where `next(error)` is called
+   */
+  router.onError(scheduleTransactionFinish)
```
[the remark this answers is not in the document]
jahtalab: We should check if ending the transaction causes any issue (based on past experience)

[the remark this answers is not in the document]
vigneshshanmugam: this was the comment which i was referring to, Not applicable for vue. we can end the transaction https://github.com/elastic/apm-agent-rum-js/pull/419/commits/2849cdfd1a0ecd582dde2cab5660379b4bb2483e

PR #464, opened 2019-10-24
feat(rum-core): align breakdown types based on navigation timing

vigneshshanmugam:
+ part of elastic/apm-agent-rum-js#289 
+ To be able to drill down in to network more and also make the explanation easier for the users since it closely matches the navigation timing spec.

### DIFF
--- a/packages/rum-core/src/performance-monitoring/breakdown.js
+++ b/packages/rum-core/src/performance-monitoring/breakdown.js
@@ -32,9 +32,12 @@ import { PAGE_LOAD } from '../common/constants'
  * Interested events from the Navigation timing API
  */
 const pageLoadBreakdowns = [
-  ['fetchStart', 'responseEnd', 'Network'],
-  ['domLoading', 'domInteractive', 'DOM Processing'],
-  ['domInteractive', 'loadEventEnd', 'Page Render']
+  ['domainLookupStart', 'domainLookupEnd', 'DNS'],
+  ['connectStart', 'connectEnd', 'TCP'],
+  ['requestStart', 'responseStart', 'Request'],
+  ['responseStart', 'responseEnd', 'Response'],
+  ['domLoading', 'domComplete', 'Processing'],
+  ['loadEventStart', 'loadEventEnd', 'Load']
 ]
 
 function getValue(value) {
--- a/packages/rum-core/test/performance-monitoring/breakdown.spec.js
+++ b/packages/rum-core/test/performance-monitoring/breakdown.spec.js
@@ -64,7 +64,7 @@ describe('Breakdown metrics', () => {
     const pageLoadTransaction = createTransaction(PAGE_LOAD)
     pageLoadTransaction.end()
     const breakdown = captureBreakdown(pageLoadTransaction, TIMING_LEVEL1_ENTRY)
-    const spanTypes = ['Network', 'DOM Processing', 'Page Render']
+    const spanTypes = ['Request', 'Response', 'Processing', 'Load']
 
     const spans = breakdown.slice(1)
 
--- a/packages/rum-core/test/performance-monitoring/breakdown.spec.js
+++ b/packages/rum-core/test/performance-monitoring/breakdown.spec.js
@@ -64,7 +64,14 @@ describe('Breakdown metrics', () => {
     const pageLoadTransaction = createTransaction(PAGE_LOAD)
     pageLoadTransaction.end()
     const breakdown = captureBreakdown(pageLoadTransaction, TIMING_LEVEL1_ENTRY)
-    const spanTypes = ['Request', 'Response', 'Processing', 'Load']
+    const spanTypes = [
+      'DNS',
+      'TCP',
+      'Request',
+      'Response',
+      'Processing',
+      'Load'
+    ]
 
     const spans = breakdown.slice(1)
 
--- a/packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
+++ b/packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
@@ -67,6 +67,7 @@ describe('Capture hard navigation', function() {
         _end: 542,
         _start: 165
       },
+      { name: 'Fire "DOMContentLoaded" event', _end: 654, _start: 634 },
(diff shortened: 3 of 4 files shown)


packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -32,9 +32,12 @@ import { PAGE_LOAD } from '../common/constants'
...
 const pageLoadBreakdowns = [
-  ['fetchStart', 'responseEnd', 'Network'],
-  ['domLoading', 'domInteractive', 'DOM Processing'],
-  ['domInteractive', 'loadEventEnd', 'Page Render']
+  ['domainLookupStart', 'domainLookupEnd', 'DNS'],
+  ['connectStart', 'connectEnd', 'TCP'],
+  ['requestStart', 'responseStart', 'Request'],
+  ['responseStart', 'responseEnd', 'Response'],
+  ['domLoading', 'domComplete', 'Processing'],
+  ['loadEventStart', 'loadEventEnd', 'Load']
```
jahtalab: I think `Load event` is more clear here.

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -32,9 +32,12 @@ import { PAGE_LOAD } from '../common/constants'
  * Interested events from the Navigation timing API
  */
 const pageLoadBreakdowns = [
-  ['fetchStart', 'responseEnd', 'Network'],
-  ['domLoading', 'domInteractive', 'DOM Processing'],
-  ['domInteractive', 'loadEventEnd', 'Page Render']
+  ['domainLookupStart', 'domainLookupEnd', 'DNS'],
```
jahtalab: Don't we want to include the time between `fetchStart` and `domainLookupStart`? The documentation suggests this time is used for checking application cache.

packages/rum-core/test/performance-monitoring/breakdown.spec.js
```diff
@@ -64,7 +64,7 @@ describe('Breakdown metrics', () => {
     const pageLoadTransaction = createTransaction(PAGE_LOAD)
     pageLoadTransaction.end()
     const breakdown = captureBreakdown(pageLoadTransaction, TIMING_LEVEL1_ENTRY)
-    const spanTypes = ['Network', 'DOM Processing', 'Page Render']
+    const spanTypes = ['Request', 'Response', 'Processing', 'Load']
```
jahtalab: This should include `DNS`, `TCP` as well.

packages/rum-core/src/performance-monitoring/breakdown.js
vigneshshanmugam: AppCache https://www.html5rocks.com/en/tutorials/appcache/beginner/ is no more and its not a useful metric in this scenario and we don't need to account for duration of the whole transaction in the breakdown.

packages/rum-core/test/performance-monitoring/breakdown.spec.js
vigneshshanmugam: The example data has 0 value for them, hence we ignore 0 duration from posting, we dont see them here.

packages/rum-core/src/performance-monitoring/breakdown.js
vigneshshanmugam: Kinda agree with you, but it's lengthy. `Load` or `Onload` is bit small. Plus we have explanation docs for them, so we don't need to be more explicit here.

axw: @jahtalab do you think "Load" is unclear? I suppose "Load Event" would be more in line with the spec, but this seems clear enough to me.

packages/rum-core/src/performance-monitoring/breakdown.js
```diff
@@ -32,9 +32,12 @@ import { PAGE_LOAD } from '../common/constants'
...
+  ['connectStart', 'connectEnd', 'TCP'],
```
axw: Should we be calling this Connect rather than TCP, to account for QUIC? Or perhaps we just substitute QUIC for TCP when we identify its use?

vigneshshanmugam: Might be a bit misleading for QUIC since it operates on UDP layer, But the features of QUIC includes 0-RTT which would eliminate this metric altogether. Shall we revisit this decision in the future?

packages/rum-core/src/performance-monitoring/breakdown.js
jahtalab: I think "Load" is an overloaded :) term. Personally if I see "Load" in another UI, I would be thinking which "Load" is it about, since there are multiple parts of a page loading process that can be referred to as "Load".

vigneshshanmugam: Can we rename it to `Onload` ?

packages/rum-core/test/performance-monitoring/breakdown.spec.js
jahtalab: Decided to add an navigatio entry to have all the entry types!

packages/rum-core/src/performance-monitoring/breakdown.js
jahtalab: we decided to use `Load` for now, we can change it later if necessary!

packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
```diff
@@ -118,8 +120,9 @@ describe('Capture hard navigation', function() {
         transactionEnd
       )
       expect(spans.map(mapSpan)).toEqual([
-        { name: 'Fire "DOMContentLoaded" event', _end: 1498, _start: 1464 },
```
jahtalab: Is removing `Fire "DOMContentLoaded" event` intentional? I see that the difference in the event entries is zero, but I think we should provide the required entries so that these spans are tested properly. 
IMO, over time this would result in our tests becoming more comprehensive.

vigneshshanmugam: Will check what I can do, it was hard to satisfy all the values.

PR #465, opened 2019-10-24
docs: add vue package integration docs [skip ci]

vigneshshanmugam:
+ part of elastic/apm-agent-rum-js#460

### DIFF
--- a/docs/vue-integration.asciidoc
+++ b/docs/vue-integration.asciidoc
@@ -55,35 +55,31 @@ Instance of the agent can be accessed inside all the components using `this.$apm
 [source,html]
 ----
 <template>
-  <h1>Current Transaction Name</h1>
-  <h2>{{ name }}</h2>
+  <div>Component timings as span</div>
 </template>
+
 <script>
 export default {
-  name: 'test',
-  data () {
+  data() {
     return {
-      name: null
+      span: null
     }
   },
-  created () {
-    this.getName()
-  }
-  methods: {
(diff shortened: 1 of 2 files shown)


vigneshshanmugam:
Thanks @bmorelli25 Applied the suggestions :)

docs/vue-integration.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+    }
+  },
+  created () {
+    this.getName()
+  }
+  methods: {
+    getName() {
+      let transaction = this.$apm.getCurrenTransaction()
+      if (transaction) {
+        this.name = transaction.name
```
jahtalab: Maybe add a more common use-case here, like creating a span side the component

docs/vue-integration.asciidoc
```diff
@@ -0,0 +1,89 @@
...
+      }
+    }
+  }
+}
+</script>
+----
+
+
+`ApmVuePlugin` uses the Navigation guards functionality provided by the VueRouter and will not work
+with any other router instance provided via the plugin options.
```
jahtalab: I think the `will not work with any other router instance provided via the plugin options` part is ambiguous. Can you please explain what you wanted to express here?

vigneshshanmugam: I was going for `router` option passed to ApmVuePlugin should be an instance of VueRouter.

PR #466, opened 2019-10-24
docs: Add Central Config documentation

jahtalab:
part of elastic/apm-agent-rum-js#253

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -83,7 +83,7 @@ A list of instrumentations which can be disabled. Valid options are `xmlhttprequ
 === `environment`
 
 * *Type:* String
-* *Default:* ``
+* *Default:* `''`
 
 Environment where the service being monitored is deployed, e.g. "production", "development", "test", etc.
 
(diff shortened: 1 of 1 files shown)


docs/configuration.asciidoc
```diff
@@ -239,5 +239,22 @@ By default the agent can only send (up to) `20` transactions every `30000` milli
...
 
+[float]
+[[central-config]]
+==== `centralConfig`
+
+* *Type:* Boolean
+* *Default:* `false`
+
+This option activates APM Agent Configuration via Kibana.
+If set to `true`, the agent will fetch the central configuration via the APM Server on the initialization phase.
```
vigneshshanmugam: Better to say it as the agent starts polling for configuration during the initialisation phase and the config is stored in the

jahtalab: Personally I'm not a fan of `polling` as it is ambiguous and implies repeated fetching, I think using fetch is more clear.

vigneshshanmugam: Yeah polling was wrong, i rephrased it a bit.

bmorelli25: I think what you have works. I also think @vigneshshanmugam's suggestion would work. For fun, if I were writing this, here's how I would phrase it:
```
If set to `true`, the agent starts fetching configurations via the APM Server during the initialization phase.
These central configurations are cached in `sessionStorage`, and will not be fetched again until
the session is closed and/or `sessionStorage` is cleared. In most cases, this means when the tab/window of the page is closed.
```

docs/configuration.asciidoc
```diff
@@ -239,5 +239,22 @@ By default the agent can only send (up to) `20` transactions every `30000` milli
...
+The agent caches the central configuration in the `sessionStorage` and it does not fetch them again until
+the session is closed and/or `sessionStorage` is cleared, in most cases this means when the tab/window of the page is closed.
+
+NOTE: Currently, only <<transaction-sample-rate, transaction sample rate>> can be configured via Kibana.
+
+NOTE: This feature requires APM Server v7.5 or later and that the APM Server is configured with `kibana.enabled: true`.
+More information is available in {kibana-ref}/agent-configuration.html[APM Agent configuration].
```
bmorelli25: I'm in the process of updating the attribute we use to link to the APM app docs (because they're gonna be moving). Can you use `apm-app-ref` instead please?

```suggestion
More information is available in {apm-app-ref}/agent-configuration.html[APM Agent configuration].
```

jahtalab:
Thanks all for the comments! I've made some changes!

PR #469, opened 2019-10-25
feat(rum-core): improve the debug logs with transaction details

### LINKED ISSUE #452: Improve debug logs with transaction details
+ Since we introduced `managed` transaction and there can be more than 1 transaction active at a time in the application, it becomes harder to read debug logs since we are not sure to which transaction the spans or tasks are associated with. 

Makes it much harder to read the debug logs from the user. We can add transaction details like {name, type} to the existing logs to keep it simple and also useful.

vigneshshanmugam:
+ fixes elastic/apm-agent-rum-js#452 
+ Annotates the debug logs with transaction name and type on each log
+ Prefix the log with `[Elastic APM] `  for grouping on the console. 

Looks like below image on the console. 
Screenshot 2019-11-11 at 11 40 33

### DIFF
--- a/packages/rum-core/src/common/service-factory.js
+++ b/packages/rum-core/src/common/service-factory.js
@@ -43,7 +43,7 @@ class ServiceFactory {
     })
     this.registerServiceCreator('LoggingService', function() {
       return new LoggingService({
-        prefix: '[JS Agent] '
+        prefix: '[Elastic APM] '
       })
     })
     this.registerServiceCreator('ApmServer', function() {
--- a/packages/rum-core/src/common/service-factory.js
+++ b/packages/rum-core/src/common/service-factory.js
@@ -42,7 +42,9 @@ class ServiceFactory {
       return new ConfigService()
     })
     this.registerServiceCreator('LoggingService', function() {
-      return new LoggingService()
+      return new LoggingService({
+        prefix: '[JS Agent] '
+      })
     })
     this.registerServiceCreator('ApmServer', function() {
       return new ApmServer(
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -66,11 +66,6 @@ class PerformanceMonitoring {
       const payload = this.createTransactionPayload(tr)
       if (payload) {
         this._apmServer.addTransaction(payload)
-      } else if (__DEV__) {
-        this._logginService.debug(
-          'Could not create a payload from the Transaction',
-          tr
-        )
       }
     })
 
@@ -217,7 +212,7 @@ class PerformanceMonitoring {
 
     if (!duration) {
       if (__DEV__) {
-        let message = 'Transaction was discarded! '
+        let message = `Transaction(${tr.name}) was discarded! `
         if (duration === 0) {
           message += `Transaction duration is 0`
         } else {
@@ -231,7 +226,9 @@ class PerformanceMonitoring {
     if (duration > transactionDurationThreshold) {
       if (__DEV__) {
         this._logginService.debug(
-          `Transaction was discarded! Transaction duration (${duration}) is greater than the transactionDurationThreshold configuration (${transactionDurationThreshold})`
+          `Transaction(${
+            tr.name
+          }) was discarded! Transaction duration (${duration}) is greater than the transactionDurationThreshold configuration (${transactionDurationThreshold})`
         )
       }
       return false
@@ -240,7 +237,9 @@ class PerformanceMonitoring {
     if (tr.spans.length === 0) {
       if (__DEV__) {
         this._logginService.debug(
-          `Transaction was discarded! Transaction does not include any spans`
+          `Transaction(${
+            tr.name
+          }) was discarded! Transaction does not include any spans`
         )
       }
       return false
@@ -273,7 +272,9 @@ class PerformanceMonitoring {
       if (!wasBrowserResponsive) {
         if (__DEV__) {
           this._logginService.debug(
-            'Transaction was discarded! Browser was not responsive enough during the transaction.',
+            `Transaction(${
+              tr.name
+            }) was discarded! Browser was not responsive enough during the transaction.`,
             ' duration:',
             duration,
             ' browserResponsivenessCounter:',
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -212,7 +212,7 @@ class PerformanceMonitoring {
 
     if (!duration) {
       if (__DEV__) {
-        let message = `Transaction(${tr.name}) was discarded! `
+        let message = `Transaction(${tr.id}, ${tr.name}) was discarded! `
         if (duration === 0) {
           message += `Transaction duration is 0`
         } else {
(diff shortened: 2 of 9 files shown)


vigneshshanmugam:
Jenkins is not able to build due to an APM Jenkins CI issue.

packages/rum-core/src/common/service-factory.js
```diff
@@ -42,7 +42,9 @@ class ServiceFactory {
       return new ConfigService()
     })
     this.registerServiceCreator('LoggingService', function() {
-      return new LoggingService()
+      return new LoggingService({
+        prefix: '[JS Agent] '
```
jahtalab: What do you think about using `Elastic APM` here instead?

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -66,11 +66,6 @@ class PerformanceMonitoring {
       const payload = this.createTransactionPayload(tr)
       if (payload) {
         this._apmServer.addTransaction(payload)
-      } else if (__DEV__) {
-        this._logginService.debug(
-          'Could not create a payload from the Transaction',
-          tr
-        )
```
jahtalab: Why is this no longer needed?

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -217,7 +212,7 @@ class PerformanceMonitoring {
 
     if (!duration) {
       if (__DEV__) {
-        let message = 'Transaction was discarded! '
+        let message = `Transaction(${tr.name}) was discarded! `
```
jahtalab: Maybe it should be the transaction id instead, WDYT?

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -180,7 +176,7 @@ class TransactionService {
     tr.onEnd = () => this.handleTransactionEnd(tr)
 
     if (__DEV__) {
-      this._logger.debug('TransactionService.startTransaction', tr)
+      this._logger.debug('startTransaction', { name: tr.name, type: tr.type })
```
jahtalab: Also having transaction helps here, you can include name and type in the message `startTransaction(${name},${type})`

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -125,11 +122,10 @@ class TransactionService {
...
         this._logger.debug(
-          'Redefining the current transaction',
-          tr,
-          name,
-          type,
-          perfOptions
+          'Redefining current transaction from',
+          { name: tr.name, type: tr.type },
+          'to',
+          { name, type, options: perfOptions }
```
jahtalab: I think here having the full transaction helps. But you can instead include name and type in the message and pass the transaction as the last argument

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: We have a more specific log of why we are discarding it. Its just double log which feels unnecessary. 

We have both of these
1. we discard the transaction
2. could not create payload. 

Having just the 1st one feels clear and not polluting the logs as well.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: We could definitely log id. But wont it be hard for user to figure out what that means?

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: Sure, so remove options and instead add transaction instead?

packages/rum-core/src/common/service-factory.js
vigneshshanmugam: Yeah sounds good.

vigneshshanmugam: Done.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: Logs id and name

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: Changed the format now.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -140,7 +136,7 @@ class TransactionService {
       tr.redefine(name, undefined, perfOptions)
     } else {
       if (__DEV__) {
-        this._logger.debug('Ending old transaction', tr)
+        this._logger.debug(`ending old transaction(${tr.id}, ${tr.name})`, tr)
```
jahtalab: nit: Maybe use `previous` instead of `old`.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -125,11 +122,10 @@ class TransactionService {
        */
       if (__DEV__) {
         this._logger.debug(
-          'Redefining the current transaction',
-          tr,
-          name,
-          type,
-          perfOptions
+          `redefining transaction(${tr.id}, ${tr.name}, ${tr.type})`,
```
jahtalab: nit: `Redefining Transaction` (capital letters)

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -190,21 +186,24 @@ class TransactionService {
   handleTransactionEnd(tr) {
     return Promise.resolve().then(
       () => {
-        if (this.shouldIgnoreTransaction(tr.name)) {
+        const { name, type } = tr
+        if (this.shouldIgnoreTransaction(name)) {
           if (__DEV__) {
-            this._logger.debug('TransactionService transaction is ignored', tr)
+            this._logger.debug(
+              `transaction(${tr.id}, ${name}, ${type}) is ignored`
```
jahtalab: nit: `Transaction`

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -224,12 +223,15 @@ class TransactionService {
         }
         this._config.events.send(TRANSACTION_END, [tr])
         if (__DEV__) {
-          this._logger.debug('TransactionService transaction ended', tr)
+          this._logger.debug(`end transaction(${tr.id}, ${tr.name})`, tr)
```
jahtalab: nit: Capital first letters!

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -224,12 +223,15 @@ class TransactionService {
...
         }
       },
       err => {
         if (__DEV__) {
-          this._logger.debug('TransactionService transaction onEnd', err)
+          this._logger.debug(
+            `error ending transaction(${tr.id}, ${tr.name})`,
```
jahtalab: nit: Capital first letters!

vigneshshanmugam: I went with small since we have function names as `startTransaction` and it was a bit odd seeing only these in capital letters.

jahtalab:
We decided to use small letters since we have the prefix `[Elastic APM]`!

PR #489, opened 2019-11-18
fix: use rum endpoint for central config

jahtalab:


.travis.yml
```diff
@@ -34,23 +34,23 @@ jobs:
       name: 'Analyse size of JS bundles'
     - stage: test
       env:
-        - STACK_VERSION=7.3.0
+        - STACK_VERSION=8.0.0-SNAPSHOT
```
vigneshshanmugam: We already do this on Jenkins? Any reason why  you switched to 8.0 instead of 7.5.0?

jahtalab: 7.5 is not released yet! The same issue would exist on any new feature for future versions. The snapshot would give us the latest APM server version!

PR #503, opened 2019-11-20
fix(rum): page-load transaction duration should reflect load event

### LINKED ISSUE #474: "page-load" transaction._end value is not proper
Hi, we've been using for couple of months. And recently, we found that there are some cases where `transaction.duranton` is abnormally long. After read through the code, I found that `transaction._end` will get the latest `span._end` value, which I think is not very proper:
1. if user fire a xhr that takes very long time before `load event`, transaction will not be sent before the xhr ends although `load` event is fired immediately.
2. normally we'll gather information from kibana -> APM, where `Avg.duration` and `95th  percentile` value is calculated using `transaction.duration` value, which is not corresponding to the real `load event` time.
3. the function name here is called `getLatestNonXHRSpan`, but is it implemented properly? If YES, it's understandable because xhr will not influence the `load event` time where resource fetching will.
https://github.com/elastic/apm-agent-rum-js/blob/228e157851f4df7448f8bfcdd4b4b57129707992/packages/rum-core/src/common/utils.js#L360

And if I misunderstood something, sorry about that.🤭

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#474 
+ `page-load` transaction duration must always reflect the window.load event value. Even if there are tasks associated with the transaction, they should be truncated and added to the page-load transaction instead of extending the duration of the transaction

### DIFF
--- a/packages/rum/src/apm-base.js
+++ b/packages/rum/src/apm-base.js
@@ -25,6 +25,7 @@
 
 import {
   getInstrumentationFlags,
+  scheduleMacroTask,
   PAGE_LOAD,
   ERROR
 } from '@elastic/apm-rum-core'
@@ -126,7 +127,6 @@ class ApmBase {
     const transactionService = this.serviceFactory.getService(
       'TransactionService'
     )
-
     /**
      * Name of the transaction is set in transaction service to
      * avoid duplicating the logic at multiple places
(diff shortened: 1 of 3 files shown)


packages/rum/src/apm-base.js
```diff
@@ -135,15 +135,15 @@ class ApmBase {
       managed: true,
       canReuse: true
     })
-
     if (tr) {
       tr.addTask(PAGE_LOAD)
```
vigneshshanmugam: we still have the task here since the page-load cannot be closed by any other detectfinish logic

vigneshshanmugam:
We had a discussion and decided that page-load transaction can extend beyond the duration of the load event and we wanted to capture the user experience as a whole for the hard navigation instead of ending the transaction as soon as the load event.

PR #506, opened 2019-11-28
ci(jenkins): trigger opbeans release process

v1v:
## What does this pull request do?

Add a final stage in the release process to bump the opbeans agent dependency.

## Why is it important?

When a new tag, aka a new release, is created then, the automation will kick the opbeans build which consists of the following tasks:
- [x] Create a tag to trigger the release pipeline in the opbeans-frontend. That particular tag will match with the same one that the one in the agent itself.
- [x] Monitor tags which are matching the pattern `@elastic/apm-rum@x.y.z`

The opbeans-frontend pipeline will trigger the build for the just created tag.

## Tasks
- [x] Agree the version should be either x.y.z or agent tag-based, like https://github.com/elastic/apm-agent-rum-js/releases/tag/%40elastic%2Fapm-rum%404.5.1

> Versioning for both will always match with the tag.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -195,7 +195,10 @@ pipeline {
                 dir("${OPBEANS_REPO}"){
                   git credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba',
                       url: "git@github.com:elastic/${OPBEANS_REPO}.git"
-                  // The opbeans-go pipeline will trigger a release for the release tag
+                  sh script: ".ci/bump-version.sh '${env.BRANCH_NAME}'", label: 'Bump version'
+                  // The opbeans pipeline will trigger a release for the master branch
+                  gitPush()
+                  // The opbeans pipeline will trigger a release for the release tag
                   gitCreateTag(tag: "${env.BRANCH_NAME}")
                 }
               }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -178,6 +179,32 @@ pipeline {
...
+          }
+          stages {
+            stage('Opbeans') {
+              environment {
+                REPO_NAME = "${OPBEANS_REPO}"
+              }
+              steps {
+                deleteDir()
+                dir("${OPBEANS_REPO}"){
+                  git credentialsId: 'f6c7695a-671e-4f4f-a331-acdce44ff9ba',
```
vigneshshanmugam: don't we need to encrypt this token? or it it read only?

v1v: Already encrypted, that's the internal hash id, nothing to be worried about

v1v:
Regarding the test failures, this is not related to this particular PR, but an environmental issue when provisioning one of the CI Workers for the tests `Test / Rum:8.0.0-SNAPSHOT#@elastic/apm-rum-core`

```
[2019-11-28T16:36:57.335Z] apm-ci-immutable-ubuntu-1604-1574958786363281979 was marked offline: Connection was broken: java.nio.channels.ClosedChannelException
```

jenkins run the tests please

v1v:
Retriggered and everything is now green! Merging the PR now

PR #508, opened 2019-12-02
chore(benchmarks): memory allocation per function during page-load [s…

vigneshshanmugam:
+ part of elastic/apm-agent-rum-js#26 
+ This PR uses the chrome allocation profiler to measure the memory allocation of top functions from the agent that might help identify memory leaks
+ Generates a new build with keeping the function names of the apm agent to help with identify the functions with more memory allocation. 

Adds some keys to the benchmark documents that matches with the function names that uses the most memory 

```
   "memory.createResourceTimingSpan.bytes": 65664,
    "memory.generateRandomId.bytes": 33280,
    "memory.createResourceTimingSpans.bytes": 32864,
    "memory.getCurrentScript.bytes": 32816,
    "memory.patchFetch.bytes": 32816,
    "memory.baseExtend.bytes": 32784,
    "memory.isCORSSupported.bytes": 38376,
```

### DIFF
--- a/packages/rum/test/benchmarks/analyzer.js
+++ b/packages/rum/test/benchmarks/analyzer.js
@@ -29,6 +29,7 @@ const { readFileSync } = require('fs')
 const path = require('path')
 const zlib = require('zlib')
 const webpack = require('webpack')
+const UglifyJSPlugin = require('uglifyjs-webpack-plugin')
 const {
   getWebpackConfig,
   BUNDLE_TYPES
@@ -40,7 +41,7 @@ const dist = path.join(__dirname, '../../dist')
 function customApmBuild(filename) {
   /**
    * Match it with the default webpack prod build of elasticApm
-   * expect the function names are not mangled
+   * expect function names are not mangled and source map is not generated
    */
   const config = {
     entry: path.join(__dirname, '../../src/index.js'),
@@ -50,7 +51,21 @@ function customApmBuild(filename) {
       library: '[name]',
       libraryTarget: 'umd'
     },
-    ...getWebpackConfig(BUNDLE_TYPES.BROWSER_ESM_PROD)
+    ...getWebpackConfig(BUNDLE_TYPES.BROWSER_ESM_PROD),
+    ...{
+      devtool: false,
+      optimization: {
+        minimizer: [
+          new UglifyJSPlugin({
+            sourceMap: false,
+            extractComments: true,
+            uglifyOptions: {
+              keep_fnames: true
+            }
+          })
+        ]
+      }
+    }
   }
 
   return new Promise((resolve, reject) => {
@@ -225,9 +240,9 @@ function capturePayloadInfo(payload) {
 function getMemoryAllocationPerFunction({ profile }) {
   const allocations = []
   /**
-   * Whitelist built functions from the sample
+   * exclude Function names from the memory sample
    */
-  const whitelist = [
+  const excludeFunctionNames = [
     '(V8 API)',
     '(BYTECODE_COMPILER)',
     '__webpack_require__',
@@ -241,7 +256,7 @@ function getMemoryAllocationPerFunction({ profile }) {
     if (
       selfSize > 0 &&
       functionName !== '' &&
-      !whitelist.includes(functionName)
+      !excludeFunctionNames.includes(functionName)
     ) {
       allocations.push({
         name: functionName,
--- a/packages/rum/test/benchmarks/analyzer.js
+++ b/packages/rum/test/benchmarks/analyzer.js
@@ -37,7 +37,7 @@ const { runs, noOfImages } = require('./config')
 
 const dist = path.join(__dirname, '../../dist')
 
-function customApmBuild() {
+function customApmBuild(filename) {
   /**
    * Match it with the default webpack prod build of elasticApm
    * expect the function names are not mangled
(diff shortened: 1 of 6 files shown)


packages/rum/test/benchmarks/server.js
```diff
@@ -93,4 +95,10 @@ module.exports = function startServer() {
       resolve(server)
     })
   })
-}
+})
+
+!(async () => {
+  if (require.main === module) {
```
vigneshshanmugam: added this so that we can just start the server for testing with `node packages/rum/test/benchmarks/server.js`

packages/rum/test/benchmarks/analyzer.js
```diff
@@ -28,19 +28,48 @@ const stats = require('stats-lite')
...
+
+  return new Promise((resolve, reject) => {
+    webpack(config, err => {
+      if (err) {
+        reject(err)
+      }
+      console.info('custom apm build - ', filename, 'generated')
+      resolve()
+    })
+  })
```
jahtalab: Why this needs to be async?

packages/rum/test/benchmarks/analyzer.js
```diff
@@ -28,19 +28,48 @@ const stats = require('stats-lite')
...
+   */
+  const config = {
+    entry: path.join(__dirname, '../../src/index.js'),
+    output: {
+      filename,
+      path: path.join(dist, 'bundles'),
+      library: '[name]',
+      libraryTarget: 'umd'
+    },
+    ...getWebpackConfig(BUNDLE_TYPES.BROWSER_ESM_PROD)
```
jahtalab: What do you think about moving the configs to dev-utils? e.g. having a separate function for getting the benchmark configs. One plus of this is that we can only set `keep_fnames: true ` on this configuration instead of changing the default production config.

packages/rum/test/benchmarks/analyzer.js
```diff
@@ -181,12 +222,57 @@ function capturePayloadInfo(payload) {
...
+function getMemoryAllocationPerFunction({ profile }) {
+  const allocations = []
+  /**
+   * Whitelist built functions from the sample
+   */
+  const whitelist = [
+    '(V8 API)',
+    '(BYTECODE_COMPILER)',
+    '__webpack_require__',
+    'scriptId'
```
jahtalab: What is the `scriptId`?

packages/rum/test/benchmarks/analyzer.js
```diff
@@ -181,12 +222,57 @@ function capturePayloadInfo(payload) {
   }
 }
 
+function getMemoryAllocationPerFunction({ profile }) {
+  const allocations = []
+  /**
+   * Whitelist built functions from the sample
+   */
+  const whitelist = [
```
jahtalab: I think you mean blacklist here, but even then I think something like `excludeFunctionNames` is better.

packages/rum/test/benchmarks/analyzer.js
```diff
@@ -130,9 +160,20 @@ function calculateResults(resultMap) {
...
+       * Add consumed memory in bytes per each function to the result
        */
-      if (!Array.isArray(value)) {
+      if (metricName === 'memory') {
+        const reducedValue = value.reduce((acc, curr) => {
+          acc.push(...curr)
+          return acc
+        }, [])
+        reducedValue.forEach(obj => {
+          result[`memory.${obj.name}.bytes`] = obj.size
```
jahtalab: This field will be parsed in Elasticsearch into hierarchy of objects, is that right?

packages/rum/test/benchmarks/profiler.js
```diff
@@ -91,9 +103,17 @@ function gatherRawMetrics(browser, url) {
       Object.assign(metrics, timings)
     })
     /**
-     * Start the profiler before navigating to the URL
+     * Perform a garbage collection to do a clean check everytime
+     */
+    await client.send('HeapProfiler.collectGarbage')
```
jahtalab: Why is this necessary? Don't we start with a new window every time?

packages/rum/test/benchmarks/server.js
```diff
@@ -39,28 +39,28 @@ function generateImageUrls(port, number) {
...
-function getRandomBundleContent() {
-  let content = APM_BUNDLE
+function getRandomBundleContent(apmBundle) {
+  let content = apmBundle
   content += `var scriptId = ${Date.now()};`
   return content
 }
 
-module.exports = function startServer() {
+const startServer = (module.exports = filename => {
```
jahtalab: I think it's cleaner to define the `startServer` function and then assign it to the exports.

packages/rum/test/benchmarks/analyzer.js
vigneshshanmugam: Its generated from webpack.

packages/rum/test/benchmarks/profiler.js
vigneshshanmugam: Just to make sure memory is not reused, yes we do start a new window, but just in case.

packages/rum/test/benchmarks/analyzer.js
vigneshshanmugam: so we can await on the build part before starting the server.

packages/rum/test/benchmarks/analyzer.js
vigneshshanmugam: `build.js` is a common for all the builds. I prefer to keep the configs along with the files that are actually doing the job, we can definitely do the same `keep_fname: true` here also without overrding the default one. it would be the same wherever we keep this config.

Issue #513, opened 2019-12-04
Provide ECS destination field on spans


jahtalab:
See https://github.com/elastic/apm/issues/115 for more.

vigneshshanmugam:
These fields are also tested with latest APM server master. Closing the issue.

PR #515, opened 2019-12-04
feat(rum-core): enrich span context with desination metadata

### LINKED ISSUE #490: Add destination.service to spans for service maps feature
See https://github.com/elastic/apm/issues/180 for more
Separate but related: https://github.com/elastic/apm-agent-rum-js/issues/513

task:
- [ ] Add actual payload tests for `context.destination`
- [ ] make sure the snapshot is update with the destination specs from the APM server

vigneshshanmugam:
+ relates elastic/apm-agent-rum-js#490 elastic/apm-agent-rum-js#513 
+ Add the necessary `span.desination.*` fields for all outgoing spans to help with service maps and siem intergration
+ This PR changes the overall context addition which happens across multiple place and unifies them in a single place to ease the development and addition of new fields on context easier. 
+ Custom URL parser is enhanced with more fields `port` and `hostname` to make the destinations fields addition easier. 
+ Context addition happens on span end and transaction end instead of doing it during the middle of the transaction and span creation process. 

## TODO
+ [x] Fix parsing of hostname for `http://[::1]:8080/`
+ [x] Add tests for context
+ [x] Add tests for URL parsing of port and hostname

### DIFF
--- a/packages/rum-core/src/common/context.js
+++ b/packages/rum-core/src/common/context.js
@@ -27,6 +27,9 @@ import Url from './url'
 import { PAGE_LOAD } from './constants'
 import { getPageMetadata, getServerTimingInfo } from './utils'
 
+const LEFT_SQUARE_BRACKET = 91 // [
+const RIGHT_SQUARE_BRACKET = 93 // ]
+
 /**
  * Both Navigation and Resource timing level 2 exposes these below information
  *
@@ -54,13 +57,53 @@ function getResponseContext(perfTimingEntry) {
   return respContext
 }
 
+function getDestination(parsedUrl) {
+  const { port, protocol, hostname, origin } = parsedUrl
+
+  const isDefaultSchemeAndPort =
+    (protocol === 'http:' && port === '80') ||
+    (protocol === 'https:' && port === '443')
+
+  /**
+   * If hostname begins with [ and ends with ] then its an IPV6 address
+   *
+   * since address and port are recorded separately, we are recording the
+   * info in canonical form without square brackets
+   *
+   * IPv6 check is done here instead of in the URL parser to keep the URL parser
+   * functionality minimal and it also makes things easier for service construction
+   */
+  const ipv6Hostname =
+    hostname.charCodeAt(0) === LEFT_SQUARE_BRACKET &&
+    hostname.charCodeAt(hostname.length - 1) === RIGHT_SQUARE_BRACKET
+
+  let address = hostname
+  if (ipv6Hostname) {
+    address = hostname.slice(1, -1)
+  }
+
+  return {
+    service: {
+      name: !isDefaultSchemeAndPort ? origin : protocol + '//' + hostname,
+      resource: hostname + ':' + port,
+      type: 'external'
+    },
+    address,
+    port
+  }
+}
+
 function getResourceContext(data) {
   const { entry, url } = data
+  const parsedUrl = new Url(url)
+
+  const destination = getDestination(parsedUrl)
   return {
     http: {
       url,
       response: getResponseContext(entry)
-    }
+    },
+    destination
   }
 }
 
@@ -68,23 +111,14 @@ function getExternalContext(data) {
   const { url, method, target, response } = data
   const parsedUrl = new Url(url)
 
-  const { href, port, protocol, hostname, origin } = parsedUrl
-  const isDefaultPort = port === '80' || port === '443'
+  const destination = getDestination(parsedUrl)
 
   const context = {
     http: {
       method,
-      url: href
+      url: parsedUrl.href
     },
-    destination: {
-      service: {
-        name: !isDefaultPort ? origin : protocol + '//' + hostname,
-        resource: hostname + ':' + port,
-        type: 'external'
-      },
-      address: hostname,
-      port
-    }
+    destination
   }
 
   let statusCode
--- a/packages/rum-core/src/common/context.js
+++ b/packages/rum-core/src/common/context.js
@@ -99,7 +99,7 @@ function getDestination(parsedUrl, type) {
       type
     },
     address,
-    port: portNumber
+    port: Number(portNumber)
   }
 }
 
--- a/packages/rum-core/src/common/context.js
+++ b/packages/rum-core/src/common/context.js
@@ -68,7 +68,7 @@ function getExternalContext(data) {
   const { url, method, target, response } = data
   const parsedUrl = new Url(url)
 
-  const { href, port, protocol, hostname } = parsedUrl
+  const { href, port, protocol, hostname, origin } = parsedUrl
   const isDefaultPort = port === '80' || port === '443'
 
   const context = {
@@ -86,8 +86,9 @@ function getExternalContext(data) {
       port
     }
   }
+
   let statusCode
-  if (target && typeof target.status !== 'undefined') {
+  if (target) {
     statusCode = target.status
   } else if (response) {
     statusCode = response.status
--- a/packages/rum-core/src/common/context.js
+++ b/packages/rum-core/src/common/context.js
@@ -1,0 +1,131 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2017-present, Elasticsearch BV
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import Url from './url'
+import { PAGE_LOAD } from './constants'
+import { getPageMetadata, getServerTimingInfo } from './utils'
+
+/**
+ * Both Navigation and Resource timing level 2 exposes these below information
+ *
+ * for CORS requests without Timing-Allow-Origin header, transferSize & encodedBodySize will be 0
+ */
+function getResponseContext(perfTimingEntry) {
+  const {
+    transferSize,
+    encodedBodySize,
+    decodedBodySize,
+    serverTiming
+  } = perfTimingEntry
+
+  const respContext = {
+    transfer_size: transferSize,
+    encoded_body_size: encodedBodySize,
+    decoded_body_size: decodedBodySize
+  }
+  const serverTimingStr = getServerTimingInfo(serverTiming)
+  if (serverTimingStr) {
+    respContext.headers = {
+      'server-timing': serverTimingStr
+    }
+  }
+  return respContext
+}
+
+function getResourceContext(data) {
+  const { entry, url } = data
+  return {
+    http: {
+      url,
+      response: getResponseContext(entry)
+    }
+  }
+}
+
+function getExternalContext(data) {
+  const { url, method, target, response } = data
+  const parsedUrl = new Url(url)
+
+  const { href, port, protocol, hostname } = parsedUrl
+  const isDefaultPort = port === '80' || port === '443'
+
+  const context = {
+    http: {
+      method,
+      url: href
+    },
+    destination: {
+      service: {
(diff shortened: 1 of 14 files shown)


packages/rum-core/src/common/context.js
```diff
@@ -0,0 +1,132 @@
...
+  const serverTimingStr = getServerTimingInfo(serverTiming)
+  if (serverTimingStr) {
+    respContext.headers = {
+      'server-timing': serverTimingStr
+    }
+  }
+  return respContext
+}
+
+function getResourceContext(data) {
```
axw: Is there any particular reason why we shouldn't or can't capture destination info for resources as well?

packages/rum-core/src/common/context.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+}
+
+function getExternalContext(data) {
+  const { url, method, target, response } = data
+  const parsedUrl = new Url(url)
+
+  const { href, port, protocol, hostname, origin } = parsedUrl
+  const isDefaultPort = port === '80' || port === '443'
```
axw: this should really be dependent on the protocol/scheme

packages/rum-core/src/common/url.js
```diff
@@ -136,6 +136,20 @@ class Url {
 
     this.protocol = protocol || location.protocol
 
+    /**
+     * Construct port and hostname from host and assign defaults based
+     * on http and https protocols
+     */
+    if (/:\d+$/.test(this.host)) {
+      const value = this.host.split(':')
```
axw: perhaps use capturing groups in the regex? that might also simplify removing square brackets from IPv6 addresses

packages/rum-core/src/common/context.js
vigneshshanmugam: They would be pointing to CDN or some other third part urls. Thought they are not useful for SIEM since it cannot be an attack vector. WDYT?

packages/rum-core/test/common/context.spec.js
```diff
@@ -0,0 +1,170 @@
...
+        url,
+        status_code: 200
+      },
+      destination: {
+        service: {
+          name: 'http://[::1]',
+          resource: '[::1]:80',
+          type: 'external'
+        },
+        address: '[::1]',
```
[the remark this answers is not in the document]
vigneshshanmugam: Any reason why we need special treatment for address alone when compared to service.name and resource?

packages/rum-core/src/common/context.js
vigneshshanmugam: That part is taken care here right? https://github.com/elastic/apm-agent-rum-js/blob/b0d89b4ef23e8d3ece2dcbd30f383a7dcdf3c4e2/packages/rum-core/src/common/url.js#L149 Am i missing something?

axw: You're _setting_ the default port there, but (and I know this is pathological) what if the URI explicitly specifies the port, like `https://host:80`? Port 80 is not the default port for https.

packages/rum-core/test/common/context.spec.js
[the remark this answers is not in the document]
axw: Yes :) From https://github.com/elastic/apm/issues/115#issuecomment-562452269:

> Note: for IPv6 addresses, surrounding the address with square brackets is only relevant where you have (or may have) the port alongside, e.g. [::1]:80, http://[::1]:80, http://[::1] (square brackets still required because the port is optional). The square brackets are there to disambiguate the IPv6 address, which contains colons, from the proceeding port, which is separated by a colon.
>
> When recording an IPv6 address separately from a port, as we are in the case of context.destination.address, then it should be recorded in its canonical form: without square brackets, e.g. ::1.

In `destination.service.name` the port may is optionally present (so brackets are required either way to disambiguate), and in `destination.service.resource` the port is always present.

packages/rum-core/src/common/context.js
axw: > They would be pointing to CDN or some other third part urls. Thought they are not useful for SIEM since it cannot be an attack vector. WDYT?

SIEM-wise, I'm not sure. Seems like it could be interesting to know which third-party services users are being sent to, in case that changes unexpectedly.

We do want to capture the `destination.service.*`, so we can display CDNs and such in the service map.

packages/rum-core/test/common/context.spec.js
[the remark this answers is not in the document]
vigneshshanmugam: Thanks for the explanation Andrew 👍

packages/rum-core/src/common/context.js
vigneshshanmugam: Cool, will incorporate the fields for those as well.

packages/rum-core/src/common/context.js
vigneshshanmugam: if there is a explicit mention of port, it wont hit this `else` block right?  we would be using 80 for the port.

axw: > if there is a explicit mention of port, it wont hit this else block right? we would be using 80 for the port.

In the code you linked to, yeah. But I'm talking about `function getExternalContext`, which is only looking at the resulting port number (which may be the default, or may have been explicitly specified) without also looking at the scheme, to decide that it's a default port.

In case it's still not clear what I'm on about, add case to `'should add external span context with all fields'` like this:

```js
url = 'https://[::1]:80/'
span = createSpanWithData(url)
```

IIANM, your code will set `destination.service.name` to `https://[::1]`, but it should be `https://[::1]:80` as 80 is not the default port for https.

packages/rum-core/src/common/context.js
jahtalab: I think they might even be relevant for SIEM, e.g loading JS file from an unknown location.

packages/rum-core/src/common/context.js
vigneshshanmugam: oops, good point. Sorry for the confusion and thanks for explaining. Will fix it.

packages/rum-core/src/common/url.js
vigneshshanmugam: Decided to not use capture groups and went with explicit ipv6 check to keep things simpler.

packages/rum-core/test/common/context.spec.js
```diff
@@ -0,0 +1,207 @@
...
+          name: 'http://[::1]',
+          resource: '[::1]:80',
+          type: 'external'
+        },
+        address: '::1',
+        port: '80'
+      }
+    })
+
+    url = 'https://[::1]:80/'
```
axw: might be worth adding a comment, so it's clear this wasn't a typo :)

packages/rum-core/src/common/context.js
```diff
@@ -0,0 +1,166 @@
...
+  let address = hostname
+  if (ipv6Hostname) {
+    address = hostname.slice(1, -1)
+  }
+
+  return {
+    service: {
+      name: !isDefaultSchemeAndPort ? origin : protocol + '//' + hostname,
+      resource: hostname + ':' + port,
+      type: 'external'
```
axw: this should be 'resource' for resource-type spans

vigneshshanmugam: I actually thought about this and left it to `external` since we make a HTTP request anyways for fetching the resource. wouldn't this confuse users when using service maps feature that some are resource and some are external?

packages/rum-core/test/common/context.spec.js
vigneshshanmugam: sure, will do 👍

packages/rum-core/test/performance-monitoring/transaction.spec.js
```diff
@@ -76,13 +76,6 @@ describe('transaction.Transaction', function() {
     tr.end()
   })
 
-  it('should store context.page.url', function() {
-    var tr = new Transaction('/', 'transaction')
-    tr.detectFinish()
-    var location = tr.context.page.url
-    expect(location).toBe(window.location.href)
-  })
```
jahtalab: Isn't this test still relevant since we add the context in transaction service?

packages/rum-core/src/common/url.js
```diff
@@ -136,6 +136,24 @@ class Url {
...
+      this.hostname = value.join(':')
+    } else {
+      this.hostname = this.host
+      this.port =
+        this.protocol === 'http:'
+          ? '80'
+          : this.protocol === 'https:'
+          ? '443'
+          : ''
+    }
```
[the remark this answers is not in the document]
vigneshshanmugam: URL native instrumentation does not set `port` for default ports and it makes there.But for us, It would make out service and address construction rely on additional checks.  

IPv6 parsing is on par with browser URL implementation and I have added comment in context.js on reasoning 
Screenshot 2019-12-10 at 12 18 06

packages/rum-core/test/performance-monitoring/transaction.spec.js
vigneshshanmugam: Let me revert this. Thought its not super useful.

packages/rum-core/src/common/context.js
```diff
@@ -0,0 +1,166 @@
...
+   *
+   * since address and port are recorded separately, we are recording the
+   * info in canonical form without square brackets
+   *
+   * IPv6 check is done here instead of in the URL parser to keep the URL parser
+   * functionality minimal and it also makes things easier for service construction
+   */
+  const ipv6Hostname =
+    hostname.charCodeAt(0) === LEFT_SQUARE_BRACKET &&
+    hostname.charCodeAt(hostname.length - 1) === RIGHT_SQUARE_BRACKET
```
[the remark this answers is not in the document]
vigneshshanmugam: I added the reasoning already on why its in this place, One of the reason is 
1. Browser URL parser has `hostname` with square brackets, so we do the same
2. We would have to add square brackets  `service.name` and `service.resource` since the format requires that except `address`.
3. Unnecessary parsing logic/overhead inside in the URL parser. 

We can move it inside the parser later on if there is absolute need.

packages/rum-core/src/common/url.js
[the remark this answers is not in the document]
jahtalab: Regarding the default port, we can set it explicitly for the service map.

jahtalab:
@vigneshshanmugam , Generally I think the URL class should emulate the native API as close as possible. The main reason here is to give users consistent experience across platforms, specially since we use URL class in a few different places and not just for service map. Another reason for the close replication of native behaviour is that, we should be able to remove our URL class and use the native API without any changes in the end result. We don't have to implement the entire specification (e.g. for IPv6) but the parts that we implement should be similar to the native API.

IMO our custom needs for url parsing should live outside the URL class.

packages/rum-core/test/performance-monitoring/transaction.spec.js
vigneshshanmugam: We handle it as part of the transaction service end transaction. I am removing this from here since there are already tests  in transaction service and also for context related tests.

vigneshshanmugam:
@jahtalab I agree with you on some of the points and decided to align with Browser URL implementation on `port`,`hostname` and `host` parsing and pushed the changes. 

hostname for ipv6 is a different scenario and we will still keep it as part of `context.js` since its as per the spec implementation.

packages/rum-core/src/common/context.js
vigneshshanmugam: Changed to `resource` for now. I also added the comment in the master issue if its okay to keep it as `resource`

axw: Yeah, potentially should be the same regardless of the reason for the request. I saw your question on the main issue, so let's see what Ron thinks.

packages/rum-core/src/common/context.js
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  return {
+    service: {
+      name: protocol + '//' + host,
+      resource: hostname + ':' + portNumber,
+      type
+    },
+    address,
+    port: portNumber
```
axw: Sorry, I missed this detail before: the port field should be an integer type. At the moment this is a string. If the port is unknown (unspecified, and default is unknown), the field should be omitted.

vigneshshanmugam: Ahh missed the intake API type. Will fix it. Thanks

packages/rum-core/src/common/url.js
```diff
@@ -137,21 +151,23 @@ class Url {
...
-     * Construct port and hostname from host and assign defaults based
-     * on http and https protocols
+     * Construct port and hostname from host
+     *
+     * Port numbers are not added for default ports of a given protocol
+     * and hostname would match host when port is not present
      */
+    this.hostname = this.host
+    this.port = ''
     if (/:\d+$/.test(this.host)) {
```
jahtalab: Wouldn't it be simpler to do what is done here: https://github.com/unshiftio/url-parse/blob/master/index.js#L41, i.e. adding a new rule for the port?

packages/rum-core/test/performance-monitoring/transaction.spec.js
jahtalab: Do we have a test that checks the context to have the `window.location.href` value?  which one are you referring to ?

packages/rum-core/src/common/url.js
vigneshshanmugam: It's actually slower to do that way, since the for loop has to check for each instruction whether it is a Regex thats why we don't use that.

vigneshshanmugam:
@jahtalab We don't 100% adhere to the URL spec because of how query param and other relative URLs are passed. We can do that in a separate PR when we adhere to the SPEC in other fields as well.

packages/rum-core/src/common/url.js
jahtalab: I don't think the perf impact is actually significant since the loop goes over a constant array of only 6 items. I'm ok with keeping it this way though.

vigneshshanmugam:
We can do that later on, Let's not block this huge PR for those cases. We can always add tests and enhance it later on.

jahtalab:
We had call and decide to merge this one and add the tests in a separate PR, we have task in the original issue.

PR #517, opened 2019-12-06
feat: capture http-request transactions

### LINKED ISSUE #60: Automatically create transactions for AJAX requests
Currently, only those XHR Requests, which are within a manually created transaction are recorded.

In some frameworks, like JSF you have very little control over the JavaScript so manually creating transactions for button clicks which initiate AJAX requests is not viable. Even if the developer has direct control over the ajax requests, instrumenting a big application would be a lot of effort.

I propose that the RUM agent should automatically create a transaction for XHR Requests in case there is no currently active one. This would make the auto-instrumentation cover more of the application without the user having to programmatically add tracing logic to their application. That, in turn, makes it easier to adopt the RUM agent.

In terms of distributed tracing, it would also bring visibility into where in the frontend a backend request was initiated and how much time the network interaction took.

jahtalab:
Fixes #60

### DIFF
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -40,7 +40,8 @@ import {
   AFTER_EVENT,
   FETCH,
   HISTORY,
-  XMLHTTPREQUEST
+  XMLHTTPREQUEST,
+  HTTP_REQUEST_TYPE
 } from '../common/constants'
 import {
   truncateModel,
@@ -62,6 +63,7 @@ class PerformanceMonitoring {
      * We need to run this event listener after all of user-registered listener,
      * since this event listener adds the transaction to the queue to be send to APM Server.
      */
+
     this._configService.events.observe(TRANSACTION_END + AFTER_EVENT, tr => {
       const payload = this.createTransactionPayload(tr)
       if (payload) {
@@ -122,6 +124,13 @@ class PerformanceMonitoring {
         (requestUrl.relative
           ? requestUrl.path
           : stripQueryStringFromUrl(requestUrl.href))
+
+      if (!transactionService.getCurrentTransaction()) {
+        transactionService.startTransaction(spanName, HTTP_REQUEST_TYPE, {
+          managed: true
+        })
+      }
+
       const span = transactionService.startSpan(spanName, 'external.http')
       const taskId = transactionService.addTask()
 
--- a/packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
+++ b/packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
@@ -679,21 +679,26 @@ describe('PerformanceMonitoring', function() {
     ])
   })
 
-  it('should create http-request transaction if no current transaction exist', done => {
-    const transactionService = serviceFactory.getService('TransactionService')
+  function createXHRTask(method, url) {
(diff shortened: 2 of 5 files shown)


packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -122,6 +124,13 @@ class PerformanceMonitoring {
         (requestUrl.relative
           ? requestUrl.path
           : stripQueryStringFromUrl(requestUrl.href))
+
+      if (!transactionService.getCurrentTransaction()) {
+        transactionService.startTransaction(spanName, HTTP_REQUEST_TYPE, {
```
vigneshshanmugam: Are we using the same spanName as transaction name for consistency or is there any reasoning behind this?

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -678,4 +678,46 @@ describe('PerformanceMonitoring', function() {
...
+    expect(transactionService.startTransaction).toHaveBeenCalledWith(
+      'GET /',
+      'http-request',
+      jasmine.objectContaining({ managed: true })
+    )
+
+    var tr = transactionService.getCurrentTransaction()
+    expect(tr).toBeDefined()
+
+    setTimeout(() => {
```
vigneshshanmugam: Instead of depending on a timeout, cleaner approach would be to use the `transaction:end` event to do the check for the payload errors.

jahtalab: @vigneshshanmugam , the setTimeout here is to simulate the latency of a XHR. It's not for the transaction to end.

vigneshshanmugam: Ahh, Got you. cant we do req.send and listen for the `transaction:end` event instead of timeouts. 

Reason why i am so against timeouts is that, its not reliable and would hang the test even if there is a bug in a depedency.

jahtalab: IMO, it's fine to use setTimeout as long as we properly call the `done` function. Not sure why just using setTimeout has to make the tests unreliable.

vigneshshanmugam: Just fake one of the timers somewhere(can happen in dependencies or in our code) and we will never see the tests calling done. we can leave it to `setTimeout`,not a huge deal.

jahtalab:
@vigneshshanmugam , Tests are passing now, please have another look!

packages/rum/test/e2e/manual-timing/app.js
```diff
@@ -33,7 +33,11 @@ const apm = createApmBase({
   operationMode: 'manual'
 })
 
-const transaction = apm.startTransaction('transaction-name', 'transaction-type')
+const transaction = apm.startTransaction(
+  'transaction-name',
+  'transaction-type',
+  { managed: true }
```
vigneshshanmugam: The test was to check for manual type right? can you explain why we need this here now?

jahtalab: I think we forgot to update this test when we introduced the managed option, if you look at the rest of the code, we actually expect the that the xhrs to be captured by this transaction. So the test is for manual but managed transactions. we might need another test for manual but unmanaged transactions as well, it's out of the scope of this PR though.

PR #521, opened 2019-12-10
perf(rum-core): move unexposed configs inside constants to reduce bundlesize

### LINKED ISSUE #502: move unexposed config service options to constants
+ Subtask of elastic/apm-agent-rum-js#159 
+ Config service currently holds a large amount of configs that are not documented which could be potentially moved to constants to reduce the bundle size. 

- Minification cannot be run safely on objects and by moving it to constants the minfier can easily mangle the variable names exported from the constants across multiple files

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#502 
+ Also documented `ignoreTransactions` which was not before.


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -147,7 +144,7 @@ class TransactionService {
 
     tr.captureTimings = true
 
-    if (type === PAGE_LOAD) {
+    if (tr.type === PAGE_LOAD) {
```
vigneshshanmugam: There was a bug associated with how we redefine when we start managed transaction in frameworks and it did not have browserResponsive check set to false for page load transaction.

vigneshshanmugam:
@jahtalab Can you have a look.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -382,8 +367,9 @@ class PerformanceMonitoring {
           lastSpan.subType === span.subType &&
           lastSpan.action === span.action &&
           lastSpan.name === span.name &&
-          span.duration() / transDuration < threshold &&
-          (span._start - lastSpan._end) / transDuration < threshold
+          span.duration() / transDuration < SIMILAR_SPAN_THRESHOLD_PERCENTAGE &&
+          (span._start - lastSpan._end) / transDuration <
+            SIMILAR_SPAN_THRESHOLD_PERCENTAGE
```
jahtalab: I don't see a lot of benefit in using the constant here, instead of in the caller function. 
Furthermore, with a function argument we can test this function more thoroughly, i.e. with different ratios.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -405,17 +391,12 @@ class PerformanceMonitoring {
...
-    var duration = transaction.duration()
-    var expectedCount = Math.floor(duration / interval)
-    var wasBrowserResponsive = counter + buffer >= expectedCount
+  checkBrowserResponsiveness(transaction) {
+    const counter = transaction.browserResponsivenessCounter
+    const duration = transaction.duration()
+    const expectedCount = Math.floor(duration / BROWSER_RESPONSIVENESS_INTERVAL)
 
-    return wasBrowserResponsive
+    return counter + BROWSER_RESPONSIVENESS_BUFFER >= expectedCount
```
jahtalab: The same here for function arguments.
Image if we want to test this function with different intervals or different buffers!

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: We don't change the transaction type when we redefine so I'm not sure if I follow.

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -172,29 +172,28 @@ describe('PerformanceMonitoring', function() {
...
     expect(resp).toBe(true)
 
     tr._end = 1601
     tr.browserResponsivenessCounter = 2
-    resp = performanceMonitoring.checkBrowserResponsiveness(tr, 500, 2)
+    resp = performanceMonitoring.checkBrowserResponsiveness(tr)
     expect(resp).toBe(true)
 
-    tr._end = 3001
+    tr._end = 3501
```
jahtalab: If we didn't change the logic for `checkBrowserResponsiveness` this value should not be changed then.

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -241,35 +240,30 @@ describe('PerformanceMonitoring', function() {
...
       ' duration:',
-      3000,
+      3500,
       ' browserResponsivenessCounter:',
-      3,
-      'interval:',
-      500
+      3
     )
   })
```
jahtalab: Same as above for this test

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: when we call redefine the type would be undefined which would skip this if check which is the source of bug.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: Basically to avoid the validation part inside functions and reducing bundle size bytes. We can add the args back since it would be interesting to test as you mentioned, but will remove the validation since we know its a constant and its a super generic function to test for all possible cases.

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
vigneshshanmugam: This test depends on configs which are removed, so it does not apply here.

PR #528, opened 2019-12-12
chore(deps): upgrade build tools deps and fix security warnings

### LINKED ISSUE #527: fix security warnings and use terser plugin for minification
+ https://github.com/webpack-contrib/uglifyjs-webpack-plugin is deprecated and we have to migrate to terser
+ security warnings is raised from a dependency of uglify-js-plugin. 
+ also upgrade build dependencies like babel and webpack

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#527 
+ Took the leverage of refactoring our webpack build configs and get it from single central place for release, test and bundling.


jahtalab:
Had a call with @vigneshshanmugam and we decided to add a new function for the distribution bundles and refactor `getWebpackConfig` to reuse the common parts in the new function.

packages/rum/webpack.config.js
```diff
@@ -41,76 +40,29 @@ const devConfig = entry => ({
...
-        options: getBabelConfig(BUNDLE_TYPES.BROWSER_DEV)
-      }
-    ]
-  },
-  plugins: [
-    new EnvironmentPlugin({
-      NODE_ENV: 'development'
-    })
-  ]
+  ...getWebpackReleaseConfig(BUNDLE_TYPES.BROWSER_DEV)
```
jahtalab: I think it's better to put this at the top, so it will be overwritten by our custom config in this file! 
I can see there are no conflicting configs at the moment, but to make sure we get the order right.

packages/rum/webpack.config.js
```diff
@@ -41,76 +40,29 @@ const devConfig = entry => ({
...
-    }),
-    new BundleAnalyzerPlugin({
-      analyzerMode: 'static',
-      reportFilename: join(REPORTS_DIR, `${name}-report.html`),
-      generateStatsFile: true,
-      statsFilename: join(REPORTS_DIR, `${name}-stats.json`),
-      openAnalyzer: false
-    })
-  ]
+  ...getWebpackReleaseConfig(BUNDLE_TYPES.BROWSER_PROD, name)
```
jahtalab: The same as above!

dev-utils/build.js
```diff
@@ -139,67 +141,116 @@ function getWebpackEnv(env = 'development') {
...
+  const mode = isEnvProduction ? 'production' : 'development'
+
+  return {
+    devtool: isEnvProduction ? 'source-map' : false,
+    mode,
     stats: {
       colors: true,
-      warnings: false
+      assets: true,
+      modules: false
```
jahtalab: Wasn't `modules:false` the reason we didn't get the right bundle analyzer report?

vigneshshanmugam: Nope, it was just hidden inside the flag on UI. modules: false will disable the file level comment that we get on stdout.

Issue #530, opened 2019-12-13
TransactionService refactoring


jahtalab:
Some of the logic in TransactionService is repeat in multiple places (e.g. `createTransaction`, `createPerfOptions`), we need to do some refactoring to fix these.

PR #531, opened 2019-12-13
refactor(rum): small optimizations on the transaction service

### LINKED ISSUE #522: easy optimisation wins on Transaction service
+ browserResponsive interval should not be started by default as soon as the transaction is created since we dont want it to run anyways on `page-load` transactions
+ options can be created only when there is an active transaction https://github.com/elastic/apm-agent-rum-js/blob/4732c73bd8513621170f7d647d2d10ffa09cbe57/packages/rum-core/src/performance-monitoring/transaction-service.js#L42-L44

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#522 
+ Options would be created only when there is no transaction exists
+ transaction responsiveness counter should not get started for page-load transactions. We could also get rid of this logic, but I am going to leave it to the other refactor elastic/apm-agent-rum-js#530

### DIFF
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -44,17 +44,16 @@ class TransactionService {
   }
 
   ensureCurrentTransaction(options) {
+    const tr = this.getCurrentTransaction()
+    if (tr) {
+      return tr
+    }
     if (!options) {
       options = this.createOptions()
     }
-    var tr = this.getCurrentTransaction()
-    if (tr) {
-      return tr
-    } else {
-      options.canReuse = true
-      options.managed = true
-      return this.createTransaction(undefined, undefined, options)
-    }
+    options.canReuse = true
+    options.managed = true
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -68,9 +67,9 @@ class TransactionService {
   }
 
   createTransaction(name, type, options) {
-    var tr = new Transaction(name, type, options)
+    const tr = new Transaction(name, type, options)
     this.setCurrentTransaction(tr)
-    if (options.checkBrowserResponsiveness) {
+    if (type !== PAGE_LOAD && options.checkBrowserResponsiveness) {
```
jahtalab: I don't think we should add a check for page load here. we should make sure that we set `options. checkBrowserResponsiveness` on page loads transactions.

vigneshshanmugam: Not really, we should not start the counter from the first. right now we do start the timer which starts a timer for no reason and then disable it on post-processing. We should handle it here and do not even care on the end.

PR #533, opened 2019-12-16
fix(rum-core): create temporary transaction on startSpan

jahtalab:
This PR refactors transaction service to consolidate similar code paths. Furthermore, it also adds the temporary transaction on startSpan and addTask.

Fixes: #530


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -292,7 +286,14 @@ class TransactionService {
   }
 
   startSpan(name, type, options) {
-    const tr = this.ensureCurrentTransaction()
+    const tr = this.ensureCurrentTransaction(
+      undefined,
+      TEMPORARY_TYPE,
+      this.createOptions({
```
vigneshshanmugam: I would move these logic inside the function and get rid of passing around parameters here. WDYT? Would reduce some uncessary code for `startSpan` and addTask.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -43,18 +44,18 @@ class TransactionService {
     this.currentTransaction = undefined
   }
 
-  ensureCurrentTransaction(options) {
-    if (!options) {
-      options = this.createOptions()
-    }
-    var tr = this.getCurrentTransaction()
+  ensureCurrentTransaction(name, type, options) {
```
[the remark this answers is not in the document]
jahtalab: I think in using "ensure" is better here, but to be honest, I was looking for some shorter name, but I couldn't think of any. Let me know if you have suggestions.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: `ensureCurrentTransaction` replaces `createTransaction` which we had before. I think this way it makes `ensureCurrentTransaction` more useful in different scenarios where we don't want to create options. see the tests for examples.

vigneshshanmugam: I was also saying this because, it would reduce the code size. If we can come up with a way to make both work, it would be better.

packages/rum-core/src/performance-monitoring/transaction-service.js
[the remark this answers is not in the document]
vigneshshanmugam: Hmmm, cant really think of any as well.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -43,18 +44,18 @@ class TransactionService {
...
+    let tr = this.getCurrentTransaction()
     if (tr) {
       return tr
     } else {
-      options.canReuse = true
-      options.managed = true
-      return this.createTransaction(undefined, undefined, options)
+      tr = new Transaction(name, type, options)
+      this.setCurrentTransaction(tr)
+      if (options && options.checkBrowserResponsiveness) {
```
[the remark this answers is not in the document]
jahtalab: The code https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/src/performance-monitoring/performance-monitoring.js#L265 is for whether we should discard the transaction based on responsiveness or not, it is still necessary.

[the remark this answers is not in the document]
vigneshshanmugam: It is indeed, but we don't need to check for config flag since we set the flag on tr.options manually based on the config flag.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -67,24 +66,28 @@ class TransactionService {
...
+          let tr = this.getCurrentTransaction()
+          if (tr) {
+            tr.browserResponsivenessCounter++
+          } else {
+            clearRespInterval()
+          }
+        }, BROWSER_RESPONSIVENESS_INTERVAL)
       }
-    }, BROWSER_RESPONSIVENESS_INTERVAL)
+    } else if (typeof this.respIntervalId !== 'undefined') {
```
vigneshshanmugam: This else if block seems like a dead code to me. Am i missing something?

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -68,15 +66,28 @@ class TransactionService {
...
 
-  startCounter(transaction) {
-    transaction.browserResponsivenessCounter = 0
-    const id = setInterval(function() {
-      if (transaction.ended) {
-        window.clearInterval(id)
-      } else {
-        transaction.browserResponsivenessCounter++
+  ensureRespInterval() {
+    let currentTr = this.getCurrentTransaction()
```
vigneshshanmugam: This is double work, When we call `ensureRespInterval`, we have already created a current transaction, So instead of again getting the current transaction, we can simply pass that flag in args. 

```js
ensureRespInterval(checkBrowserResponsiveness) {
  if (!checkBrowserResponsiveness) {
     return
   }
 // code here
}

```

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -68,15 +66,28 @@ class TransactionService {
...
+
+    const clearRespInterval = () => {
+      window.clearInterval(this.respIntervalId)
```
vigneshshanmugam: nit: use window on both `setInterval` and `clearInterval` or can be removed from both places.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -258,11 +258,7 @@ class PerformanceMonitoring {
       tr.resetSpans()
     }
 
-    const checkBrowserResponsiveness = this._configService.get(
-      'checkBrowserResponsiveness'
-    )
-
-    if (checkBrowserResponsiveness && tr.options.checkBrowserResponsiveness) {
+    if (tr.options.checkBrowserResponsiveness) {
```
vigneshshanmugam: Instead of this, we can use the counter itself to check whether the transaction has passed an interval. 

`if (tr. browserResponsivenessCounter > 0)`

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -153,6 +164,8 @@ class TransactionService {
       }
     }
 
+    this.ensureRespInterval()
```
vigneshshanmugam: It seems like instead of calling this function always and checking for `checkBrowserResponsiveness`, we can just move this inside else block of if `(tr.type === PAGE_LOAD)` and get rid of some code from this function.

jahtalab:
@vigneshshanmugam , the `browserResponsivenessCounter` is only updated on the 500ms intervals, we can't rely on that since even before the counter is updated there could be calls to `startTransaction` which in turn might redefine transaction. we already have plans to redefine the type based on priority which means the `checkBrowserResponsiveness` might also be changed!

packages/rum-core/src/performance-monitoring/performance-monitoring.js
jahtalab: We can't rely on `browserResponsivenessCounter ` since it's only updated in 500ms intervals. see my main comment

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: IMO, we should keep the special cases for the transaction type in one place as much as possible. `ensureRespInterval` function only needs to know about `checkBrowserResponsiveness` and that's all the information it needs. if we have a check in that function we need to remember to update that code path as well every time we change something.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: Since the `checkBrowserResponsiveness` can change (e.g. via redefined) we might end up in this code path.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: Let's move `checkBrowserResponsiveness` to the caller function!

vigneshshanmugam:
The changed would be incorporated on the next PR, so decided to merge this one.

PR #537, opened 2019-12-17
feat(rum): add typescript typings

### LINKED ISSUE #70: Provide typescript definition
There are multiple benefits of providing typings for the users. First generally IDEs have support for typing files and can provide auto-completion even in JS code. It also helps users identify and fix issues related to the API changes.

- [x] Documentation
- [x] Tests to check the definition with actual typescript compiler

Postponed:
- [ ] Run the tests in Jasmine

vigneshshanmugam:
+ related: elastic/apm-agent-rum-js#70 
+ Added tests for the types which would error out if the API is used improperly.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -47,7 +47,8 @@ module.exports = {
     ],
     'react/prop-types': 0,
     '@typescript-eslint/explicit-function-return-type': 0,
-    '@typescript-eslint/no-explicit-any': 0
+    '@typescript-eslint/no-explicit-any': 0,
+    '@typescript-eslint/member-delimiter-style': 0
   },
   settings: {
     react: {
--- a/dev-utils/karma.js
+++ b/dev-utils/karma.js
@@ -35,7 +35,7 @@ const baseLaunchers = getBrowserList().map(launcher => ({
   ...launcher
 }))
 const specPattern =
-  'test/{*.spec.+(js|ts),!(e2e|integration|node|bundle)/*.spec.+(js|ts)}'
+  'test/{*.spec.+(js|ts),!(e2e|integration|node|bundle|types)/*.spec.+(js|ts)}'
 const { tunnelIdentifier } = getSauceConnectOptions()
 
 /**
--- a/docs/advanced-topics.asciidoc
+++ b/docs/advanced-topics.asciidoc
@@ -10,8 +10,8 @@
 === Using with TypeScript
 
 RUM agent publishes the type definitions for the `@elastic/apm-rum` package via `types` property in `package.json`.
-If you are importing the package on a Typescript codebase, you will get automatic code completion, hover info,
-and method signature information on supported platforms (Visual Studio Code, etc.).
+If you are importing the package on a TypeScript codebase, you will get automatic code completion, hover info
+and method signature information on supported platforms (Visual Studio Code, etc)
 
 
 include::./custom-transactions.asciidoc[Custom Transactions]
--- a/docs/advanced-topics.asciidoc
+++ b/docs/advanced-topics.asciidoc
@@ -3,15 +3,15 @@
 
 * <<custom-transactions, Custom Transactions>>
 * <<custom-transaction-name, Initial page load transaction names>>
+* <<typescript>>
 
 
-[float]
 [[typescript]]
-==== Using with TypeScript
+=== Using with TypeScript
 
 RUM agent publishes the type definitions for the `@elastic/apm-rum` package via `types` property in `package.json`.
-If you are importing the package on a Typescript codebase, you will get automatic code completion, hover info
-and method signature information on supported platforms (Visual Studio Code, etc)
+If you are importing the package on a Typescript codebase, you will get automatic code completion, hover info,
+and method signature information on supported platforms (Visual Studio Code, etc.).
 
 
 include::./custom-transactions.asciidoc[Custom Transactions]
--- a/docs/advanced-topics.asciidoc
+++ b/docs/advanced-topics.asciidoc
@@ -4,5 +4,15 @@
 * <<custom-transactions, Custom Transactions>>
 * <<custom-transaction-name, Initial page load transaction names>>
 
+
+[float]
+[[typescript]]
+==== Using with TypeScript
+
+RUM agent publishes the type definitions for the `@elastic/apm-rum` package via `types` property in `package.json`.
+If you are importing the package on a Typescript codebase, you will get automatic code completion, hover info
+and method signature information on supported platforms (Visual Studio Code, etc)
+
+
 include::./custom-transactions.asciidoc[Custom Transactions]
 include::./custom-transaction-name.asciidoc[Initial page load transaction names]
--- a/packages/rum/src/index.d.ts
+++ b/packages/rum/src/index.d.ts
@@ -37,13 +37,13 @@ declare module '@elastic/apm-rum' {
     addLabels(labels: Labels): void
     setInitialPageLoadName(name: string): void
     startTransaction(
-      name?: string,
-      type?: string,
+      name?: string | null,
+      type?: string | null,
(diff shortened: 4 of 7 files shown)


vigneshshanmugam:
Did not add the docs part yet, Not sure where to add it though. Thoughts? @jahtalab @bmorelli25

bmorelli25:
I'm not 100% sure. I might need to see what this looks like before being of any help. Does it make sense to add another subheading under the API reference topic?

dev-utils/karma.js
```diff
@@ -35,7 +35,7 @@ const baseLaunchers = getBrowserList().map(launcher => ({
   ...launcher
 }))
 const specPattern =
-  'test/{*.spec.+(js|ts),!(e2e|integration|node|bundle)/*.spec.+(js|ts)}'
+  'test/{*.spec.+(js|ts),!(e2e|integration|node|bundle|types)/*.spec.+(js|ts)}'
```
jahtalab: Why the `types` is necessary here?

packages/rum/test/types/index.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  })
+  span.addContext({
+    test: {
+      foo: '1',
+      bar: 2
+    }
+  })
+  span.end()
+  span.duration()
+}
```
jahtalab: Shouldn't these be jasmine style tests, e.g. `it('should init ApmBase', ()=>{})`

.eslintrc.js
```diff
@@ -47,7 +47,8 @@ module.exports = {
     ],
     'react/prop-types': 0,
     '@typescript-eslint/explicit-function-return-type': 0,
-    '@typescript-eslint/no-explicit-any': 0
+    '@typescript-eslint/no-explicit-any': 0,
+    '@typescript-eslint/member-delimiter-style': 0
```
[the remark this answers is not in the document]
vigneshshanmugam: We cant, 0, 1, 2 or none, warn, error are eslint rules. 0 seems simpler.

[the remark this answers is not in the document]
vigneshshanmugam: We cant, 0, 1, 2 or none, warn, error are eslint rules. 0 seems simpler.

packages/rum/test/types/index.spec.ts
vigneshshanmugam: No, These are types tests. If you modify any API here, typescript will throw error and test would fail.

dev-utils/karma.js
vigneshshanmugam: To ignore from running in karma as part of unit tests.

vigneshshanmugam:
@jahtalab 
> We should only provide typings for the methods and configs that we have documented, otherwise, we would have to commit to not changing those if we add them here.

Good point, will ignore configs that are not in docs. 

> I think it is useful to have some comments for each of the methods that describes what it does and it's params, the description would also show up in IDEs autocompletion. My only concern is that it duplicates our docs and we have fix in multiple places, It would be great if we could automate this process but it might be out of the scope of this PR, WDYT?

If we need this, we have to do JSDOC style annotations for each method exposed in the ApmBase and also on required classes which would be inferred by typescript and included in the types. We can do this separately as it would require a lot of work. 

> In regards to documenting this, I think adding Typescript to advanced topics might be a good option.

Sounds good to me as well.

vigneshshanmugam:
@bmorelli25 I added it aspart of Advanced topics as suggested by Hamid since it seems better fit.

docs/advanced-topics.asciidoc
```diff
@@ -4,5 +4,15 @@
...
 * <<custom-transaction-name, Initial page load transaction names>>
 
+
+[float]
+[[typescript]]
+==== Using with TypeScript
+
+RUM agent publishes the type definitions for the `@elastic/apm-rum` package via `types` property in `package.json`.
+If you are importing the package on a Typescript codebase, you will get automatic code completion, hover info
+and method signature information on supported platforms (Visual Studio Code, etc)
```
bmorelli25:
```suggestion
and method signature information on supported platforms (Visual Studio Code, etc.).
```

docs/advanced-topics.asciidoc
```diff
@@ -4,5 +4,15 @@
 * <<custom-transactions, Custom Transactions>>
 * <<custom-transaction-name, Initial page load transaction names>>
 
+
+[float]
+[[typescript]]
+==== Using with TypeScript
+
+RUM agent publishes the type definitions for the `@elastic/apm-rum` package via `types` property in `package.json`.
+If you are importing the package on a Typescript codebase, you will get automatic code completion, hover info
```
bmorelli25: Oxford comma 😛 
```suggestion
If you are importing the package on a Typescript codebase, you will get automatic code completion, hover info,
```

docs/advanced-topics.asciidoc
```diff
@@ -4,5 +4,15 @@
 * <<custom-transactions, Custom Transactions>>
 * <<custom-transaction-name, Initial page load transaction names>>
 
+
+[float]
```
bmorelli25:
```suggestion
```

docs/advanced-topics.asciidoc
```diff
@@ -4,5 +4,15 @@
...
+[[typescript]]
+==== Using with TypeScript
```
bmorelli25: This needs to be a level 3 heading without the float tag so it doesn't float to the overview
Screen Shot 2019-12-20 at 10 17 31 AM


```suggestion
=== Using with TypeScript
```

docs/advanced-topics.asciidoc
```diff
@@ -4,5 +4,15 @@
 * <<custom-transactions, Custom Transactions>>
 * <<custom-transaction-name, Initial page load transaction names>>
```
bmorelli25: Let's add this to the jump table as well

```suggestion
* <<custom-transaction-name, Initial page load transaction names>>
* <<typescript>>
```

vigneshshanmugam:
@bmorelli25 Thanks again for the suggestions 🎉

packages/rum/src/index.d.ts
```diff
@@ -0,0 +1,143 @@
...
+  const init: Init
+
+  class ApmBase {
+    init: Init
+    isEnabled(): boolean
+    observe(name: TransactionEvents, callback: (tr: Transaction) => void): void
+    config(options?: AgentConfigOptions): void
+    setUserContext(user: UserObject): void
+    setCustomContext(custom: object): void
+    addLabels(labels: Labels): void
```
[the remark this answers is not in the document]
vigneshshanmugam: Thought Ignoring is better since method i going to be removed soon anyways.

packages/rum/test/types/index.spec.ts
vigneshshanmugam: Hmmm, It would be harder to do it that way since this is a compile time test. We are testing to see errors if its thrown from typescript compiler. What you are asking for would require we run jasmine but before it we would have to capture and throw the errors from typescript to jasmine. It feels complicated set up to me. Happy to hear if you have any ideas.

jahtalab:
@vigneshshanmugam , Let's remove the `ApmBase.config()` since I think seeing that in autocomplete might be confusing given that calling config function after init is not supported for all config options. also let's mark other ones as undocumented or experimental.

packages/rum/test/types/index.spec.ts
jahtalab: Let's discuss this today. I think there might be some benefits in actually running the tests in jasmines. For example we can make the typescript pass by only changing the definition file but the actual ApmBase might no longer have that method or something along those lines.

jahtalab:
Had a call and we discussed:
- The benefits of actually running the tests with jasmine (Separate PR)
- We should add notes about the possibility of removing the undocumented methods.

PR #541, opened 2019-12-18
chore(rum-core): add native URL parser tests aganist polyfill

### LINKED ISSUE #534: Provide tests to check url.js polyfill against the native URL API
The url.js polyfill should be compatible with the native URL as much as possible. 
we should add tests that checks url.js outputs against the native API. 

In those case that we differ from the native API we should either fix our polyfill or document the difference in our tests.

vigneshshanmugam:
+ fix elastic/apm-agent-rum-js#534 
+ Added incompatible tests in separate level to identify if we can align with spec closer in future.

### DIFF
--- a/packages/rum-core/test/common/url.spec.js
+++ b/packages/rum-core/test/common/url.spec.js
@@ -271,10 +271,13 @@ describe('Url parser', function() {
 })
 
 /**
+ * Test native URL implementation only on supported platforms
  * In IE 11 - window.URL is not supported but it exists as an object
- * Run these tests only on supported platforms
+ * Microsoft Edge URL parsing for IPv6 is broken on `host` and `hostnames`
  */
-if (window.URL.toString().indexOf('native code') !== -1) {
+const userAgent = window.navigator.userAgent
+const browsersToIgnore = /(Trident|Edge)/
+if (userAgent && !browsersToIgnore.test(userAgent)) {
   describe('Native URL API Compatability', () => {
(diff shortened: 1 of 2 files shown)


packages/rum-core/test/common/url.spec.js
```diff
@@ -269,3 +269,73 @@ describe('Url parser', function() {
     )
   })
 })
+
+/**
+ * In IE 11 - window.URL is not supported but it exists as an object
+ * Run these tests only on supported platforms
+ */
+if (window.URL.toString().indexOf('native code') !== -1) {
```
vigneshshanmugam: We might have to skip this run on IE and Edge. Probably we can run only on chrome and ignore others.

vigneshshanmugam: Ignored the test right now on IE and Edge. Test run on rest of the browsers.

vigneshshanmugam:
@jahtalab In IE we cant do anything since its not supported. So we would have to skip running from there. And also we don't need to run on all browsers to know how we are differing from SPEC. 

For Edge, yes we can have that scenario separately for IPV6 address.

jahtalab:
@vigneshshanmugam , your own tests show that we DO need to test on all browsers to see if they are different from the spec 😉

vigneshshanmugam:
All of thee code review comments are addressed and the tests are adjusted to run on Microsoft Edge as well.

PR #553, opened 2020-01-01
feat(rum-vue): only hook router if provided

dsanders11:
Not all projects will use router, but the plugin currently assumes router is always provided.

### DIFF
--- a/packages/rum-vue/src/index.js
+++ b/packages/rum-vue/src/index.js
@@ -34,7 +34,13 @@ export const ApmVuePlugin = {
      */
     apm.init(config)
 
-    routeHooks(router, apm)
+    /**
+     * Hook router if provided
+     */
+    if (router) {
+      routeHooks(router, apm)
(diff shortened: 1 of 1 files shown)


dsanders11:
> But I would like to understand the use case if users would be interested to use the package apm-plugin-vue which would do simply intilaize the agent and make apm available to vue app instead of manually importing the apm-rum package and doing it by themselves?

I assumed this package was still a work in-progress with potential for more features in the future. There's further integration which can be done, such as hooking [`Vue.config.errorHandler`](https://vuejs.org/v2/api/#errorHandler), which as the documentation mentions, is something Sentry and Bugsnag do.

vigneshshanmugam:
Neat, thanks for the information and also links to the global error handler. As you have mentioned, the package is still in early stages of development and we can add more features. I will create a new  issue/feature request to add the error handler part. Thanks much

PR #554, opened 2020-01-02
chore: bump handlebars to fix security vulnerability

vigneshshanmugam:
+ fix for https://github.com/elastic/apm-agent-rum-js/pull/550


vigneshshanmugam:
They are already updated with handlebars >4.3.1. This was the issue with karma-coverage -> Istanbul -> handlebars which is an indirect dependency, so this solves for that problem. Thats why you don't see any changes in the `package.json`.

vigneshshanmugam:
Merging the PR as the review concerns are addressed on the PR.

PR #563, opened 2020-01-09
chore(benchmark): get predictable results on each run

vigneshshanmugam:
+ Since the disk cache was shared between runs, the results for `basic` and `heavy` scenarios were varying by a small margin which is not the case. Now, we run the benchmark on a incognito browser context which does not share cache for each session. Makes the results predictable.


vigneshshanmugam:
@jahtalab The difference shouldn't be huge. Just the jumpy behaviour of init/CPU time between basic and heavy should be fixed from now.

PR #565, opened 2020-01-10
chore: apmbase export types must reflect constructor properties

vigneshshanmugam:
+ Hope this the last , figured out this from this test file https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-angular/test/index.ts. 
+ Also matched the exported members with the one from JS to keep it clear. 

We dont use `apmBase` directly on most test so it was hard to figure out this case.

### DIFF
--- a/packages/rum/src/index.d.ts
+++ b/packages/rum/src/index.d.ts
@@ -28,6 +28,7 @@ declare module '@elastic/apm-rum' {
   const init: Init
 
   class ApmBase {
+    serviceFactory: ServiceFactory
     constructor(serviceFactory: ServiceFactory, disable: boolean)
     init: Init
     /**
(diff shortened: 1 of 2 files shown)


packages/rum/src/index.d.ts
```diff
@@ -53,8 +54,9 @@ declare module '@elastic/apm-rum' {
     captureError(error: Error | string): void
     addFilter(fn: FilterFn): void
   }
-  export { init, ApmBase as apm, ApmBase as apmBase, ApmBase }
+  const apmBase: ApmBase
```
jahtalab: I wonder why this wasn't caught in our tests. Before this commit, we exported `ApmBase as apm`, which would be the class itself instead of its instance and shouldn't have the exposed APIs that we're testing for.

vigneshshanmugam: Because we dont expect users to do `apm.serviceFactory` and I was about to catch this only because we were mocking ApmServer factory in our tests.

Issue #478, opened 2019-11-12
use performance observer to capture new metric types


vigneshshanmugam:
Most of the new metrics available in the browser is supported only via the new [Performance Observer API](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver)

We should ideally use that moving forward and use the old way for non supported browsers. 

```js
// old way
const entries = perf.getEntriesByType('paint')

// new way
const observer = new PerformanceObserver(list => {
  const entries = list.getEntries()
})
observer.observe({ entryTypes: ['paint'] })
```

Some of the metrics that are only available in the new API are
+ Long tasks
+ LCP(Largest Contentful Paint)
+ Element timings

### Pending tasks
+ [x] Implementation of PerformanceObserver, capturing longtasks
+ [x] Longtask Config
+ [x] Add docs about Longtasks and how to interpret them in the UI

PR #581, opened 2020-01-23
feat(rum-core): capture long tasks and lcp using performance observer

### LINKED ISSUE #376: add support for Largest Contentful Paint(LCP) in timeline
+ Add support for LCP in addition to the already available marks as part of the page-load transaction. https://github.com/elastic/apm-agent-rum-js/blob/c743f701a1cd65104612461fd19f36faceba5ed4/packages/rum-core/src/common/utils.js#L384-L399

Spec draft - https://wicg.github.io/largest-contentful-paint/
Blog post - https://web.dev/largest-contentful-paint/
Available from chrome 77

vigneshshanmugam:
+ fix #376 
+ Introduces **Performance Observer** to capture `Longtask` and `Largest contentful paint` during transaction lifetime. 
Screenshot 2020-01-27 at 12 21 47

+ LCP is added as a mark whereas `Longtasks` are created as spans to associate them in the breakdowns which provides a overview of how much time it spent on doing long tasks.
Screenshot 2020-01-27 at 12 21 24

### DIFF
--- a/packages/rum-core/src/performance-monitoring/capture-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-navigation.js
@@ -34,7 +34,11 @@ import {
   MEASURE,
   FIRST_CONTENTFUL_PAINT
 } from '../common/constants'
-import { stripQueryStringFromUrl, PERF, isPTSupported } from '../common/utils'
+import {
+  stripQueryStringFromUrl,
+  PERF,
+  isPerfTimelineSupported
+} from '../common/utils'
 
 /**
  * Navigation Timing Spans
@@ -245,18 +249,19 @@ function getNavigationTimingMarks() {
 }
 
 /**
- * Paint Timing Metrics that is available during page load
- * `first-paint` and `first-contentful-paint`
- * https://www.w3.org/TR/paint-timing/
+ * Get the `first-contentful-paint` metric that is available during
+ * page-load using the Paint Timing Metrics API
+ * SPEC - https://www.w3.org/TR/paint-timing/
  */
 function getFirstContentfulPaint() {
   let fcp
-  if (isPTSupported()) {
+  if (isPerfTimelineSupported()) {
     const entries = PERF.getEntriesByType(PAINT)
     if (entries.length > 0) {
       const timing = PERF.timing
       /**
-       * To avoid capturing the unload event handler effect in paint timings
+       * To avoid capturing the unload event handler effect
+       * as part of the transaction duration
        */
       const unloadDiff = timing.fetchStart - timing.navigationStart
       const fcpEntry = entries.filter(
@@ -344,7 +349,7 @@ function captureNavigation(transaction) {
     transaction.addMarks(getPageLoadMarks())
   }
 
-  if (isPTSupported()) {
+  if (isPerfTimelineSupported()) {
     const trStart = transaction._start
     /**
      * Capture resource timing information as spans
--- a/packages/rum-core/src/performance-monitoring/capture-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-navigation.js
@@ -28,9 +28,13 @@ import {
   RESOURCE_INITIATOR_TYPES,
   MAX_SPAN_DURATION,
   USER_TIMING_THRESHOLD,
-  PAGE_LOAD
+  PAGE_LOAD,
+  PAINT,
+  RESOURCE,
+  MEASURE,
+  FIRST_CONTENTFUL_PAINT
 } from '../common/constants'
-import { stripQueryStringFromUrl, getPageLoadMarks } from '../common/utils'
+import { stripQueryStringFromUrl, PERF, isPTSupported } from '../common/utils'
 
 /**
  * Navigation Timing Spans
@@ -208,6 +212,80 @@ function getApiSpanNames({ spans }) {
   return apiCalls
 }
 
+const navigationTimingKeys = [
+  'fetchStart',
+  'domainLookupStart',
+  'domainLookupEnd',
+  'connectStart',
+  'connectEnd',
+  'secureConnectionStart',
+  'requestStart',
+  'responseStart',
+  'responseEnd',
+  'domLoading',
+  'domInteractive',
+  'domContentLoadedEventStart',
+  'domContentLoadedEventEnd',
+  'domComplete',
+  'loadEventStart',
+  'loadEventEnd'
+]
+
+function getNavigationTimingMarks() {
+  const timing = PERF.timing
+  const fetchStart = timing.fetchStart
+  const marks = {}
+  navigationTimingKeys.forEach(function(timingKey) {
+    const m = timing[timingKey]
+    if (m && m >= fetchStart) {
+      marks[timingKey] = m - fetchStart
+    }
+  })
+  return marks
+}
+
+/**
+ * Paint Timing Metrics that is available during page load
+ * `first-paint` and `first-contentful-paint`
+ * https://www.w3.org/TR/paint-timing/
+ */
+function getFirstContentfulPaint() {
+  let fcp
+  if (isPTSupported()) {
+    const entries = PERF.getEntriesByType(PAINT)
+    if (entries.length > 0) {
+      const timing = PERF.timing
+      /**
+       * To avoid capturing the unload event handler effect in paint timings
+       */
+      const unloadDiff = timing.fetchStart - timing.navigationStart
+      const fcpEntry = entries.filter(
+        entry => entry.name === FIRST_CONTENTFUL_PAINT
+      )
+      const { startTime } = fcpEntry[0]
+      fcp = unloadDiff >= 0 ? startTime - unloadDiff : startTime
+    }
+  }
+  return fcp
+}
+
+function getPageLoadMarks() {
+  const marks = getNavigationTimingMarks()
+  const agent = {
+    timeToFirstByte: marks.responseStart,
+    domInteractive: marks.domInteractive,
+    domComplete: marks.domComplete
+  }
+  const fcp = getFirstContentfulPaint()
+  if (fcp) {
+    agent.firstContentfulPaint = fcp
+  }
+  return {
+    navigationTiming: marks,
+    agent
+  }
+}
+
 function captureNavigation(transaction) {
   /**
    * Do not capture timing related information when the
@@ -218,7 +296,6 @@ function captureNavigation(transaction) {
     return
   }
 
-  const perf = window.performance
   /**
    * Both start and end threshold decides if a span must be
    * captured as part of the transaction
@@ -246,7 +323,7 @@ function captureNavigation(transaction) {
     const trStart = 0
     transaction._start = trStart
 
-    const timings = perf.timing
+    const timings = PERF.timing
     createNavigationTimingSpans(
       timings,
       timings.fetchStart,
@@ -267,12 +344,12 @@ function captureNavigation(transaction) {
     transaction.addMarks(getPageLoadMarks())
   }
 
-  if (typeof perf.getEntriesByType === 'function') {
+  if (isPTSupported()) {
     const trStart = transaction._start
     /**
      * Capture resource timing information as spans
      */
-    const resourceEntries = perf.getEntriesByType('resource')
+    const resourceEntries = PERF.getEntriesByType(RESOURCE)
     const apiCalls = getApiSpanNames(transaction)
 
     createResourceTimingSpans(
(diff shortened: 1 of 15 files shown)


packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
```diff
@@ -0,0 +1,132 @@
...
+  LARGEST_CONTENTFUL_PAINT,
+  PAGE_LOAD
+} from '../common/constants'
+import { noop } from '../common/utils'
+import Span from './span'
+
+/**
+ * Create Spans for the long task entries
+ *
+ * TODO: Make use of the long task attribution data
```
vigneshshanmugam: currently even if we add it to the context, it wont be any help to the user since the UI does not make use of it. We can add this later.

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
```diff
@@ -0,0 +1,132 @@
...
+     */
+    try {
+      /**
+       * Except longtasks other entries support buffered
+       * performance entries
+       */
+      if (type === PAGE_LOAD) {
+        this.po.observe({ type: LARGEST_CONTENTFUL_PAINT, buffered: true })
+      }
+      this.po.observe({ type: LONG_TASK })
```
jahtalab: The correct signature is`observe({entryTypes:[LONG_TASK]})` . These seems to work as well, but it's better to follow the docs. Is there any reason you're using this?

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
```diff
@@ -0,0 +1,132 @@
...
+     * Safari throws an error when PerformanceObserver is
+     * observed for unknown entry types
+     */
+    try {
+      /**
+       * Except longtasks other entries support buffered
+       * performance entries
+       */
+      if (type === PAGE_LOAD) {
+        this.po.observe({ type: LARGEST_CONTENTFUL_PAINT, buffered: true })
```
jahtalab: Would you please mentioned that the largest contentfull paint is not official and reference the drafts: https://wicg.github.io/largest-contentful-paint/

Also, maybe mention that we use `buffered:true` to capture these events if the happened before calling observe.

packages/rum-core/src/performance-monitoring/capture-navigation.js
```diff
@@ -208,6 +212,80 @@ function getApiSpanNames({ spans }) {
...
+  let fcp
+  if (isPTSupported()) {
+    const entries = PERF.getEntriesByType(PAINT)
+    if (entries.length > 0) {
+      const timing = PERF.timing
+      /**
+       * To avoid capturing the unload event handler effect in paint timings
+       */
+      const unloadDiff = timing.fetchStart - timing.navigationStart
+      const fcpEntry = entries.filter(
```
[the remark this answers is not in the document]
vigneshshanmugam: you mean array.find? or move this to utils and call it as find?

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
vigneshshanmugam: type: LONG_TASK super seeds the entryTypes proposal and also entryTypes does not allow buffered flag. since everything is chrome only at the moment, we can as well use the singular form instead of passing in multiple fields. I will add a comment.

vigneshshanmugam:
@jahtalab 
>A lot of the refactoring done in this PR are not necessarily related to the main feature. I think we should avoid mixing too many changes into a single PR. This also makes it faster to get feedback and merge.

I will try to keep them separate going forward. But it was basically to move a check to util and push entry types as constants

> Is there any other type of information we can make available in the long task spans, currently it seems that it would be hard to actually find out what is happing during the long task.

We would not be able to associate to the users code since we would need stack trace of the current running code during long task to associate. I have added a TODO for now in the code to explore it more

> Instead of the approach taken here can we observe the events when the agent inits and create spans on the current transaction as they come along? This approach seems simpler to me since we don't need to listen for transaction events and only check if the current transaction exist

I thought about this before and the approach feels less flexible 
 + We would be observing eagerly than lazily -> What i mean here is we would observe and discard if there is no currently running transaction instead of starting/stopping when required. 
+ Perf observer for long tasks has performance cost associated with it and that is one of the reasons why buffered flag is not supported for now. Having an ability to start/disconnect provides a bandage instead of keeping the observer in always on mode. 
+ We would be working with every-growing perf entry buffer instead of start/stop during transaction lifetime. And I cannot think of a place where would we disconnect if we go with the mentioned approach. 

I am happy to hear your thoughts on why you want to move way from transaction events. 

>We should also consider adding long tasks only for managed transactions

Good point, I will use the captureTimings flag to do this.

packages/rum-core/src/performance-monitoring/capture-navigation.js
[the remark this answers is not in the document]
vigneshshanmugam: It is supported even in IE9 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter, I don't think its required to create an util for this use case.

vigneshshanmugam:
Notes for reviewers.  

+ Longtasks buffered flag will be shipped in Chrome 81
+ Code that is causing the long tasks will be hard to measure using RUM as the attribution data does not give us the stack trace. We can do it only for Synthetic tests where we can get the trace data from chrome. 
+ We can enable the long tasks capturing behind a flag and link users to the actual document on what could have caused it. Helps users who have not heard about long tasks yet.

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
jahtalab: We will add this as comments!

jahtalab:
Had a meeting with @vigneshshanmugam :
* We will add the documentation in a separate PR. we should document longtasks and explain self and unknonw and also some hints about how to debug these spans.
* We decided to add references to the w3c spec in the code: https://w3c.github.io/longtasks/
* We also discussed the approach taken here.

vigneshshanmugam:
@jahtalab I have updated the code with more description on spec and other issues that I have encountered. Also made use of the longtask attribution data which could give some clue on what was the culprit.

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    span.end(end)
+
+    longTaskSpans.push(span)
+  }
+  return longTaskSpans
+}
+
+export function onPerformanceEntry(list, transaction) {
```
jahtalab: nit: use a more descriptive name here, like `capturePerformanceEntries`

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
```diff
@@ -0,0 +1,209 @@
...
+ */
+function isManaged(transaction) {
+  return transaction.captureTimings
+}
+
+/**
+ * Checks whether the give transaction is a page load transaction
+ */
+function isPageLoadTransaction(transaction) {
+  return transaction.type === PAGE_LOAD
```
jahtalab: Can we avoid having these transaction type conditions here? the reason is that having these conditions all over the code makes it hard to figure out what is captured for a particular transaction. we should keep the special cases in the Transaction service as much as possible.

vigneshshanmugam:
We had a discussion and decided on few things
+ Going to move around the stop and start logic inside transaction-service as events could be triggered by other transactions and spans would be captured incorrectly. 
+ move transaction specific details from the performance recorder and let transaction service handle it. 
+ add tests to make sure the events are in order.

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
```diff
@@ -0,0 +1,188 @@
...
+  for (let i = 0; i < longtasks.length; i++) {
+    /**
+     * name of the long task can be any one of the type listed here
+     * depeding on the origin of the longtask
+     * https://w3c.github.io/longtasks/#sec-PerformanceLongTaskTiming
+     */
+    const { name, startTime, duration, attribution } = longtasks[i]
+    const end = startTime + duration
+    const kind = LONG_TASK
+    const span = new Span(`Longtask(${name})`, kind, { startTime })
```
jahtalab: can `name` ever be undefined?

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -691,4 +695,62 @@ describe('TransactionService', function() {
...
+
+    it('should start/stop performance recorder for managed transaction', async () => {
+      const pageLoadTr = trService.startTransaction('test', PAGE_LOAD, {
+        managed: true
+      })
+      expect(startSpy).toHaveBeenCalledTimes(2)
+      expect(startSpy.calls.allArgs()).toEqual([
+        ['largest-contentful-paint'],
+        ['longtask']
+      ])
```
jahtalab: Good call to check the args 👍

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -691,4 +695,62 @@ describe('TransactionService', function() {
     tr = transactionService.getCurrentTransaction()
     expect(tr.type).toBe('page-load')
   })
+
+  describe('performance entry recorder', () => {
```
jahtalab: Maybe another test for the scenario if `startTransaction` is called when another non-reusable transaction already running. This would cover the scenario of making sure that `stop` is called before the `start` for the new transaction.

packages/rum-core/test/common/utils.spec.js
```diff
@@ -70,19 +70,6 @@ describe('lib/utils', function() {
...
-    expect(marks.fetchStart).toBeGreaterThanOrEqual(0)
-    expect(marks.domInteractive).toBeGreaterThanOrEqual(0)
-    expect(marks.domComplete).toBeGreaterThanOrEqual(0)
-    expect(marks.loadEventEnd).toBeGreaterThanOrEqual(0)
-  })
-
-  it('should getPaintTimingMarks', function() {
-    var marks = utils.getPaintTimingMarks()
-    expect(marks).toEqual({})
-  })
```
jahtalab: Did you mean to move these tests? or do you think they are not necessary anymore?

Maybe we should replace these with some improved tests, WDYT?

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
vigneshshanmugam: it defaults to `unknown`, so not really.

packages/rum-core/test/common/utils.spec.js
vigneshshanmugam: we already have the test for FCP and agent marks in captureNavigation. Having this test doesn't provide any value and we have to expose that function just for the tests, so removed it.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
vigneshshanmugam: Yeah sounds good, will add it.

packages/rum-core/src/performance-monitoring/perf-entry-recorder.js
```diff
@@ -0,0 +1,188 @@
...
+
+  for (let i = 0; i < longtasks.length; i++) {
+    /**
+     * name of the long task can be any one of the type listed here
+     * depeding on the origin of the longtask
+     * https://w3c.github.io/longtasks/#sec-PerformanceLongTaskTiming
+     */
+    const { name, startTime, duration, attribution } = longtasks[i]
+    const end = startTime + duration
+    const kind = LONG_TASK
```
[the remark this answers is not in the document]
vigneshshanmugam: I am also not sure though, We would need like a generic one something like `browser-events`. then we can simply group them all in to this category instead of expanded ones currently like `hard-navigation`, `resource`, `measure` and so on.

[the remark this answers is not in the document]
vigneshshanmugam: The reason why I did not put it as `app` is also because these can come from third party scripts and also ad's. Just FYI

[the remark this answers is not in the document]
jahtalab: I think `longtask` is too specific, we probably should group these kind of spans under something like `browser-metric` or `perf-metric`. These are generic enough that we can use them for spans that don't fit into other kinds.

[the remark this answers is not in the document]
vigneshshanmugam: i remember it now, the reason why it is `longtask` was because we calculate breakdowns based on span type and having them as longtasks would give an advantage to user that % of transaction is spent on long tasks vs if we use `browser-metric or events` its not super helpful. WDYT?

Issue #496, opened 2019-11-19
Capture user interactions


jahtalab:
* We should capture user interactions such as clicks, scrolls, mouse and keyboard (and their delays)
* We should investigate the best way to report and visualize this information (report as part of the active transactions or create a transaction or report aggregated counts and delays )

Todo: 
- [ ] Add documentation

Related:
* https://github.com/elastic/apm-agent-rum-js/issues/407

jahtalab:
Focusing on click event for now, it can either result in a new transaction or in an action that results in some type of task (e.g. XHR), we probably don't need to capture the click callback as a span.
We should use the DOM element to drive the transaction name.

we can also capture aggregate timing for the event handlers like:
```
{
  click: {
   count: 2,
   totalTime: 50
}
}
```

alexandreservian:
Hi, at my work we use apm rum. But I dont know how to use **user-interaction** in the buttons. Example: I had many times to click a button dispach one feach for apiRest. In other test, this same button dispach a new transaction and one span. My question is. How do I use correctly  **user-interaction**. And How do I define the name of button EX: "Click - buy-cards'

PR #588, opened 2020-01-27
feat(rum-core): add EventTarget patch

jahtalab:
part of #496 
This PR adds patching for EventTarget. It doesn't not register any event listeners for the EventTarget patch events, will do that in a separate PR.
- [x] Add more tests

### DIFF
--- a/packages/rum-core/src/common/patching/event-target-patch.js
+++ b/packages/rum-core/src/common/patching/event-target-patch.js
@@ -30,11 +30,9 @@ import {
   REMOVE_EVENT_LISTENER_STR,
   EVENT_TARGET
 } from '../constants'
-
 import { apmSymbol } from './patch-utils'
 
 const eventTypes = ['click']
-
 const eventTypeSymbols = {}
 
 for (let i = 0; i < eventTypes.length; i++) {
@@ -51,6 +49,15 @@ export function patchEventTarget(callback) {
   const nativeAddEventListener = proto[ADD_EVENT_LISTENER_STR]
   const nativeRemoveEventListener = proto[REMOVE_EVENT_LISTENER_STR]
 
+  /**
+   * This function checks whether we have already patched an event listener
+   * according to the rules defined in:
+   * https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener
+   * https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
+   * In short, the uniqueness of an event listener depends on even type, capture argument
+   * and listener function instance. This ensures that no two event listeners that have the same value
+   * for these arguments are registered.
+   */
   function findTaskIndex(existingTasks, eventType, listenerFn, capture) {
     for (let i = 0; i < existingTasks.length; i++) {
       const task = existingTasks[i]
@@ -65,14 +72,15 @@ export function patchEventTarget(callback) {
     return -1
   }
 
+  /**
+   * isCapture unifies the multiple ways 'useCapture'
+   * argument can be passed to addEventlistener
+   * For more: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
+   */
   function isCapture(options) {
     let capture
-    if (options === undefined) {
-      capture = false
-    } else if (options === true) {
-      capture = true
-    } else if (options === false) {
-      capture = false
+    if (typeof options === 'boolean') {
+      capture = options
     } else {
       capture = options ? !!options.capture : false
     }
@@ -129,6 +137,27 @@ export function patchEventTarget(callback) {
     return wrappingFn
   }
 
+  function getWrappingFn(target, eventType, listenerFn, options) {
+    const eventSymbol = eventTypeSymbols[eventType]
+    let existingTasks = target[eventSymbol]
+
+    if (existingTasks) {
+      let capture = isCapture(options)
+      let taskIndex = findTaskIndex(
+        existingTasks,
+        eventType,
+        listenerFn,
+        capture
+      )
+      if (taskIndex !== -1) {
+        let task = existingTasks[taskIndex]
+        existingTasks.splice(taskIndex, 1)
+        return task.wrappingFn
+      }
+    }
+    return listenerFn
+  }
+
   proto[ADD_EVENT_LISTENER_STR] = function(
     eventType,
     listenerFn,
@@ -162,24 +191,15 @@ export function patchEventTarget(callback) {
     }
     const target = this
 
-    const eventSymbol = eventTypeSymbols[eventType]
+    let wrappingFn = getWrappingFn(
+      target,
+      eventType,
+      listenerFn,
+      optionsOrCapture
+    )
 
-    let existingTasks = target[eventSymbol]
-    let capture = isCapture(optionsOrCapture)
     let args = [...arguments]
-    if (existingTasks) {
-      let taskIndex = findTaskIndex(
-        existingTasks,
-        eventType,
-        listenerFn,
-        capture
-      )
-      if (taskIndex !== -1) {
-        let task = existingTasks[taskIndex]
-        args[1] = task.wrappingFn
-        existingTasks.splice(taskIndex, 1)
-      }
-    }
+    args[1] = wrappingFn
     return nativeRemoveEventListener.apply(target, args)
   }
 }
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -0,0 +1,185 @@
...
+  function isCapture(options) {
+    let capture
+    if (options === undefined) {
+      capture = false
+    } else if (options === true) {
+      capture = true
+    } else if (options === false) {
+      capture = false
+    } else {
+      capture = options ? !!options.capture : false
```
vigneshshanmugam: there are also other options like `passive` and `once`. any reason why we are skipping them? if yes, adding comments would be good idea.

packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -0,0 +1,185 @@
...
+ */
+
+import {
+  SCHEDULE,
+  INVOKE,
+  ADD_EVENT_LISTENER_STR,
+  REMOVE_EVENT_LISTENER_STR,
+  EVENT_TARGET
+} from '../constants'
+
```
vigneshshanmugam: nit: no line between imports.

packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -0,0 +1,185 @@
...
+        task.listenerFn === listenerFn &&
+        task.capture === capture
+      ) {
+        return i
+      }
+    }
+    return -1
+  }
+
+  function isCapture(options) {
```
vigneshshanmugam: Seeing the below logic, Shouldn't something like this be enough? 
```
isCapture = (options && options.capture || options) === true;
```

packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -0,0 +1,185 @@
...
+    optionsOrCapture
+  ) {
+    if (eventTypes.indexOf(eventType) < 0 || typeof listenerFn !== 'function') {
+      return nativeRemoveEventListener.apply(this, arguments)
+    }
+    const target = this
+
+    const eventSymbol = eventTypeSymbols[eventType]
+
+    let existingTasks = target[eventSymbol]
```
vigneshshanmugam: Can you also move this logic inside unwrap function like you did for wrapping.

packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -0,0 +1,185 @@
...
+    let capture = isCapture(optionsOrCapture)
+    let args = [...arguments]
+    if (existingTasks) {
```
vigneshshanmugam: existingTasks.length > 0

packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -0,0 +1,185 @@
...
+      eventType,
+      listenerFn,
+      capture,
+      wrappingFn
+    }
+
+    existingTasks.push(task)
+
+    function wrappingFn(event) {
+      task.event = event
```
vigneshshanmugam: do we need event part of the task? since we already have event type as part of the task.

packages/rum-core/src/common/patching/event-target-patch.js
jahtalab: This only checks if the `existingTasks` is not undefined otherwise if there are no elements it won't find the wrapping function.

packages/rum-core/src/common/patching/event-target-patch.js
jahtalab: I've simplified the function a little bit. The above expression won't work if option is `{capture:1}`. but, In any case I prefer to be explicit and expand the logical expression to avoid errors.

packages/rum-core/src/common/patching/event-target-patch.js
jahtalab: I think the actual event can be used to derive information about it, like the `srcElement` which could be used in the naming the transaction.

jahtalab:
Thanks @vigneshshanmugam for the review, 

* regarding different approaches since they mostly related to the implementation it is difficult to describe them without writing some code. I don't see a lot of benefit in documenting these to be honest.
* Re. the xhrs the current doesn't effect event types that we are not interested in therefore those would have the native behaviour. However, I've added some tests to make sure unknown event types are not affected.

packages/rum-core/src/common/patching/event-target-patch.js
vigneshshanmugam: I understand that part, but this would either be undefined || [ ] for the first time. So instead of just checking for undefined, you could do both `existingTasks && existingTasks.length > 0` and that would avoid unnecessary function calls like `isCapture` and `findTaskIndex`. 

Does it make sense?

packages/rum-core/test/common/event-target-patch.spec.js
```diff
@@ -0,0 +1,142 @@
...
+    element.dispatchEvent(event)
+    expect(count).toBe(1)
+    expect(events.length).toBe(0)
+    element.removeEventListener(eventType, listener)
+    element.dispatchEvent(event)
+    expect(count).toBe(1)
+  })
+
+  if (window.EventTarget) {
+    it('should patch addEventListener', () => {
```
vigneshshanmugam: nit: patch addEventListener/removeEventListener or just event target

vigneshshanmugam: nit: patch addEventListener/removeEventListener or just event target

packages/rum-core/test/common/event-target-patch.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      let count = 0
+      let element = document.createElement('div')
+      const listener = e => {
+        expect(e.type).toBe('click')
+        count++
+      }
+
+      element.addEventListener('click', listener)
+      element.addEventListener('click', listener)
+      element.addEventListener('click', listener, { capture: false })
```
vigneshshanmugam: needs test for passive event listeners.

packages/rum-core/src/common/patching/event-target-patch.js
vigneshshanmugam: Can we just pass that instead, reason was it would be hard for the Garbage collection to kick in till the object is being passed around. we can also do this in the coming PR where we will use these tasks and create transaction. What do you think?

jahtalab:
@vigneshshanmugam and I discussed the different approaches that we could take and they mostly come down to storing the wrapped function on an object and it seems the target is the better place to do this.

packages/rum-core/test/common/event-target-patch.spec.js
jahtalab: @vigneshshanmugam , I've looked into `passive` and `once` options, it still doesn't seem to be relevant to us to test and given that these are not supported on most of the browsers we test on (even chrome have added it in 51), I don't see a lot of benefit in adding these at this point.

vigneshshanmugam: We can do test like running them only on browsers supporting it. I do not want to block the PR for this, but it would be great to have the tests anyway.

packages/rum-core/test/common/event-target-patch.spec.js
```diff
@@ -0,0 +1,164 @@
...
+      let element = document.createElement('div')
+      const listener = e => {
+        expect(e.type).toBe('click')
+        count++
+      }
+
+      element.addEventListener('click', listener)
+      element.addEventListener('click', listener)
+      /**
+       * Form with options object is only supported on
```
vigneshshanmugam: I didn't quite get this comment. what do you mean by `form` here? I am confused because this comment says we cannot pass object and line 144 does exactly that.

packages/rum-core/test/common/event-target-patch.spec.js
```diff
@@ -0,0 +1,164 @@
...
+
+  afterAll(function() {
+    cancelFn()
+  })
+
+  beforeEach(function() {
+    events = []
+  })
+
+  function createEvent(eventType) {
```
vigneshshanmugam: we have a similar function to create events. https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/test/error-logging/error-logging.spec.js#L317-L331

Can we make use of that and move it to util?

PR #590, opened 2020-01-27
fix(rum-core): do not capture timing spans for unsampled transactions

### LINKED ISSUE #579: Avoid capturing multiple timings if the transaction is unsampled
We should not capture resource timing API, user timing, breakdown metrics if the transaction is upsampled

vigneshshanmugam:
+ fix #579

### DIFF
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -156,8 +156,6 @@ class TransactionService {
       tr = this.ensureCurrentTransaction(name, type, perfOptions)
     }
 
-    tr.captureTimings = true
-
     if (tr.type === PAGE_LOAD) {
       tr.options.checkBrowserResponsiveness = false
       if (perfOptions.pageLoadTraceId) {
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -175,6 +173,14 @@ class TransactionService {
       }
     }
 
+    /**
+     * For unsampled transactions, avoid capturing various timing informations
+     * as spans as they are dropped anyways before sending to the server
```
jahtalab: nit: you should use 'information' instead of 'informations',  also 'as they' is redundant.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -170,6 +170,25 @@ describe('TransactionService', function() {
     tr2.detectFinish()
   })
 
+  it('should not capture timings as spans for unsampled transactions', done => {
+    const unMock = mockGetEntriesByType()
+
+    config.events.observe(TRANSACTION_END, () => {
```
jahtalab: Should use the transaction that is passed to the event listener instead

PR #595, opened 2020-01-30
chore: enable id_field_data for elasticsearch

jahtalab:
This PR enables id_field_data for elasticsearch instance defined in the docker-compose file. This option is required to avoid elasticsearch warnings. For more see https://github.com/elastic/apm-integration-testing/pull/696

### DIFF
--- a/dev-utils/docker-compose.yml
+++ b/dev-utils/docker-compose.yml
@@ -10,6 +10,7 @@ services:
       - "cluster.routing.allocation.disk.threshold_enabled=false"
       - "discovery.type=single-node"
       - "path.data=/usr/share/elasticsearch/data/${STACK_VERSION}"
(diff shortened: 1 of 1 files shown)


dev-utils/docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
       - "cluster.routing.allocation.disk.threshold_enabled=false"
       - "discovery.type=single-node"
       - "path.data=/usr/share/elasticsearch/data/${STACK_VERSION}"
+      - "indices.id_field_data.enabled=true"
```
vigneshshanmugam: seems like the old version does not support it, we would need to feature detect and only do it like how apm-integration-testing is doing it.

jahtalab:
We would have to manually add the environment option in our local tests

PR #597, opened 2020-01-31
docs: interpreting longtasks as spans

vigneshshanmugam:
+ Added config for monitoring longtasks `monitorLongtasks` which is set to true by default. 
+ Added doc on how to interpret these tasks in the UI.

### DIFF
--- a/docs/advanced-topics.asciidoc
+++ b/docs/advanced-topics.asciidoc
@@ -1,37 +1,37 @@
 [[advanced-topics]]
 == Advanced Topics
 
-* <<custom-transactions, Custom Transactions>>
-* <<custom-transaction-name, Initial page load transaction names>>
 * <<longtasks>>
 * <<typescript>>
+* <<custom-transactions, Custom Transactions>>
+* <<custom-transaction-name, Initial page load transaction names>>
 
 [[longtasks]]
-=== How to interpret Longtask Spans in the UI
+=== How to interpret long task spans in the UI
 
-Longtasks is a new performance metric that helps measure the responsiveness and
-helps understand the bad user experience on the web. It enables detecting tasks
-that monopolize the UI thread for extended periods (greater than 50 milliseconds) and block other
-critical tasks from being executed as stated in the official spec[https://github.com/w3c/longtasks].
+Long tasks is a new performance metric that can be used for measuring the
+responsiveness of an application and helps developers to understand the bad user
+experience. It enables detecting tasks that monopolize the UI thread for
+extended periods (greater than 50 milliseconds) and block other
+critical tasks from being executed as stated in the https://github.com/w3c/longtasks[official spec].
 
-RUM agent automatically captures these Longtasks and include them as Spans as
-part of the transaction. Since Longtasks currently does not have the full information on
+RUM agent automatically captures these Long tasks and include them as spans as
+part of the transaction. Since long tasks currently does not have the full information on
 which part of code cause slowness, it would be hard to interpret these spans.
-Some tips below would interpret these spans in a meaningful way
+Below you can find some tips to help with interpreting long task spans:
 
-+ Name of the Longtask span could be `self`, `same-origin`, etc. would imply the
++ Name of the long task span (eg: `self`, `same-origin`) would imply the
   origin of the task. It could be the current browsing context or inside iframes.
 
 + Context of the span is enriched with useful information like `attribution`
   (the type of work such as script, layout, etc), `type`, `id` and `name` which
   determines the culprit container(such as window, iframe, embed or object)
-  responsible for the long task
+  responsible for the long task.
 
 With the help of the transaction timeline and the span timings, one could dig
-deeper by marking the slow application code using User Timing
-API[https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark]
-which would be again captured as Spans by the agent and combining them along with Longtasks would
-reveal the true source code location.
+deeper by marking the slow application code using https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark[User Timing
+API] which would be again captured as spans by the agent and combining them along
+with long tasks would reveal the true source code location.
 
 [[typescript]]
 === Using with TypeScript
--- a/docs/advanced-topics.asciidoc
+++ b/docs/advanced-topics.asciidoc
@@ -20,12 +20,12 @@ part of the transaction. Since long tasks currently does not have the full infor
 which part of code cause slowness, it would be hard to interpret these spans.
 Below you can find some tips to help with interpreting long task spans:
 
-+ Name of the long task span (eg: `self`, `same-origin`) would imply the
+* The name of the long task span, e.g.: `self`, `same-origin`, etc., implies the
   origin of the task. It could be the current browsing context or inside iframes.
 
-+ Context of the span is enriched with useful information like `attribution`
-  (the type of work such as script, layout, etc), `type`, `id` and `name` which
-  determines the culprit container(such as window, iframe, embed or object)
+* Context of the span is enriched with useful information like `attribution`
+  (the type of work, such as script, layout, etc), `type`, `id` and `name`, which
+  determines the culprit container (such as window, iframe, embed or object)
   responsible for the long task.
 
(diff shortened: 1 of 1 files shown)


docs/advanced-topics.asciidoc
```diff
@@ -3,8 +3,35 @@
 
 * <<custom-transactions, Custom Transactions>>
 * <<custom-transaction-name, Initial page load transaction names>>
+* <<longtasks>>
 * <<typescript>>
 
+[[longtasks]]
+=== How to interpret Longtask Spans in the UI
+
+Longtasks is a new performance metric that helps measure the responsiveness and
```
jahtalab:
```suggestion
Long tasks is a new performance metric that can be used for measuring the responsiveness of an application and 
```

docs/advanced-topics.asciidoc
```diff
@@ -3,8 +3,35 @@
...
+helps understand the bad user experience on the web. It enables detecting tasks
```
jahtalab:
```suggestion
helps application developers to understand the user experience on their application. It enables detecting tasks
```

docs/advanced-topics.asciidoc
```diff
@@ -3,8 +3,35 @@
...
+that monopolize the UI thread for extended periods (greater than 50 milliseconds) and block other
+critical tasks from being executed as stated in the official spec[https://github.com/w3c/longtasks].
+
+RUM agent automatically captures these Longtasks and include them as Spans as
+part of the transaction. Since Longtasks currently does not have the full information on
+which part of code cause slowness, it would be hard to interpret these spans.
+Some tips below would interpret these spans in a meaningful way
```
jahtalab:
```suggestion
Below you can find some tips to help with interpreting Longtask spans:
```

docs/advanced-topics.asciidoc
```diff
@@ -3,8 +3,35 @@
...
+
++ Name of the Longtask span could be `self`, `same-origin`, etc. would imply the
```
jahtalab:
```suggestion
+ Name of the Longtask span (e.g. `self`, `same-origin`) would imply the
```

docs/advanced-topics.asciidoc
```diff
@@ -3,8 +3,35 @@
...
+* <<longtasks>>
 * <<typescript>>
 
+[[longtasks]]
+=== How to interpret Longtask Spans in the UI
+
+Longtasks is a new performance metric that helps measure the responsiveness and
+helps understand the bad user experience on the web. It enables detecting tasks
+that monopolize the UI thread for extended periods (greater than 50 milliseconds) and block other
+critical tasks from being executed as stated in the official spec[https://github.com/w3c/longtasks].
```
bmorelli25:
```suggestion
critical tasks from being executed as stated in the https://github.com/w3c/longtasks[official spec].
```

docs/advanced-topics.asciidoc
```diff
@@ -3,8 +3,35 @@
...
+  origin of the task. It could be the current browsing context or inside iframes.
+
++ Context of the span is enriched with useful information like `attribution`
+  (the type of work such as script, layout, etc), `type`, `id` and `name` which
+  determines the culprit container(such as window, iframe, embed or object)
+  responsible for the long task
+
+With the help of the transaction timeline and the span timings, one could dig
+deeper by marking the slow application code using User Timing
+API[https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark]
```
bmorelli25: This, should be:
```
deeper by marking the slow application code using
https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark[User Timing API],
```

docs/advanced-topics.asciidoc
```diff
@@ -3,8 +3,35 @@
 
 * <<custom-transactions, Custom Transactions>>
 * <<custom-transaction-name, Initial page load transaction names>>
+* <<longtasks>>
 * <<typescript>>
```
bmorelli25: Kind of unrelated to this PR, but since we're touching this file, let's move the "custom-*" links to the bottom of the list and remove the custom titles so we match the order in the docs:
```
* <<longtasks>>
* <<typescript>>
* <<custom-transactions>>
* <<custom-transaction-name>>
```

docs/advanced-topics.asciidoc
vigneshshanmugam: changed this a bit since application was called in multiple places.

vigneshshanmugam:
@bmorelli25 You are right, I have adjusted the doc with with wording  `long tasks` everywhere.

docs/advanced-topics.asciidoc
```diff
@@ -1,10 +1,37 @@
...
-* <<typescript>>
 
+[[longtasks]]
+=== How to interpret long task spans in the UI
+
+Long tasks is a new performance metric that can be used for measuring the
+responsiveness of an application and helps developers to understand the bad user
+experience. It enables detecting tasks that monopolize the UI thread for
+extended periods (greater than 50 milliseconds) and block other
+critical tasks from being executed as stated in the https://github.com/w3c/longtasks[official spec].
```
bmorelli25: What if we swapped the what and the why? Something like:
```
Long tasks enables the detection of tasks that monopolize the
UI thread for extended periods of time (greater than 50 milliseconds) and block other
critical tasks from being executed -- as stated in the https://github.com/w3c/longtasks[official spec].
This new performance metric is useful for measuring the responsiveness of an application,
and can help developers better understand bad user experiences.
```

docs/advanced-topics.asciidoc
```diff
@@ -1,10 +1,37 @@
...
+
+RUM agent automatically captures these Long tasks and include them as spans as
+part of the transaction. Since long tasks currently does not have the full information on
+which part of code cause slowness, it would be hard to interpret these spans.
+Below you can find some tips to help with interpreting long task spans:
+
++ Name of the long task span (eg: `self`, `same-origin`) would imply the
```
bmorelli25:
```suggestion
* The name of the long task span, e.g.: `self`, `same-origin`, etc., implies the
```

docs/advanced-topics.asciidoc
```diff
@@ -1,10 +1,37 @@
...
+  origin of the task. It could be the current browsing context or inside iframes.
+
++ Context of the span is enriched with useful information like `attribution`
```
bmorelli25:
```suggestion
* Context of the span is enriched with useful information like `attribution`
```

docs/advanced-topics.asciidoc
```diff
@@ -1,10 +1,37 @@
...
+  (the type of work such as script, layout, etc), `type`, `id` and `name` which
```
bmorelli25:
```suggestion
  (the type of work, such as script, layout, etc), `type`, `id` and `name`, which
```

docs/advanced-topics.asciidoc
```diff
@@ -1,10 +1,37 @@
...
+  determines the culprit container(such as window, iframe, embed or object)
```
bmorelli25:
```suggestion
  determines the culprit container (such as window, iframe, embed or object)
```

docs/advanced-topics.asciidoc
```diff
@@ -1,10 +1,37 @@
...
+  responsible for the long task.
+
+With the help of the transaction timeline and the span timings, one could dig
+deeper by marking the slow application code using https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark[User Timing
+API] which would be again captured as spans by the agent and combining them along
+with long tasks would reveal the true source code location.
```
bmorelli25: This paragraph is a little tough for me. What about something like this:
```
With the help of the transaction timeline and span timings,
one could dig deeper by marking slow application code with the
https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark[User Timing API].
When these spans are then captured by the agent again, you could combine them
with long tasks to reveal the true source code location.
```

docs/advanced-topics.asciidoc
vigneshshanmugam: Don't really oppose your suggestion here. But shouldn't we start with explaining what is long task and then go ahead with why its intereting?.

PR #599, opened 2020-01-31
ci: ignoreHTTPSErrors check condition is inverted

kuisathaverat:
testing it on ITs I realized that the condition is inverted

### DIFF
--- a/dev-utils/integration-test.js
+++ b/dev-utils/integration-test.js
@@ -28,7 +28,7 @@ const puppeteer = require('puppeteer')
 async function runIntegrationTest(pageUrl) {
   const browser = await puppeteer.launch({
     headless: true,
-    ignoreHTTPSErrors: process.env.ELASTIC_APM_VERIFY_SERVER_CERT == 'true',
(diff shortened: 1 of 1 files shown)


dev-utils/integration-test.js
```diff
@@ -28,7 +28,7 @@ const puppeteer = require('puppeteer')
 async function runIntegrationTest(pageUrl) {
   const browser = await puppeteer.launch({
     headless: true,
-    ignoreHTTPSErrors: process.env.ELASTIC_APM_VERIFY_SERVER_CERT == 'true',
+    ignoreHTTPSErrors: process.env.ELASTIC_APM_VERIFY_SERVER_CERT == 'false',
```
vigneshshanmugam: are we setting this to `true` for other agents and only disabling it for RUM?

kuisathaverat: we set it to false when we use the self-signed certificate to avoid verified the certificate, the default value is true

https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-verify-server-cert

PR #600, opened 2020-02-03
feat(rum-core): add config option to monitor longtasks

vigneshshanmugam:
+ Adds a config option to monitor longtasks as part of all managed transaction
+ The config is set to `false` by default for now. I am trying to understand the performance impact of capturing longtasks and PerformanceObserver in general. will set it to true once I find some details. Right now we use a single PO at the start, but I am not sure if its the correct approach as the spec does not talk about reusability and also if we observer/disconnect using same observer, observing on entry types multiples times triggers warnings. We are still able to measure longtaks, but its safer now to disable it till we learn more.  #601

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -303,4 +303,4 @@ NOTE: Spans that are captured as part of the ignored transactions would also be 
 
 Instructs the agent to start monitoring for browser tasks that block the UI
 thread and might delay other user inputs by affecting the overall page
-responsiveness. Learn more about <<longtasks, Longtask spans>> and how to interpret them.
+responsiveness. Learn more about <<longtasks, long task spans>> and how to interpret them.
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -292,3 +292,15 @@ const options = {
 ----
 
 NOTE: Spans that are captured as part of the ignored transactions would also be ignored.
+
+
+[float]
+[[monitor-longtasks]]
+==== `monitorLongtasks`
+
+* *Type:* Boolean
+* *Default:* `false`
+
+Instructs the agent to start monitoring for browser tasks that block the UI
+thread and might delay other user inputs by affecting the overall page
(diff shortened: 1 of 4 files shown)


docs/configuration.asciidoc
```diff
@@ -291,4 +291,16 @@ const options = {
...
+[float]
+[[monitor-longtasks]]
+==== `monitorLongtasks`
+
+* *Type:* Boolean
+* *Default:* `false`
+
+Instructs the agent to start monitoring for browser tasks that block the UI
+thread and might delay other user inputs by affecting the overall page
+responsiveness. Learn more about <<longtasks, Longtask spans>> and how to interpret them.
```
vigneshshanmugam: docs build will fail till we merge the previous one #597

vigneshshanmugam: Pr is merged, should be good to go.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -758,15 +762,36 @@ describe('TransactionService', function() {
...
     })
+
+    it('should not record longtasks when monitorLongtasks is false', async () => {
+      config.setConfig({
+        monitorLongtasks: false
+      })
+      const pageLoadTr = trService.startTransaction('test', PAGE_LOAD, {
+        managed: true
+      })
+      expect(startSpy).toHaveBeenCalledTimes(1)
```
jahtalab: I think it would be better to recreate the spy in `beforeEach` for this `describe` set instead of reseting it every time. 

Also, in the `beforeEach` we can set the `recorder` object as well

vigneshshanmugam: We need to reset spies in the middle of tests as well https://github.com/elastic/apm-agent-rum-js/pull/600/files#diff-984832cc3eb80072b4dcfa9401a47af1L728 not just always in the end or the start, Would this change help us? 

I can move the resetSpeis to beforeEach, I just feel both would do the same job.

Issue #46, opened 2018-08-31
Limit the length for non-keyword fields


jahtalab:
Currently only the keyword fields are limited at 1024 characters. Need to apply a higher limit (possibly 10000 characters) to non-keyword fields as well.

PR #604, opened 2020-02-04
feat: capture click user-interaction transactions

### LINKED ISSUE #496: Capture user interactions

jahtalab:
Part of #496 

This PR captures click events as `user-interaction` transactions. 

image

### DIFF
--- a/dev-utils/webdriver.js
+++ b/dev-utils/webdriver.js
@@ -345,6 +345,14 @@ function getBrowserInfo() {
   }
 }
 
+function getBrowserFeatures() {
+  return browser.executeAsync(function(done) {
+    done({
+      EventTarget: !!window.EventTarget
+    })
+  })
+}
+
 function isChromeLatest() {
   const { name, version } = getBrowserInfo()
   const isChrome = name.indexOf('chrome') !== -1
@@ -360,5 +368,6 @@ module.exports = {
   isChromeLatest,
   getWebdriveBaseConfig,
   getBrowserInfo,
-  waitForApmServerCalls
+  waitForApmServerCalls,
+  getBrowserFeatures
 }
--- a/packages/rum-core/src/common/constants.js
+++ b/packages/rum-core/src/common/constants.js
@@ -66,6 +66,7 @@ const MAX_SPAN_DURATION = 5 * 60 * 1000
 const PAGE_LOAD = 'page-load'
 const ROUTE_CHANGE = 'route-change'
 const TYPE_CUSTOM = 'custom'
+const USER_INTERACTION = 'user-interaction'
 const HTTP_REQUEST_TYPE = 'http-request'
 const TEMPORARY_TYPE = 'temporary'
 const NAME_UNKNOWN = 'Unknown'
@@ -73,6 +74,7 @@ const NAME_UNKNOWN = 'Unknown'
 const TRANSACTION_TYPE_ORDER = [
   PAGE_LOAD,
   ROUTE_CHANGE,
+  USER_INTERACTION,
   HTTP_REQUEST_TYPE,
   TYPE_CUSTOM,
   TEMPORARY_TYPE
@@ -174,5 +176,6 @@ export {
   BROWSER_RESPONSIVENESS_BUFFER,
   SIMILAR_SPAN_TO_TRANSACTION_RATIO,
   TEMPORARY_TYPE,
+  USER_INTERACTION,
   TRANSACTION_TYPE_ORDER
 }
--- a/packages/rum-core/src/common/instrument.js
+++ b/packages/rum-core/src/common/instrument.js
@@ -23,7 +23,14 @@
  *
  */
 
-import { XMLHTTPREQUEST, FETCH, HISTORY, PAGE_LOAD, ERROR } from './constants'
+import {
+  XMLHTTPREQUEST,
+  FETCH,
+  HISTORY,
+  PAGE_LOAD,
+  ERROR,
+  EVENT_TARGET
+} from './constants'
 
 export function getInstrumentationFlags(instrument, disabledInstrumentations) {
   /**
@@ -34,7 +41,8 @@ export function getInstrumentationFlags(instrument, disabledInstrumentations) {
     [FETCH]: false,
     [HISTORY]: false,
     [PAGE_LOAD]: false,
-    [ERROR]: false
+    [ERROR]: false,
+    [EVENT_TARGET]: false
   }
 
   if (!instrument) {
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -100,16 +100,22 @@ class PerformanceMonitoring {
         task.source === EVENT_TARGET &&
         task.eventType === 'click'
       ) {
-        let target = task.target
-        let name = target.getAttribute('name')
+        const target = task.target
+        const name = target.getAttribute('name')
 
         let additionalInfo = ''
         if (name) {
           additionalInfo = `["${name}"]`
         }
 
-        let tagName = target.tagName.toLowerCase()
-        let tr = transactionService.startTransaction(
+        const tagName = target.tagName.toLowerCase()
+
+        /**
+         * We reduce the reusability threshold to make sure
+         * we only capture async activities (e.g. network)
+         * related to this interaction.
+         */
+        const tr = transactionService.startTransaction(
           `Click >> ${tagName}${additionalInfo}`,
           USER_INTERACTION,
           {
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -41,7 +41,9 @@ import {
   FETCH,
   HISTORY,
   XMLHTTPREQUEST,
+  EVENT_TARGET,
   HTTP_REQUEST_TYPE,
+  USER_INTERACTION,
   BROWSER_RESPONSIVENESS_INTERVAL,
   BROWSER_RESPONSIVENESS_BUFFER,
   SIMILAR_SPAN_TO_TRANSACTION_RATIO
@@ -83,6 +85,42 @@ class PerformanceMonitoring {
 
     if (flags[FETCH]) {
       patchEventHandler.observe(FETCH, this.getFetchSub())
+    }
+
+    if (flags[EVENT_TARGET]) {
+      patchEventHandler.observe(EVENT_TARGET, this.getEventTargetSub())
+    }
+  }
+
+  getEventTargetSub() {
+    const transactionService = this._transactionService
+    return (event, task) => {
+      if (
+        event === SCHEDULE &&
+        task.source === EVENT_TARGET &&
+        task.eventType === 'click'
+      ) {
+        let target = task.target
+        let name = target.getAttribute('name')
+        let classes = target.getAttribute('class')
+
+        let additionalInfo = ''
+        if (name) {
+          additionalInfo = `["${name}"]`
+        } else if (classes) {
+          classes = classes.split(' ').join('.')
+          additionalInfo = `.${classes}`
+        }
+        let tagName = target.tagName.toLowerCase()
+        transactionService.startTransaction(
+          `Click >> ${tagName}${additionalInfo}`,
+          USER_INTERACTION,
+          {
+            managed: true,
+            canReuse: true
+          }
+        )
+      }
     }
   }
 
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -102,24 +102,29 @@ class PerformanceMonitoring {
       ) {
         let target = task.target
         let name = target.getAttribute('name')
-        let classes = target.getAttribute('class')
 
         let additionalInfo = ''
         if (name) {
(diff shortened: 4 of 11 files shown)


dev-utils/webdriver.js
```diff
@@ -345,6 +345,14 @@ function getBrowserInfo() {
   }
 }
 
+function getBrowserFeatures() {
```
vigneshshanmugam: nit: getSupportedFeatures or getSupportedBrowserFeatures would be better.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -84,6 +86,42 @@ class PerformanceMonitoring {
...
+    const transactionService = this._transactionService
+    return (event, task) => {
+      if (
+        event === SCHEDULE &&
+        task.source === EVENT_TARGET &&
+        task.eventType === 'click'
+      ) {
+        let target = task.target
+        let name = target.getAttribute('name')
+        let classes = target.getAttribute('class')
```
vigneshshanmugam: Using class might not be scalable as there are lots of conventions and if the websites uses css modules or something similar concept, those would be `autogenarated` from the bundler (Ex: webpack) and it wont help the user in any way. 

One idea would be to figure out where the element is present on the page, something like `xpath`. Example from github `/html/body/div[1]/header/div[5]/details/details-menu/div[3]/div/details/details-dialog/form/div[1]/button`, We can also use it in context than having it as long name.

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -796,4 +797,35 @@ describe('PerformanceMonitoring', function() {
 
     tr.end()
   })
+
+  if (window.EventTarget) {
```
vigneshshanmugam: Need a test for this scenario click triggers a history `pushstate` change and have to check transaction details expected.

packages/rum-core/src/common/constants.js
```diff
@@ -66,13 +66,15 @@ const MAX_SPAN_DURATION = 5 * 60 * 1000
...
 const TYPE_CUSTOM = 'custom'
+const USER_INTERACTION = 'user-interaction'
 const HTTP_REQUEST_TYPE = 'http-request'
 const TEMPORARY_TYPE = 'temporary'
 const NAME_UNKNOWN = 'Unknown'
 
 const TRANSACTION_TYPE_ORDER = [
   PAGE_LOAD,
   ROUTE_CHANGE,
+  USER_INTERACTION,
```
vigneshshanmugam: Clicks would trigger a route change, not the other way around. I am not 100% sure if this order would work out for us. WDYT?

packages/rum-core/src/common/constants.js
```diff
@@ -66,13 +66,15 @@ const MAX_SPAN_DURATION = 5 * 60 * 1000
 const PAGE_LOAD = 'page-load'
 const ROUTE_CHANGE = 'route-change'
 const TYPE_CUSTOM = 'custom'
+const USER_INTERACTION = 'user-interaction'
```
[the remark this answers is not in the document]
jahtalab: Yeah, it shows up in the UI.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -84,6 +86,42 @@ class PerformanceMonitoring {
...
+  getEventTargetSub() {
+    const transactionService = this._transactionService
+    return (event, task) => {
+      if (
+        event === SCHEDULE &&
+        task.source === EVENT_TARGET &&
+        task.eventType === 'click'
+      ) {
+        let target = task.target
+        let name = target.getAttribute('name')
```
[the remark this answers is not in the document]
jahtalab: Here, we're getting the `name` attribute of the element (if present) not the `tagName`, e.g. `<button name="test"/>`.

jahtalab:
@vigneshshanmugam ,

Re. filtering transactions, when we removed the condition to only send transactions without spans we also decided to add additional conditions to filter out unimportant transaction I think we should do this for all transactions separately. I couldn't find an issue for it so I created #608.

packages/rum-core/src/common/constants.js
jahtalab: This order means that a `user-interaction` transaction can be redefined as `route-change` but not the other way around. That seems to me the way it should be, Why do you think this doesn't work?

packages/rum-core/src/performance-monitoring/performance-monitoring.js
jahtalab: To be honest I only added the classes as a back up but not entirely happy about it because of the issues that you mentioned. But, I also don't think the xpath is good approach since it will clutter the UI. However, the classes are only a backup to the name attribute, so there's a way for users who can't use classes to fix their elements to have a name.

packages/rum-core/src/common/instrument.js
```diff
@@ -34,7 +41,8 @@ export function getInstrumentationFlags(instrument, disabledInstrumentations) {
     [FETCH]: false,
     [HISTORY]: false,
     [PAGE_LOAD]: false,
-    [ERROR]: false
+    [ERROR]: false,
+    [EVENT_TARGET]: false
```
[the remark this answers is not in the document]
jahtalab: Added a todo to the original issue.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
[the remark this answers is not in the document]
jahtalab: click event listeners are not fired for disabled elements

dev-utils/webdriver.js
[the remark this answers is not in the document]
jahtalab: Not sure though, I prefer having 'browser' in the name since this is suppose to run in the context of the browser. Also, this returns an object containing for the features but those features might be supported or not (depending on the value)

jahtalab:
Meeting summary:
- We will add classes and xpath to the context of the transaction (instead of the name).
- Re. route change vs user-interaction, we prioritise the route-change in all cases even if the routing library is not instrumented
- We should disable the user interaction feature until #608 is fixed
- We should set reusability threshold to 1000ms only for click transactions

jahtalab:
@vigneshshanmugam , I decided against adding xpath for now but we can add it if there's a request for it in the future. Also, I've added a filter to discard managed transactions with no spans.

packages/rum-core/test/performance-monitoring/transaction.spec.js
```diff
@@ -144,5 +144,13 @@ describe('transaction.Transaction', function() {
 
     transaction._start = transaction._start - 10000
     expect(transaction.canReuse()).toBe(false)
+
+    transaction = new Transaction('transaction', 'transaction', {
+      canReuse: true,
+      reuseThreshold: 100
```
vigneshshanmugam: passing in `startTime`: 0 in here and changing transaction._start = 101 in next line would make this easier to understand

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -102,24 +102,29 @@ class PerformanceMonitoring {
...
           {
             managed: true,
-            canReuse: true
+            canReuse: true,
+            reuseThreshold: 100
           }
         )
+
+        if (tr) {
+          let classes = target.getAttribute('class')
```
vigneshshanmugam: Thought about this once again, Its good idea to truncate this or not add this if the length of the classnames exceed a certain threshold. WDYT about like this
```
target.classList.value.length > 250 // dont add the context
```

This is a guess that, user is relying on auto generation and wont be of any help. Atleast we are reducing the payload size

robgil:
# [Codecov](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/604?src=pr&el=h1) Report
> Merging [#604](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/604?src=pr&el=desc) into [master](https://codecov.io/gh/elastic/apm-agent-rum-js/commit/a492f290d39ab40adb24a5f494b402dd08a7f10a&el=desc) will **increase** coverage by `0.35%`.
> The diff coverage is `n/a`.



```diff
@@            Coverage Diff             @@
##           master     #604      +/-   ##
==========================================
+ Coverage   93.90%   94.26%   +0.35%     
==========================================
  Files          47       47              
  Lines        2231     2475     +244     
  Branches      448      542      +94     
==========================================
+ Hits         2095     2333     +238     
- Misses        133      139       +6     
  Partials        3        3              
```

| [Impacted Files](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/604?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [packages/rum-core/src/common/config-service.js](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/604/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWNvcmUvc3JjL2NvbW1vbi9jb25maWctc2VydmljZS5qcw==) | `96.94% <0.00%> (-0.73%)` | :arrow_down: |
| [packages/rum-core/src/common/constants.js](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/604/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWNvcmUvc3JjL2NvbW1vbi9jb25zdGFudHMuanM=) | `100.00% <0.00%> (ø)` | :arrow_up: |
| [packages/rum-core/src/common/instrument.js](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/604/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWNvcmUvc3JjL2NvbW1vbi9pbnN0cnVtZW50Lmpz) | `100.00% <0.00%> (ø)` | :arrow_up: |
| [...c/performance-monitoring/performance-monitoring.js](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/604/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWNvcmUvc3JjL3BlcmZvcm1hbmNlLW1vbml0b3JpbmcvcGVyZm9ybWFuY2UtbW9uaXRvcmluZy5qcw==) | `96.25% <0.00%> (+1.18%)` | :arrow_up: |

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -827,5 +843,37 @@ describe('PerformanceMonitoring', function() {
...
       expect(newTr.name).toBe('Click >> button["purchase"]')
     })
+
+    it('should respect the transaction type priority order', function() {
+      const historySubFn = performanceMonitoring.getHistorySub()
+      const cancelHistorySub = patchEventHandler.observe(HISTORY, historySubFn)
+      let etsub = performanceMonitoring.getEventTargetSub()
+      const cancelEventTargetSub = patchEventHandler.observe(
+        EVENT_TARGET,
+        (event, task) => {
```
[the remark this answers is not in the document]
jahtalab: It doesn't work with a stub we need to call the observer callback.

packages/rum-core/test/performance-monitoring/transaction.spec.js
jahtalab: This doesn't work, we want to make the start go back not forward.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
jahtalab: Let's added a generic truncation for all custom contexts separately.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -102,24 +102,29 @@ class PerformanceMonitoring {
...
         let tagName = target.tagName.toLowerCase()
-        transactionService.startTransaction(
+        let tr = transactionService.startTransaction(
           `Click >> ${tagName}${additionalInfo}`,
           USER_INTERACTION,
           {
             managed: true,
-            canReuse: true
+            canReuse: true,
+            reuseThreshold: 100
```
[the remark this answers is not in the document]
jahtalab: Yeah, let's do it separately since it touches multiple packages.

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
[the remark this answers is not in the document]
vigneshshanmugam: I meant this, `patchEventHandler.observe(EVENT_TARGET, etsub)`

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: Custom contexts doesn't require truncation as far as I remember, This is a edge case and we can handle it now in this PR or atleast in follow up PR as it creates noise since the feature is enabled by default. Can you add a task to the original isssue.

jahtalab: To be honest, I'm not sure what exactly we're trying to solve here, even the autogenerated classes in most cases are far less than 250, so to avoid them we would have to set the limit much lower than this. 

Re. generic custom context limit, we already have an issue for it: #46

PR #616, opened 2020-02-12
fix: pass correct properties to payload filters

jahtalab:
This bug was introduced around 4.0.2 release and therefore affects nearly all 4.x versions. I've added more tests to cover this case. I've also removed the undocumented `service` property from the payload passed to filters.

### DIFF
--- a/packages/rum-core/test/common/apm-server.spec.js
+++ b/packages/rum-core/test/common/apm-server.spec.js
@@ -423,6 +423,25 @@ describe('ApmServer', function() {
     expect(result).toEqual([expected])
   })
 
+  it('should pass correct payload to filters', () => {
+    configService.setConfig({
+      serviceName: 'serviceName'
+    })
+    let type = ''
+    configService.addFilter(function(payload) {
+      expect(payload[type]).toBeDefined()
+      return payload
+    })
+
+    type = 'errors'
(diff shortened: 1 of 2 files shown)


packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -423,6 +423,25 @@ describe('ApmServer', function() {
...
+      serviceName: 'serviceName'
+    })
+    let type = ''
+    configService.addFilter(function(payload) {
+      expect(payload[type]).toBeDefined()
+      return payload
+    })
+
+    type = 'errors'
+    let result = apmServer.sendErrors([{ test: 'test' }])
```
vigneshshanmugam: you would have to use await since sendErrors returns a promise. Might be hard to catch these bugs since errors would be logged later on.

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -423,6 +423,25 @@ describe('ApmServer', function() {
...
+    expect(result).toBeDefined()
+
+    type = 'transactions'
+    result = apmServer.sendTransactions([{ test: 'test' }])
```
vigneshshanmugam: same here.

PR #617, opened 2020-02-12
feat: add labels config to metadata payload

### LINKED ISSUE #584: Add global labels to meta payload document
Starting from APM server 7.6, global labels are added to spans as well. We should add the global labels in the meta document for this to happen though.

jahtalab:
Fixes #584


jahtalab:
@vigneshshanmugam, We can only remove labels (not all context) from transactions and errors. I've created an [issue for this](https://github.com/elastic/apm-agent-rum-js/issues/618) since it should be done as a breaking change.

jahtalab:
jenkins run tests please!

Issue #477, opened 2019-11-11
Support full TraceContext spec


jahtalab:
See https://github.com/elastic/apm/issues/71 for more.

- [x] Document the compatibility matrix for backend agents (#645)
- [x] Document the `traceparent` header for CORS configuration. (#645)
- [x] Provide the Node.js fix (https://github.com/elastic/apm-agent-nodejs/pull/1587)

jahtalab:
We should add the config option in 7.6 but keep it the same as today and change it to in the next major version after 7.6 release

PR #626, opened 2020-02-17
BREAKING CHANGE: use the official traceparent header

jahtalab:
Part of #477
This PR uses the official W3C tracecontext `traceparent` header instead `elastic-apm-traceparent` which we previously used.


jahtalab:
jenkins run the tests please

Issue #621, opened 2020-02-14
Review bundles size check and adjust if necessary


jahtalab:
Currently the bundle size check on CI fails since we have increase the size, we should review the analyser report and make improvements if possible. Otherwise we should adjust the size check.

vigneshshanmugam:
since we have a dedicated issue for refactoring `ServiceFactory` #238, closing this issue as that would be the next big step.

PR #629, opened 2020-02-17
feat(rum-core): use global promise when available and fallback

vigneshshanmugam:
+ part of #https://github.com/elastic/apm-agent-rum-js/issues/621
+ Since Promise is supported natively by most of the browsers except for IE 11, we are using a polyfill that is less than 1kb gzipped compared to es-promise which was 2.4kb gzipped. 
+ This PR also addresses one core issue where we use the global Promise when available and only use fallback when promise is not supported. This would also fix the issue in tests where we don't need to import Promises since we call global polyfills before tests are run. 

## Bundle size
Filename | Before | After 
-------- | --------| ---------
elastic-apm-js-base.min.js |  18.23KB  | **17.08KB**
elastic-apm-opentracing.umd.min.js | 19.71KB | **18.55KB**


vigneshshanmugam:
Trying to figure out a scheduling issue when `ZoneAwarePromise` comes in to play on Angular.

packages/rum-core/package.json
```diff
@@ -37,8 +37,8 @@
   },
   "dependencies": {
     "error-stack-parser": "^1.3.5",
-    "es6-promise": "^4.2.8",
     "opentracing": "^0.14.3",
+    "promise-polyfill": "^8.1.3",
```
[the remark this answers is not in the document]
vigneshshanmugam: Yes, otherwise if user installs our agent, it would break.

vigneshshanmugam:
@jahtalab I explained the reasoning on why it cannot be a constant, It needs to be a function so we would always use the consistent Promise even if its globally polyfilled by the user.

jahtalab:
@vigneshshanmugam , Not sure what you're referring to in regards to explanation, can you elaborate?

vigneshshanmugam:
@jahtalab https://github.com/elastic/apm-agent-rum-js/pull/629/files#diff-1353fc7f8249e73fab2adcaef339cb7bR33-R40

jahtalab:
I don't think it's good enough reason to change the use of Promise the way it's suggested in this PR. It also seems to me using a function actually introduces more inconsistencies since it becomes important when we call the function and if promise polyfill changes we would see inconsistent errors. A better solution would be to document the fact that if users load their polyfill before the agent we do use that instead.

packages/rum-angular/test/e2e/with-router/app.e2e-spec.js
```diff
@@ -48,6 +48,15 @@ describe('Angular router integration', function() {
       'expected contact list to be rendered'
     )
 
+    /**
+     * TODO: There are differences in how micro task is schedluled in promise compatible
```
vigneshshanmugam: made an issue here - https://github.com/elastic/apm-agent-rum-js/issues/636

packages/rum-angular/src/apm-service.ts
```diff
@@ -101,11 +101,20 @@ export class ApmService {
         }
         /**
          * Schedule the transaction finish logic on the next
-         * micro task since most of the angular components wait for
+         * micro task queue ssince most of the angular components wait for
```
jahtalab: nit: `since`

packages/rum-angular/src/apm-service.ts
```diff
@@ -101,11 +101,20 @@ export class ApmService {
...
          * Observables resolution on ngInit to fetch the neccessary
          * data for mounting
+         *
+         * TODO: There are subtle differences in how micro task is schedluled while
+         * usisng promise polyfill libraries and browsers, Native promises run
```
jahtalab: nit: `using`

Bamieh:
# [Codecov](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/629?src=pr&el=h1) Report
> Merging [#629](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/629?src=pr&el=desc) into [master](https://codecov.io/gh/elastic/apm-agent-rum-js/commit/2ef506ab1fc3a92d643e96edf0e226ac39b238e4&el=desc) will **decrease** coverage by `0.10%`.
> The diff coverage is `n/a`.



```diff
@@            Coverage Diff             @@
##           master     #629      +/-   ##
==========================================
- Coverage   94.13%   94.02%   -0.11%     
==========================================
  Files          47       48       +1     
  Lines        2181     2194      +13     
  Branches      432      436       +4     
==========================================
+ Hits         2053     2063      +10     
- Misses        125      128       +3     
  Partials        3        3              
```

| [Impacted Files](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/629?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [.../src/performance-monitoring/transaction-service.js](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/629/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWNvcmUvc3JjL3BlcmZvcm1hbmNlLW1vbml0b3JpbmcvdHJhbnNhY3Rpb24tc2VydmljZS5qcw==) | `90.47% <0.00%> (-0.60%)` | :arrow_down: |
| [packages/rum-core/src/index.js](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/629/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWNvcmUvc3JjL2luZGV4Lmpz) | `100.00% <0.00%> (ø)` | :arrow_up: |
| [packages/rum-core/src/common/polyfills.js](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/629/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWNvcmUvc3JjL2NvbW1vbi9wb2x5ZmlsbHMuanM=) | `66.66% <0.00%> (ø)` | |
| [packages/rum-angular/src/apm-service.ts](https://codecov.io/gh/elastic/apm-agent-rum-js/pull/629/diff?src=pr&el=tree#diff-cGFja2FnZXMvcnVtLWFuZ3VsYXIvc3JjL2FwbS1zZXJ2aWNlLnRz) | `84.37% <0.00%> (+0.50%)` | :arrow_up: |

vigneshshanmugam:
Will fix the `ss` problems in next PR.

PR #630, opened 2020-02-18
feat(rum-core): log when dt header isnt injected

jahtalab:


packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -198,6 +198,10 @@ class PerformanceMonitoring {
       const target = data.target
       if (isDtEnabled && isSameOrigin && target) {
         this.injectDtHeader(span, target)
+      } else if (__DEV__) {
+        this._logginService.debug(
+          `Could not inject distributed tracing header to the request ('${requestUrl.origin}') from the current origin ('${currentUrl.origin}')`
```
vigneshshanmugam: since we are calling it ass request, it should be the full path or we should called it as `origin`

Issue #640, opened 2020-02-19
Test pipeline optimisation


jahtalab:
- Move `Coverage` job to the last
- Run "Integration test" job in parallel with other tests to reduce the test time.

cc @elastic/observablt-robots

PR #641, opened 2020-02-19
ci(jenkins): test pipeline optimisation

v1v:
### Highlights
- Run APM-ITs after running the tests.
- Run Coverage at the very end

The `Integration Tests` stage cannot be easily triggered in parallel with the `Test` stage. So there were two options, either run the ITs firstly and then the Test stage or the other way around. As agreed we will run after the `Test` stage.

Closes https://github.com/elastic/apm-agent-rum-js/issues/640

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,7 @@ pipeline {
     GITHUB_CHECK_ITS_NAME = 'Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/master'
     OPBEANS_REPO = 'opbeans-frontend'
+    PATH = "${env.PATH}:${env.WORKSPACE}/bin"
   }
   options {
     timeout(time: 3, unit: 'HOURS')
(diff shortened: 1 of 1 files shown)


v1v:
> Would you please elaborate a bit on the challenges with running Integration in parallel, I'm asking because from the tests alone they don't interfere with each other.

Sure thing.

From the implementation point of view, the parallel test stages use a different specific implementation to iterate through the files `.ci/.jenkins_*.yaml`, so the complexity to add the `Integration Tests` in parallel it's not worth from the maintainability point of view, in other words, I might need to hack with some internal Jenkins implementation. 

> it's better to keep it after the Test stage, and also maybe we can run Coverage and Integration in parallel then.

Neat! That's indeed a good approach :)

v1v:
@jahtalab the just committed changes are equivalent to:

```
... 
parallel {
    stage('coverage')
    stage('Integration Tests')
}
```

So, I don't think we need the parallel as long as the ITs are executed without the `waitFor` flag.

v1v:
@vigneshshanmugam 

> We kick off IT in other repo so its kinda non-blocking (but we still block till the other Jenkins job kicks in before we start our own Tests run) but would help speed up the test time overall.

A quick update regarding the implementation, the Tests run happens before the `ITs`. The `ITs` run doesn't block anything in the pipeline, and the overall speed should be as you noticed a bit faster. I just updated the descriptions too.

Issue #628, opened 2020-02-17
Remove unnecessary config options


jahtalab:
We should apply these only to managed transactions and remove the config option
- `checkBrowserResponsiveness`
- `groupSimilarSpans`
- `similarSpanThreshold`
- `transactionDurationThreshold` 

We should double check these with backend:
- `errorThrottleLimit`
- `errorThrottleInterval`
- `transactionThrottleLimit`
- `transactionThrottleInterval`
- `debug` -> we already have the `logLevel` so there's no need for this.
- `queueLimit` -> We might need this config if we support offline

vigneshshanmugam:
### Findings

+ **debug** - Can be safely removed as most of the agents except for python, already removed it and this option is not supported for remote config. 

+ `flushInterval` and `queueLimit` are replaced with `apiRequestTime` and `apiRequestSize` in backend agents. We might not need to follow the same since out connection doesnt stay open and our size might differ with backend agents since gzip/no-gzip. Have to come to an agreement on these

jahtalab:
We decided to keep `flushInterval` as it is today since, `apiRequestTime` supports a different format (e.g. with units) and that's not what we want to support at this point. `apiRequestSize` is not relevant for us since we send the transaction and all of it's spans within the same request.

We will keep `queueLimit` for now, but decided to merge the 4 throttling configs to only one config. This config will apply to both transactions and Errors and it will have a fixed interval (e.g. 1 minute), it can be called something like `eventPerMin` (or similar)

vigneshshanmugam:
### Subtasks

+ [ ] Remove Debug option from the code #670 
+ [ ] Single queue for reporting events #664 
+ [ ] Update docs with `eventsPerMin` and remove throttleInterval and throttleLimit.

PR #642, opened 2020-02-19
refactor(rum-core):  managed transaction configs are kept as constants

vigneshshanmugam:
+ part of #628 
+ Moved instance methods to separate functions since they were relying on any of the instance properties (better minification)
+ Moved some logic around 
     -  `resetSpans` first in the callstack so we dont need to perform the grouping and other filtering when we already know its unsampled.
     - Sorting is only done for managed transactions
     - prepare transactions would only touch the transaction span when its done processing to keep it functional
+ Tests are grouped under PerformanceUtilis to make it easier to digest the changes
+ Size benefits was around `0.09kb` for both bundles.


packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -262,74 +353,47 @@ class PerformanceMonitoring {
...
       }
-    }
-    return true
-  }
 
-  prepareTransaction(transaction) {
-    transaction.spans.sort(function(spanA, spanB) {
-      return spanA._start - spanB._start
-    })
+      if (tr.options.checkBrowserResponsiveness) {
```
jahtalab: We can remove this option here, since we're only applying it to managed transactions and there's no config option anymore.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -372,58 +436,10 @@ class PerformanceMonitoring {
   }
 
   createTransactionPayload(transaction) {
-    this.prepareTransaction(transaction)
-    const filtered = this.filterTransaction(transaction)
+    const preparedTr = prepareTransaction(transaction)
```
jahtalab: This could be a bit misleading since we still change the transaction itself, I don't think we should return the transaction but maybe we can change the function name to indicate it changes the transaction itself.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -119,7 +119,7 @@ class TransactionService {
           pageLoadSampled: config.pageLoadSampled,
           pageLoadSpanId: config.pageLoadSpanId,
           pageLoadTransactionName: config.pageLoadTransactionName,
-          checkBrowserResponsiveness: config.checkBrowserResponsiveness
+          checkBrowserResponsiveness: true
```
jahtalab: We can just remove this here since it's not a config option anymore.

packages/rum-core/src/common/constants.js
```diff
@@ -175,6 +180,7 @@ export {
   BROWSER_RESPONSIVENESS_INTERVAL,
   BROWSER_RESPONSIVENESS_BUFFER,
   SIMILAR_SPAN_TO_TRANSACTION_RATIO,
+  TRANSACTION_DURATION_THRESHOLD,
```
jahtalab: We can move this to performance monitoring file since it's only use there AFAICS.

packages/rum-core/test/benchmarks/transaction-service.bench.js
```diff
@@ -28,15 +28,12 @@ import { createServiceFactory } from '../../src/index'
...
-  configService.setConfig({
-    checkBrowserResponsiveness: false
-  })
 
   benchmark('page-load transaction overhead', () => {
     const tr = transactionService.startTransaction('/index', 'page-load', {
-      managed: true
+      managed: true,
+      // To avoid timeouts interfering with benchmark results
+      checkBrowserResponsiveness: false
```
jahtalab: Do we still need this option? This seems to be the only place that uses this option can we do something else here?

packages/rum-core/src/common/config-service.js
```diff
@@ -76,15 +76,11 @@ class Config {
       debug: false,
       logLevel: 'warn',
       breakdownMetrics: false,
-      checkBrowserResponsiveness: true,
-      groupSimilarSpans: true,
-      similarSpanThreshold: 0.05,
```
jahtalab: Did we forget to remove this previously ?

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: But,we don't want it to run for `page-load` right? So i kept it.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: Hmmm, yeah since deals mostly with the `spans`, we can call it as `adjustSpans` or `prepareSpans`?

packages/rum-core/src/common/constants.js
vigneshshanmugam: If we do that, we can also move all of the 4 there. I am okay with both.

packages/rum-core/src/common/config-service.js
vigneshshanmugam: Yes 🤦‍♂just the option.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: Lets think of a way, we can either check type and not run the check or check if `browserResponsivenessCounter` is > 0.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -55,7 +53,103 @@ import {
...
+}
+
+export function adjustTransactionSpans(transaction) {
+  /**
+   * In case of unsampled transaction, send only the transaction to apm server
+   *  without any spans to reduce the payload size
+   */
+  if (!transaction.sampled) {
+    transaction.resetSpans()
+  }
```
jahtalab: We can expand this if here like:

```
if(transaction.sampled) {
....
} else {
    transaction.resetSpans()
}
return transaction
```

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -262,74 +353,50 @@ class PerformanceMonitoring {
...
-
-    if (tr.options.checkBrowserResponsiveness) {
-      const wasBrowserResponsive = this.checkBrowserResponsiveness(
-        tr,
-        BROWSER_RESPONSIVENESS_INTERVAL,
-        BROWSER_RESPONSIVENESS_BUFFER
-      )
 
-      if (!wasBrowserResponsive) {
+      if (tr.spans.length === 0) {
```
jahtalab: Since you moved the `transaction.resetSpans()` inside `adjustTransactionSpans` this check here will prevent upsampled transactions to be sent, we should apply this check only to sampled transactions. We probably need a test as well

vigneshshanmugam: Nice catch, will fix it and add test as well.

Issue #620, opened 2020-02-14
Tests review


jahtalab:
We should review all tests and make improvements that might be necessary.

jahtalab:
@vigneshshanmugam , I'm reviewing tests. Do you have any suggestions?

jahtalab:
- Fix logs showing up after an individual test has finished
- Fix the sourcemap so that stack trace show correct lines.

jahtalab:
We should look into `jasmine.random` option since running in random order breaks our tests: https://github.com/elastic/apm-agent-rum-js/blob/master/dev-utils/karma.js#L73

PR #648, opened 2020-02-25
chore(tests): review tests and add utils

jahtalab:
Part of #620

### DIFF
--- a/dev-utils/jasmine.js
+++ b/dev-utils/jasmine.js
@@ -1,0 +1,68 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2017-present, Elasticsearch BV
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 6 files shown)


dev-utils/jasmine.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Conditional describe
+ * @param {*} description
+ * @param {*} specDefinitions
+ * @param {*} condition
+ */
+export function describeIf(description, specDefinitions, condition) {
+  let describeFn = describe
+  if (arguments.length > 2) {
```
vigneshshanmugam: merge two if's

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -461,51 +460,56 @@ describe('ApmServer', function() {
...
+          transaction_sample_rate: '0.5',
+          etag: 'test'
+        })
+        sessionStorage.removeItem(LOCAL_CONFIG_KEY)
       })
-      sessionStorage.removeItem(LOCAL_CONFIG_KEY)
-    })
-  }
+    },
+    !testConfig.stackVersion ||
```
vigneshshanmugam: any reason why we need to run tests when stackVersion is undefined?

jahtalab: We should run as many tests as we can, so it's better to run all tests and if we want to exclude these we can always explicitly pass the stack version to the command line (e.g. in our local testing)

dev-utils/jasmine.js
jahtalab: Let's merge this one, I'll fix it in the next PR that I'm working on.

PR #646, opened 2020-02-24
Ensure window isnt used on import, to prevent server side rendered envs from crashing

jooj123:
Ensure window isnt used on import, to prevent server side rendered envs from crashing with message:

```
ReferenceError: window is not defined
    at eval (webpack-internal:///./node_modules/@elastic/apm-rum-core/dist/es/common/utils.js:43:12)
    at Module../node_modules/@elastic/apm-rum-core/dist/es/common/utils.js (/mypath/build/server/main.js:7920:1)
    at __webpack_require__ (/mypath/build/server/main.js:21:30)
    at eval (webpack-internal:///./node_modules/@elastic/apm-rum-core/dist/es/error-logging/error-logging.js:3:71)
    at Module../node_modules/@elastic/apm-rum-core/dist/es/error-logging/error-logging.js (/mypath/build/server/main.js:7944:1)
    at __webpack_require__ (/mypath/build/server/main.js:21:30)
    at eval (webpack-internal:///./node_modules/@elastic/apm-rum-core/dist/es/error-logging/index.js:2:72)
    at Module../node_modules/@elastic/apm-rum-core/dist/es/error-logging/index.js (/mypath/build/server/main.js:7956:1)
    at __webpack_require__ (/mypath/build/server/main.js:21:30)
    at eval (webpack-internal:///./node_modules/@elastic/apm-rum-core/dist/es/index.js:3:72)
    at Module../node_modules/@elastic/apm-rum-core/dist/es/index.js (/mypath/build/server/main.js:7980:1)
    at __webpack_require__ (/mypath/build/server/main.js:21:30)
    at eval (webpack-internal:///./node_modules/@elastic/apm-rum/dist/es/bootstrap.js:3:79)
    at Module../node_modules/@elastic/apm-rum/dist/es/bootstrap.js (/mypath/build/server/main.js:8159:1)
    at __webpack_require__ (/mypath/build/server/main.js:21:30)
    at eval (webpack-internal:///./node_modules/@elastic/apm-rum/dist/es/index.js:5:68)
```

It was because window.performance was being accessed on `import` even though the elastic lib functions were not executed on server render

PR #651, opened 2020-02-25
chore(dev): fix uncaughtException handler from consuming test errors

### LINKED ISSUE #649: `uncaughtException` handler is consuming the errors in tests
+ If any exception is thrown from the test, our uncaught exception handler is consuming the error and returns the process exitcode as 0. 

More details #646

vigneshshanmugam:
+ fix #649 
+ Overriding `uncaughtException` would result in exitCode resulting in 0 and consuming our test errors. 
+ process.exit(exitcode) is unnecessary as the process would be exited anyways when error happens. 

This PR would result in node tests being broken because of this issue #646.

### DIFF
--- a/dev-utils/run-script.js
+++ b/dev-utils/run-script.js
@@ -169,11 +169,9 @@ function exitHandler(exitCode) {
     }
   })
   cleanUps = []
-  process.exit(exitCode)
 }
 
 process.on('exit', exitHandler)
-process.on('uncaughtException', exitHandler)
 process.on('SIGINT', exitHandler)
 
 const scripts = {


vigneshshanmugam:
Because we are overriding the default handler of Node.js you can read more here - https://nodejs.org/api/process.html#process_event_uncaughtexception

In any case,  this change is much better than other alternatives like 
+ keeping the `uncaughtException` handler and setting process.exitCode manually
+ writing to `stderr` synchronously in uncaughtException handler.

jahtalab:
Let's discuss this a bit more, I think we need to run the clean up process in the `uncaughtException ` event listener and furthermore we exit the process with the right code.

vigneshshanmugam:
jenkins run the tests please

PR #652, opened 2020-02-25
chore(bundlesize): add github workflows to check compressed size

vigneshshanmugam:
+ trying out the compressed size plugin to know size per file. 
https://github.com/marketplace/actions/compressed-size-action

Might delete it if its not super useful for us.


vigneshshanmugam:
I am not sure though, wasnt able to test it. We might need to merge and try in other PR.

PR #593, opened 2020-01-28
chore(benchmark): run benchmarks on chromium, firefox and webkit

vigneshshanmugam:
+ Just playing around and understanding how our agent performs in different browsers. 
+ uses the https://github.com/microsoft/playwright to run the benchmarks instead of puppeteer. 
+ Puppeteer now also released v2.1.0 which supports launching firefox. https://github.com/puppeteer/puppeteer/releases/tag/v2.1.0

PR #653, opened 2020-02-26
ci(jenkins): enable benchmark for PRs if changes in the benchmark files

v1v:
## What

Enable the benchmark without sending the data to ES for the pull requests that change any of the files that are directly related to the benchmark.

## Why

This will help to automatically validate any changes that could affect the benchmark implementation before merging to master.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,7 +63,7 @@ pipeline {
             script {
               dir("${BASE_DIR}"){
                 def regexps =[
-                  '^packages/rum/test/benchmarks/',
+                  '^packages/.*/test/benchmarks/.*',
                   '^scripts/benchmarks.js',
                   '^packages/rum-core/karma.bench.conf.js'
                 ]
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,6 +59,17 @@ pipeline {
             deleteDir()
             gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true)
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+            // Look for changes related to the benchmark, if so then set the env variable.
+            script {
+              dir("${BASE_DIR}"){
+                def regexps =[
+                  "^packages/rum/test/benchmarks/",
+                  "^scripts/benchmarks.js$",
+                  "^packages/rum-core/karma.bench.conf.js$"
+                ]
+                env.BENCHMARK_UPDATED = isGitRegionMatch(patterns: regexps)
+              }
+            }
           }
         }
         /**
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -59,6 +59,17 @@ pipeline {
             deleteDir()
             gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true)
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+            // Look for changes related to the benchmark, if so then set the env variable.
+            script {
+              dir("${BASE_DIR}"){
+                def regexps =[
+                  '^packages/rum/test/benchmarks/',
+                  '^scripts/benchmarks.js',
+                  '^packages/rum-core/karma.bench.conf.js'
```
vigneshshanmugam: also this `packages/rum-core/test/benchmarks`, we can have it as `packages/**/test/benchmarks/`

Jenkinsfile
```diff
@@ -59,6 +59,17 @@ pipeline {
...
+                ]
```
v1v: This is the list of files that were changed in the PR https://github.com/elastic/apm-agent-rum-js/pull/593

Jenkinsfile
```diff
@@ -129,6 +140,7 @@ pipeline {
                 branch 'master'
```
mdelapenya: Question: what's the purpose of an `allOf ( anyOf (a, b, c, d) ) `, if there is only one block for the `allOf`? Would it be worth it to simplify it to simply use the `anyOf(a, b, c, d)`? Maybe I'm wrong 🤔

mdelapenya: I know that it's not related to the changes in the PR, but I've just seen same pattern in a few reviews today

v1v: `allOf ( anyOf (a, b, c, d) )` is nto the case but `allOf ( anyOf (a, b, c, d) e)`:

There are two conditions in the `allOf`
- params.bench_ci is always required to be true and
- anyOf ( branch == master , tag ~= `v[0-9].*` , params.Run_As_Master_Branch, env.BENCHMARK_UPDATED != "false")

In other words, the bench stage is launched when an input parameter and either a branch/tag or changes in certain files of the source code.

Jenkinsfile
```diff
@@ -59,6 +59,17 @@ pipeline {
             deleteDir()
             gitCheckout(basedir: "${BASE_DIR}", githubNotifyFirstTimeContributor: true)
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
+            // Look for changes related to the benchmark, if so then set the env variable.
+            script {
+              dir("${BASE_DIR}"){
+                def regexps =[
+                  '^packages/rum/test/benchmarks/',
```
v1v:
```suggestion
                  '^packages/**/test/benchmarks/',
```

vigneshshanmugam: Yeah sounds good to me. I dont see any huge concern on this.

Jenkinsfile
mdelapenya: Oh oh I see, I did not see the second element in the AND. Thanks for clarifying!

PR #655, opened 2020-02-26
chore(tests): fix/remove disabled tests

jahtalab:
Part of #620 
This PR fixes some of the previously disabled tests and removes the rest which are unnecessary at this point. Going forward we should add comments to tests that we decided to disable about the reason for it and how/when we can fix and enabled the test.

This PR also fixes the source-map for tests. Note that `inline-source-map` seem to be the only way to get sourcemaps work with karma/jasmine.

### DIFF
--- a/dev-utils/build.js
+++ b/dev-utils/build.js
@@ -157,7 +157,7 @@ function getCommonWebpackConfig(bundleType, packageType) {
   const mode = isEnvProduction ? 'production' : 'development'
 
   return {
-    devtool: isEnvProduction ? 'source-map' : false,
+    devtool: isEnvProduction ? false : 'inline-source-map',
     mode,
     stats: {
       colors: true,
--- a/dev-utils/build.js
+++ b/dev-utils/build.js
@@ -157,7 +157,7 @@ function getCommonWebpackConfig(bundleType, packageType) {
   const mode = isEnvProduction ? 'production' : 'development'
 
   return {
-    devtool: isEnvProduction ? false : 'inline-source-map',
+    devtool: isEnvProduction ? 'source-map' : 'inline-cheap-module-source-map',
     mode,
     stats: {
       colors: true,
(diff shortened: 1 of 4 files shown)


dev-utils/build.js
```diff
@@ -157,7 +157,7 @@ function getCommonWebpackConfig(bundleType, packageType) {
   const mode = isEnvProduction ? 'production' : 'development'
 
   return {
-    devtool: isEnvProduction ? 'source-map' : false,
+    devtool: isEnvProduction ? false : 'inline-source-map',
```
vigneshshanmugam: did you try 'inline-cheap-module-source-map' ? inline-source-map is the slowest of all. Also this change affects the production bundles. I hope thats not what you intended.

vigneshshanmugam: did you try 'inline-cheap-module-source-map' ? inline-source-map is the slowest of all. Also this change affects the production bundles. I hope thats not what you intended.

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -119,38 +119,22 @@ describe('ApmServer', function() {
...
 
-    var trs = generateTransaction(19)
+    let trs = generateTransaction(19)
     trs.forEach(apmServer.addTransaction.bind(apmServer))
     expect(apmServer.transactionQueue.items.length).toBe(19)
     expect(apmServer._postJson).not.toHaveBeenCalled()
-    trs = generateTransaction(1)
-    trs.forEach(apmServer.addTransaction.bind(apmServer))
 
+    apmServer.transactionQueue.flush()
```
vigneshshanmugam: if we do this, we are sending transactions to the apm server. I feel like we are leaking here. Can you make it async and await for the response from apm server?

dev-utils/build.js
jahtalab: We should enable `'source-map'` for production and test `inline-cheap-module-source-map`

PR #657, opened 2020-02-26
chore(workflow): fix script command to bundlesize

vigneshshanmugam:
+ I understood the build script param wrongly, It needs the command needs to be ran and not what should be run. 

https://github.com/preactjs/compressed-size-action/blob/master/action.js#L61


.github/workflows/bundle-size.yml
```diff
@@ -14,5 +14,5 @@ jobs:
     - uses: preactjs/compressed-size-action@v1
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
-        build-script: "npx lerna run build:module"
+        build-script: "bundlesize"
```
jahtalab: Does this mean that this check will fail if the size exceeds the limit?

vigneshshanmugam: I am going to increase the size for now and allow no failures to make sure we catch bugs like this.

vigneshshanmugam: We cannot use this as you said, fixed it.

vigneshshanmugam: Increased the size limit for now to avoid it from breaking.

vigneshshanmugam:
Yeah thats the idea, the github actions work flow `Compressed Size` does exactly that. We can disable bundle size on travis and jenkins since its not useful for PR's from forked repos anyways.

kuisathaverat:
please, if you want additional checks, you can request them and we will add those checks to the Jenkins  CI, avoid creating new CI silos that also send notification, comments, and we are not aware of them.

vigneshshanmugam:
@kuisathaverat This is only used for posting the sizes of all bundles in comments, Didn't want to make a new feature request to the robots team since its a small change. But incase this would be interesting for the team and can make this happen, we are happy to remove this in favor of our own solution.

kuisathaverat:
>@kuisathaverat This is only used for posting the sizes of all bundles in comments, Didn't want to make a new feature request to the robots team since its a small change. But incase this would be interesting for the team and can make this happen, we are happy to remove this in favor of our own solution.

We'd preferred to manage it on Jenkins, to avoid having to maintain things in two places, in that way when it fails for some reason we know where we have to fix it.

kuisathaverat:
>@kuisathaverat This is only used for posting the sizes of all bundles in comments, Didn't want to make a new feature request to the robots team since its a small change. But incase this would be interesting for the team and can make this happen, we are happy to remove this in favor of our own solution.

We'd preferred to manage it on Jenkins, to avoid having to maintain things in two places, in that way when it fails for some reason we know where we have to fix it.

kuisathaverat:
>@kuisathaverat This is only used for posting the sizes of all bundles in comments, Didn't want to make a new feature request to the robots team since its a small change. But incase this would be interesting for the team and can make this happen, we are happy to remove this in favor of our own solution.

We'd preferred to manage it on Jenkins, to avoid having to maintain things in two places, in that way when it fails for some reason we know where we have to fix it.

kuisathaverat:
>@kuisathaverat This is only used for posting the sizes of all bundles in comments, Didn't want to make a new feature request to the robots team since its a small change. But incase this would be interesting for the team and can make this happen, we are happy to remove this in favor of our own solution.

We'd preferred to manage it on Jenkins, to avoid having to maintain things in two places, in that way when it fails for some reason we know where we have to fix it.

vigneshshanmugam:
The code was available as a Github Workflow so I went straight ahead and used it if its going to be of help to us. Please do let us know what we can do in the future for experimentation of useful packages?

kuisathaverat:
>Please do let us know what we can do in the future for experimentation of useful packages?

You can use your fork for those experiments.

vigneshshanmugam:
jenkins run the tests please

vigneshshanmugam:
> You can use your fork for those experiments.

Sure, but wont be useful for us. I still feel like we don't need everything to be in Jenkins. These small snippets would be helpful for the dev workflow. It's just gonna slow us down if we have to ask for a feature and then work on our own solution. 

-> Experiment 
-> If it provides value, then figure out an alternative if we want to manage everything via Jenkins
-> Work on a solution in Jenkins

Please do let me know your thoughts.

kuisathaverat:
BTW we have right now a Github check that makes exactly the same if you want a comment with the result of the test we have also this implemented for other stuff, so it is also implemented.

vigneshshanmugam:
@kuisathaverat Thanks Ivan, can you link me to the docs or how to do it? Happy to replace with our own solution.

kuisathaverat:
I've opened an issue to implement it https://github.com/elastic/apm-agent-rum-js/issues/667

PR #658, opened 2020-02-26
fix(rum-core): export browser responsiveness interval correctly

vigneshshanmugam:
+ hit us because of sideeffect of allowing failures on bundlesize
+ webpack warnings was not considered as blockers.

Added the fix in bundle tests that would error out on warnings as well and also used the default webpack config to make sure we align with release bundles.

### DIFF
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -44,7 +44,8 @@ import {
   EVENT_TARGET,
   HTTP_REQUEST_TYPE,
   USER_INTERACTION,
-  PAGE_LOAD
+  PAGE_LOAD,
+  BROWSER_RESPONSIVENESS_INTERVAL
 } from '../common/constants'
 import {
   truncateModel,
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -56,7 +57,6 @@ import { __DEV__ } from '../env'
 /**
  * Parameters used for Managed Transactions
  */
-const BROWSER_RESPONSIVENESS_INTERVAL = 500
```
jahtalab: Why does this needs to be moved back to constants?

vigneshshanmugam: We use it on transaction-service as well.

vigneshshanmugam:
No i meant like we allow failures on bundlesize, so the warnings seen from webpack were not visible to us and our bundle tests was not checking for warnings which was shown by webpack. I fixed this one. 

The real source of the problem is webpack stats, It compiles even if the exports/imports are not available and considers them as warnings instead of errors.

PR #664, opened 2020-02-28
BREAKING CHANGE(rum-core): process all events in a single queue for reporting

### LINKED ISSUE #628: Remove unnecessary config options

vigneshshanmugam:
+ part of #628 
+ Uses a single queue for reporting all events (transaction, errors etc)
+ Introduces `eventsPerMin` which is set to 80 by default including errors and transactions for throttling instead of `throttleLimit` and `throttleInterval` for errors and transactions. 
+ Fixed all E2E tests which depends on the method `sendErrors` and `sendTransactions`. Now the API only has `sendEvents`. 

**NOTE to reviewers: Review each commit separately**


vigneshshanmugam:
@jahtalab I wanted it to be easy for the reviewer since reviewing by commit should help ease up the process. I will rename the PR to breaking change once all issues are addressed. 
Regarding refactoring, there isnt much to do. Just removed an option from throttle so can keep it in same PR.

dev-utils/webdriver.js
```diff
@@ -233,94 +233,71 @@ function getWebdriveBaseConfig(
...
-            )
-          }
-          if (serverCalls.sendTransactions) {
-            promises = promises.concat(
-              serverCalls.sendTransactions.map(function(s) {
-                return s.returnValue
-              })
-            )
-          }
+function waitForApmServerCalls(eventsCount = 0) {
```
jahtalab: I think in the tests we want to differentiate between different even types (error, or transactions), for example in some tests we only care about transactions and we want to ignore the any number of errors.

vigneshshanmugam:
@jahtalab The API has changed from sendErrors, sendTransactions to `sendEvents` so we have to change those files anyways and its not a refactor. In order to differentiate between errors and transactions, I have just modified the code and used in the E2E. There is nothing much to be refactored as separate PR IMO. Please review commit by commit.

dev-utils/webdriver.js
vigneshshanmugam: Its differentiated in the Tests in this way 

```js
const serverCalls = waitForServerCalls(2)
const { transactions,  errors } = serverCalls.sendEvents
```

jahtalab:
We decided to bring back the individual event type count to the `waitForApmServer` function.

packages/rum-core/src/common/apm-server.js
```diff
@@ -28,21 +28,21 @@ import throttle from './throttle'
 import NDJSON from './ndjson'
 import { XHR_IGNORE } from './patching/patch-utils'
 import { truncateModel, METADATA_MODEL } from './truncate'
-import { SERVER_URL_PREFIX } from './constants'
+import { SERVER_URL_PREFIX, ERRORS, TRANSACTIONS } from './constants'
 import { Promise } from './polyfills'
 import { __DEV__ } from '../env'
 
+/**
+ * Throttling interval deafultss to 60 seconds
```
jahtalab: nit: `defaults`

packages/rum-core/src/common/apm-server.js
```diff
@@ -193,70 +190,38 @@ class ApmServer {
...
   }
 
   addTransaction(transaction) {
-    if (!this.transactionQueue) {
-      this.initTransactionQueue()
-    }
-    this.throttleAddTransaction(transaction)
+    this.throttleEvents({
+      data: transaction,
+      type: TRANSACTIONS
```
jahtalab: As discussed we can just use the type as the key: `{[type]:data}`

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -123,62 +128,23 @@ describe('ApmServer', function() {
...
+    expect(apmServer.queue.items.length).toBe(19)
     expect(apmServer._postJson).not.toHaveBeenCalled()
 
-    apmServer.transactionQueue.flush()
+    apmServer.queue.flush()
     expect(apmServer._postJson).toHaveBeenCalled()
-    expect(apmServer.transactionQueue.items.length).toBe(0)
-  })
-
-  it('should init queue if not initialized before', function(done) {
```
jahtalab: Don't we still need this test for the event queue?

packages/rum/test/specs/index.spec.js
```diff
@@ -90,12 +89,12 @@ describe('index', function() {
       throw new Error('ApmBase test error')
     } catch (error) {
       apmBase.captureError(error)
-      expect(apmServer.sendErrors).not.toHaveBeenCalled()
```
jahtalab: I think this still applies just needs to be change to `sendEvents`

packages/rum-core/test/common/apm-server.spec.js
vigneshshanmugam: This was there because we were using it for testing purposes. right now its under the assumption that `apmServer.init` is called always when we are depending on the queue in tests.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: I will also see if we can eliminate type and use the key for figuring out the event type.

packages/rum/test/specs/index.spec.js
vigneshshanmugam: yeah will add it back.

packages/rum-core/src/common/config-service.js
```diff
@@ -77,28 +77,19 @@ class Config {
       logLevel: 'warn',
       breakdownMetrics: false,
       ignoreTransactions: [],
-      errorThrottleLimit: 20,
-      errorThrottleInterval: 30000,
-      transactionThrottleLimit: 20,
-      transactionThrottleInterval: 30000,
-
+      eventsPerMin: 80,
```
jahtalab: Another option for this might be just `eventsLimit`, not having the interval (minute) in the name helps if we want to change the interval in the future without having to changing the name. Another benefit is that we don't have to use shortened `min`  in the name.

packages/rum-core/src/common/apm-server.js
```diff
@@ -51,29 +51,32 @@ class ApmServer {
...
-        language: {
-          name: 'javascript'
-        },
-        environment: cfg.get('environment')
-      },
-      labels: cfg.get('context.tags')
+  _initQueue() {
+    const queueLimit = this._configService.get('queueLimit')
+    const flushInterval = this._configService.get('flushInterval')
+    const limit = this._configService.get('eventsPerMin')
```
jahtalab: Do you we need to check if any of these config options are undefined?

packages/rum-core/src/common/apm-server.js
```diff
@@ -193,70 +190,32 @@ class ApmServer {
...
-    this.throttleAddError(error)
+    this.throttleEvents({ [ERRORS]: error })
   }
 
   addTransaction(transaction) {
-    if (!this.transactionQueue) {
-      this.initTransactionQueue()
-    }
-    this.throttleAddTransaction(transaction)
+    this.throttleEvents({ [TRANSACTIONS]: transaction })
```
jahtalab: I think it makes more sense to use `transaction` (singular) as the key here. But then we still should keep the payload the same with `transactions` key since that is passed to the filters and also there it is an array. The same with errors. WDYT?

packages/rum-core/src/common/apm-server.js
```diff
@@ -284,44 +243,46 @@ class ApmServer {
...
-    }
-
+    let ndjson = []
     const metadata = this.createMetaData()
-    ndjson.unshift(NDJSON.stringify({ metadata }))
+    ndjson.push(NDJSON.stringify({ metadata }))
+
+    ndjson = ndjson.concat(
+      this.ndjsonErrors(filteredPayload[ERRORS]),
+      this.ndjsonTransactions(filteredPayload[TRANSACTIONS])
```
jahtalab: Does the order of these events matter from the APM server perspective?

One area we should check is that since APM server sets all the timestamps for a transaction and it's spans in the same request, would that still work?

packages/rum-core/test/common/apm-server.spec.js
jahtalab: I think we should still test how the APM server behaves if we (or the user doesn't call the init) function. IMO, apm server class should handle these cases as well.

vigneshshanmugam: Not sure why this is important, Please explain the scenario when this would happen. APM server class would still be unusable without init.

packages/rum-core/src/common/config-service.js
vigneshshanmugam: Didn't we think about this first and decide then its not super informative for the user and decided to use `eventsPerMin`? . I like `eventsLimit` as well.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: Hmmm, We can be defensive and handle all cases. I am not 100% sold on the idea of testing user config when we clearly mention each config in the docs and explain the type as well.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: When writing i had similar thoughts as well, Then it introduces too many variations. So sticked with plurals to be similar to the filters. Also Logically we are adding handling `transactions` as type so i am fine with keeping as it is.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: Order does not matter as far as i checked with the server team. Can you confirm @simitt ? 

> One area we should check is that since APM server sets all the timestamps for a transaction and it's spans in the same request, would that still work?

I didn't get this question. Isn't the same even before as well when we had separate queues?

packages/rum-core/test/common/apm-server.spec.js
jahtalab: My only concern is that if the user doesn't call the init, the apm server is not going to throw any exception, and instead gracefully handles.

We don't have to add these exact tests back btw, we just need to make sure apm server doesn't throw errors if it's not initialised.

packages/rum-core/src/common/config-service.js
jahtalab: yeah, last we discussed we said to keep the interval in the name, but thinking about it again, it seems to me it's better to document the interval instead just keep the name simple.

packages/rum-core/src/common/apm-server.js
jahtalab: I think some validations would be useful at least when the user sets these values would be useful, but should do that separately.

packages/rum-core/src/common/apm-server.js
jahtalab: Previously we sent the transaction with all of its spans in the same request, which means the APM Server might check for the first event and if that's a transaction then set the timestamp of all the following spans to the same value, but now the first event might be an error. This might not be how the APM server handles requests but I want to make sure the corner cases we have for the RUM payload will still work when we make this change.

simitt: Current APM Server behavior is that the server records a `requestTime := time.Now` when the request comes in. If a span does not contain a dedicated timestamp, the spans timestamp is set to `requestTime + event.Start`. 

I am not certain if and how this behavior will change with the RUM specific optimizations @jalvz and @vigneshshanmugam are working on.

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: Yeah, we need to have validations for all configs and report warnings for non important ones and keep it as error for others like we have for `serverUrl`

packages/rum-core/src/common/apm-server.js
vigneshshanmugam: Thanks for chiming in, We are not planning to change that behaviour with the RUM payload optimisations AFAIK. @jalvz Can you confirm that the optimisations that we do for the payload wont affect the ordering of events in any way?

packages/rum-core/src/common/config-service.js
vigneshshanmugam: I can change it to `eventsLimit`. I like both so lets go for it.

packages/rum-core/test/common/apm-server.spec.js
vigneshshanmugam: > My only concern is that if the user doesn't call the init, the apm server is not going to throw any exception, and instead gracefully handles.

I am not sure if users would need to use our APMServer class alone, but if we need a graceful handling, We can do two things

1. We can do a `noop` for throttleEvents 

(OR)

2. Initialize queue during ApmServer class construction itself and get rid of init method and makes testing also a lot easier. 

WYDT?

vigneshshanmugam: I went with option 1 as it requires less changes and added tests covering the case you mentioned.

bmorelli25:
I'll add documentation for `eventsLimit` in https://github.com/elastic/apm-agent-rum-js/pull/694

packages/rum-core/src/common/apm-server.js
jalvz: Yes, that's correct. However:

> Previously we sent the transaction with all of its spans in the same request

This is no longer the case?

vigneshshanmugam: It's still the same, spans will follow transactions as they were before. This change only introduces error events inside the same payload, so it would look like this
```
<metadata>
<error 1>
<error 2>
<transaction1>
<spans of transaction 1>
<metricsets 1 if available>
<transaction2>
<spans of transaction 2>
<metricsets 2 if available>
```

packages/rum-core/test/error-logging/error-logging.spec.js
```diff
@@ -71,51 +73,47 @@ describe('ErrorLogging', function() {
...
+      expect(errorData.transaction_id).toEqual(transaction.id)
+      expect(errorData.trace_id).toEqual(transaction.traceId)
+      expect(errorData.parent_id).toEqual(transaction.id)
+      expect(errorData.transaction).toEqual({
+        type: transaction.type,
+        sampled: transaction.sampled
+      })
+      transaction.end()
+
+      apmServer
```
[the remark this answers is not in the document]
vigneshshanmugam: yeah good point, Ini resets it. but i will add it anyways.

vigneshshanmugam:
Investigating the failing saucelabs tests. Will merge once all issues are resolved with E2E.

PR #666, opened 2020-03-02
chore: fix test script for running locally

jahtalab:
When running locally `concurrency=1` is required since each of individual packages run their own servers and they conflict with each other if run in parallel. 

This doesn't effect the speed of tests running on CI since we always pass a single scope to each runner.

### DIFF
--- a/dev-utils/run-script.js
+++ b/dev-utils/run-script.js
@@ -42,6 +42,8 @@ const { generateNotice } = require('./dep-info')
 const PROJECT_DIR = join(__dirname, '../')
 const { sauceLabs } = getTestEnvironmentVariables()
 
+let cleanUps = []
+
 function runUnitTests(packagePath, startSauceConnect = 'false') {
   const karmaConfigFile = join(PROJECT_DIR, packagePath, 'karma.conf.js')
   if (startSauceConnect === 'true') {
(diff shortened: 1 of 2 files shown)


dev-utils/run-script.js
```diff
@@ -158,21 +160,28 @@ function runSauceTests(packagePath, serve = 'true', ...scripts) {
...
+    })
+    cleanUps = []
+  }
+  if (callExit) {
+    if (exitCode !== 0) {
+      console.log('Exiting with code:', exitCode)
     }
-  })
-  cleanUps = []
+    process.exit(exitCode)
```
vigneshshanmugam: we might miss the console.log above since its async and this exit would call synchronously. Its better to set the exitcode instead of doing thiss
```js
process.exitCode = exitCode
```

dev-utils/run-script.js
```diff
@@ -184,7 +193,13 @@ const scripts = {
   runNodeTests,
   runBundleTests,
   buildE2eBundles,
-  startTestServers
+  startTestServers() {
+    let servers = startTestServers.apply(this, arguments)
```
vigneshshanmugam: We do the same thing on runSauceTests as well, can you move this to the top and use it there as well.

vigneshshanmugam:
@kuisathaverat Any idea why the test result not getting reported still for this PR?

kuisathaverat:
>Any idea why the test result not getting reported still for this PR?

With all the changes I lose the JUnit step somewhere.

https://github.com/elastic/apm-agent-rum-js/issues/672

jahtalab:
@vigneshshanmugam , It seems that these changes make the travis pass even though they should fail, I'll look into it.

vigneshshanmugam:
@jahtalab Restarted the travis tests now

PR #668, opened 2020-03-02
docs: addLabels works for spans as well

jahtalab:


docs/agent-api.asciidoc
```diff
@@ -86,7 +86,8 @@ deprecated[4.1.x,Tags have been replaced with labels], Please use <<apm-add-labe
 apm.addLabels({ [name]: value })
 ----
 
-Set labels on transactions and errors.
+Set labels on transactions and errors. 
+Starting with APM Server 7.6+, the labels are added to spans as well.
```
vigneshshanmugam: this happen on the server right? For global labels?

jahtalab: That's correct!

vigneshshanmugam: Do you want to link to the server docs? just an idea

jahtalab: I don't think that's necessary since the user shouldn't need to care about how it works behind the scene.

jahtalab:
Merging this as failures are unrelated to the changes

PR #676, opened 2020-03-04
chore(benchmark): run only on chromium and exit process

vigneshshanmugam:
+ extracting some commits from this #https://github.com/elastic/apm-agent-rum-js/pull/650
+ Playwright requires Node >8 otherwise we will see this error - https://github.com/microsoft/playwright/blob/master/docs/troubleshooting.md#referenceerror-url-is-not-defined
+ Firefox and Webkit require more packages to work correctly inside docker. Will be addressed in separate PR

### DIFF
--- a/dev-utils/docker-compose.yml
+++ b/dev-utils/docker-compose.yml
@@ -125,7 +125,7 @@ services:
   node-puppeteer:
     build: ../.ci/docker/node-puppeteer
     container_name: node-puppeteer
-    image: docker.elastic.co/observability-ci/node-puppeteer:8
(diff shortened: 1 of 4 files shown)


dev-utils/docker-compose.yml
```diff
@@ -156,8 +156,8 @@ services:
 
   node-benchmark:
     build: ../.ci/docker/node-puppeteer
-    container_name: node-puppeteer
-    image: docker.elastic.co/observability-ci/node-puppeteer:8
+    container_name: node-playwright
+    image: docker.elastic.co/observability-ci/node-playwright:12
```
kuisathaverat: this image will not exists

Issue #680, opened 2020-03-04
fcpEntry[0] can be undefined if 'first-contentful-paint' is missing from performance.getEntriesByType('paint')


bjorngi:
@elastic/apm-rum-core version: 4.10.0
OS: Ubuntu 19.10
Browser: Google Chrome Version 80.0.3987.87 (Official Build) (64-bit)

After update to 4.10.0 we get an console.error when entering our web application (see below)
```
capture-navigation.js:203 Uncaught (in promise) TypeError: Cannot destructure property 'startTime' of 'fcpEntry[0]' as it is undefined.
    at getFirstContentfulPaint (capture-navigation.js:203)
    at getPageLoadMarks (capture-navigation.js:217)
    at captureNavigation (capture-navigation.js:277)
    at transaction-service.js:248
```

If `performance.getEntriesByType('paint')` only contain 'first-paint' and NOT 'first-contentful-paint' `fcpEntry` will give and empty array and therefor `fcpEntry[0]` will be undefined. This will cause the error when trying to look up `startTime` of undefined.

https://github.com/elastic/apm-agent-rum-js/blob/5e0c9a2acbd8460de5dbbf7827311e0d0f845047/packages/rum-core/src/performance-monitoring/capture-navigation.js#L267

jahtalab:
Fixed by #681

PR #681, opened 2020-03-04
Fix(rum-core): Check if first-contentful-paint exists before setting the agent marks

bjorngi:
https://github.com/elastic/apm-agent-rum-js/issues/680

### DIFF
--- a/packages/rum-core/src/performance-monitoring/capture-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-navigation.js
@@ -268,7 +268,7 @@ function getFirstContentfulPaint() {
     }, null)
     return fcp
   }
-  return null
+  return undefined
 }
 
 function getPageLoadMarks() {
--- a/packages/rum-core/src/performance-monitoring/capture-navigation.js
+++ b/packages/rum-core/src/performance-monitoring/capture-navigation.js
@@ -254,24 +254,21 @@ function getNavigationTimingMarks() {
  * SPEC - https://www.w3.org/TR/paint-timing/
  */
 function getFirstContentfulPaint() {
-  let fcp
   if (isPerfTimelineSupported()) {
-    const entries = PERF.getEntriesByType(PAINT)
-    if (entries.length > 0) {
-      const timing = PERF.timing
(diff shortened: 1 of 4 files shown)


packages/rum-core/src/performance-monitoring/capture-navigation.js
```diff
@@ -254,24 +254,21 @@ function getNavigationTimingMarks() {
...
-      const { startTime } = fcpEntry[0]
-      fcp = unloadDiff >= 0 ? startTime - unloadDiff : startTime
-    }
+    const timing = PERF.timing
+    /**
+     * To avoid capturing the unload event handler effect
+     * as part of the transaction duration
+     */
+    const unloadDiff = timing.fetchStart - timing.navigationStart
+    const fcp = PERF.getEntriesByType(PAINT).reduce((_fcp, entry) => {
```
vigneshshanmugam: can you add a initial value of `null` to the reduce function as entries can be empty array `[]` and we need an initial value for reduce to work. 

```js
reduce(fcp,entry=> {}, null)
```

bjorngi: CLA Signed and initialValue added.

vigneshshanmugam:
jenkins run the test please

jahtalab:
@bjorngi , Regarding the CLA issue, it seems that you signed the CLA with a different email compared to the [email set in the commit](https://github.com/elastic/apm-agent-rum-js/commit/e00b0e3be7e3f4e64dd6850461fd3de06319eae4.patch), would you please sign it with the correct email?

bjorngi:
>     1. would you please change the return value to `undefined`  instead of `null`?
Done

>     2. We generally add or improve tests when ever a new bug is discovered, would you please add some tests to cover these fixes? Please have a look at our [Contributing guide](https://github.com/elastic/apm-agent-rum-js/blob/master/CONTRIBUTING.md) and let us know if you need any help.
Not sure how good the test are. Added new mock for missing 'first-contentful-paint' and another test for base case.

CLA signed with correct email for key.

PR #690, opened 2020-03-09
fix(rum-core): replace >> in click transaction name with '-'

### LINKED ISSUE #689: Avoid using >> in Click transaction name
More detail [here](https://discuss.elastic.co/t/apm-rum-click-events-transaction-names/222496)

jahtalab:
Fixes #689

### DIFF
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -210,7 +210,7 @@ export default class PerformanceMonitoring {
          * related to this interaction.
          */
         const tr = transactionService.startTransaction(
-          `Click >> ${tagName}${additionalInfo}`,
+          `Click - ${tagName}${additionalInfo}`,
           USER_INTERACTION,
           {
             managed: true,
--- a/packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
+++ b/packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
@@ -695,7 +695,7 @@ describe('PerformanceMonitoring', function() {
 
       let tr = transactionService.getCurrentTransaction()
       expect(tr).toBeDefined()
-      expect(tr.name).toBe('Click >> button')
+      expect(tr.name).toBe('Click - button')
       expect(tr.context.custom).toEqual({
         classes: 'cool-button purchase-style'
       })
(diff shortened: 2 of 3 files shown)


jahtalab:
@vigneshshanmugam , Please see this discus thread: https://discuss.elastic.co/t/apm-rum-click-events-transaction-names/222496

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -210,7 +210,7 @@ export default class PerformanceMonitoring {
          * related to this interaction.
          */
         const tr = transactionService.startTransaction(
-          `Click >> ${tagName}${additionalInfo}`,
+          `Click - ${tagName}${additionalInfo}`,
```
vigneshshanmugam: Better character would be `:` instead of `-` as tagName can also contain `-` if it's a custom element.

vigneshshanmugam: Also no space between the click and colon. Name would be `Click: <otherinfo>`

jahtalab: I would prefer to limit the number of different characters we use in the name to avoid this type of issues. Unless there's a technical reason not to use `-`, specially since there are spaces to indicate the separation.

vigneshshanmugam: Yes, there is indeed a technical reason and it is `Custom Elements`. You can refer the spec here https://html.spec.whatwg.org/multipage/custom-elements.html

Most of the custom elements would be written in the form `<customized-button>` and our click transaction would be `Click - CUSTOMIZED-BUTTON`. This feels bad from the naming perspective.

jahtalab: I meant a technical reason in the sense that the character would not be shown correctly in the UI, which I don't think is the case here.  IMO, `Click - CUSTOMIZED-BUTTON`  is clear since there're space to indicate the separation, we similarly use spaces in xhr transaction name to separate the method from the url.

jahtalab: Alternatively we can come up with a standard naming convention to use for all transaction and span names, but that's out of the scope of this PR.

jahtalab: Created an issue: https://github.com/elastic/apm-agent-rum-js/issues/692

PR #696, opened 2020-03-11
BREAKING CHANGE: remove addTags method

jahtalab:
Part of #215


jahtalab:
AFAICS, The `tags` key is added to the context and eventually sent to APM Server, so we should keep it as tags until APM server changes that.

jahtalab:
jenkins run tests please

Issue #272, opened 2019-05-27
Add support for boolean and number values for labels


jahtalab:
Currently, boolean and number values for labels are converted to their string representation before being added to the payload. However, with APM Server 6.7, the support for these value types has been [added to APM Server](https://github.com/elastic/apm-server/pull/1712). We should start sending the raw values for these types.

Please note that this is a breaking change for older versions of APM server (prior to 6.7) and we should update the documentation to reflect this.

-  [ ] We should add new tests for these

felixbarny:
One way to get around making this a breaking change is to check the APM Server version and convert numbers and booleans to strings when connected to an APM Server < 6.7.
I guess it depends on how urgent the need to offer these APIs are for the RUM agent and if it's worth the complexity. Especially for the RUM agent that would, however, mean another request to the APM Server just to check the version. That makes it less feasible for the RUM agent. Still wanted to mention it because it may be relevant for other agents.

jahtalab:
Thanks @felixbarny , I think the cost of making an extra request just to check the APM server version, is too high. I don't see the breaking change as a big issue as long as we release it in a major version.

Plus, we can make documentation about how users can convert those values to string before calling the agent APIs, if the support for older versions is necessary.

PR #697, opened 2020-03-11
BREAKING CHANGE: add support for boolean and number values in addLabels

jahtalab:
Part of #272

### DIFF
--- a/packages/rum-core/src/common/utils.js
+++ b/packages/rum-core/src/common/utils.js
@@ -143,8 +143,13 @@ function isPlatformSupported() {
  */
 function setLabel(key, value, obj) {
   if (!obj || !key) return
-  var skey = removeInvalidChars(key)
-  if (value) {
+  const skey = removeInvalidChars(key)
+  let valueType = typeof value
+  if (
+    valueType !== 'undefined' &&
+    valueType !== 'boolean' &&
+    valueType !== 'number'
+  ) {
     value = String(value)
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/common/utils.js
```diff
@@ -143,8 +143,13 @@ function isPlatformSupported() {
  */
 function setLabel(key, value, obj) {
   if (!obj || !key) return
-  var skey = removeInvalidChars(key)
-  if (value) {
+  const skey = removeInvalidChars(key)
+  let valueType = typeof value
+  if (
```
vigneshshanmugam: This breaks the previous assumption when the value is `null`, We need to add a test when value is null.

packages/rum-core/src/common/utils.js
```diff
@@ -134,18 +134,16 @@ function isPlatformSupported() {
...
+ * Support for boolean and number in the APM server landed in 6.7
+ * therefore, we keep these values unchange but convert any other
+ * type to string.
  */
 function setLabel(key, value, obj) {
   if (!obj || !key) return
   const skey = removeInvalidChars(key)
   let valueType = typeof value
   if (
+    value !== null &&
```
vigneshshanmugam: we can merge these two lines to `value !== null && valueType !== 'undefined'` to `value != undefined` which handles both null and undefined.

PR #699, opened 2020-03-11
feat(rum-core): set sync field only for synchronous spans

### LINKED ISSUE #619: set `sync` flag only for synchronous spans
+ Since the agent currently sets `sync`: false for all fetch calls which is already async and for xhr based on the third parameter which would be most of the time `false`. We can only send sync flag when the request itself is synchronous. This would help avoid the UI clutter where `async` badge would be displayed all the time. 

https://github.com/elastic/kibana/issues/25928#issuecomment-455506469

vigneshshanmugam:
+ fix #619

### DIFF
--- a/packages/rum-core/src/common/patching/fetch-patch.js
+++ b/packages/rum-core/src/common/patching/fetch-patch.js
@@ -65,7 +65,6 @@ export function patchFetch(callback) {
       data: {
         target: request,
         method: request.method,
-        sync: false,
         url,
         aborted: false
       }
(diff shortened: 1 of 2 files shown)


jahtalab:
We discussed this and decided to test with older version of kibana (7.0) as well.

vigneshshanmugam:
@jahtalab Tested  with `7.0.0` stack and everything is working as expected. 
Screenshot 2020-03-17 at 14 50 01

PR #700, opened 2020-03-12
BREAKING CHANGE: tags are not added to individual events

### LINKED ISSUE #618: Remove global labels Errors and Transactions in the payload
With #617 the global config is added to the metadata document and APM server sets these labels on errors and transactions. Therefore we don't need to set them on each transaction or error individually, we should remove them to say payload size. However, we should consider this a breaking change since older versions of APM server would not recognise `metadata.labels`.

jahtalab:
Fixes #618 

instead we rely on metadata labels

### DIFF
--- a/packages/rum-core/test/common/context.spec.js
+++ b/packages/rum-core/test/common/context.spec.js
@@ -167,10 +167,10 @@ describe('Context', () => {
 
   it('should enrich transaction with context info based on type', () => {
     const transaction = new Transaction('test', 'custom')
-    let trContext = { tags: { tag1: 'tag1' } }
+    const trContext = { tags: { tag1: 'tag1' } }
     transaction.addContext(trContext)
     transaction.end()
-    let userContext = {
+    const userContext = {
       user: {
         email: 'test@example.com',
         id: '123'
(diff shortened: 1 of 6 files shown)


packages/rum-core/test/common/context.spec.js
```diff
@@ -167,12 +167,17 @@ describe('Context', () => {
 
   it('should enrich transaction with context info based on type', () => {
     const transaction = new Transaction('test', 'custom')
+    let trContext = { tags: { tag1: 'tag1' } }
```
vigneshshanmugam: nit: const

packages/rum-core/test/common/context.spec.js
```diff
@@ -167,12 +167,17 @@ describe('Context', () => {
...
+    transaction.addContext(trContext)
     transaction.end()
-    const configContext = {
+    let userContext = {
```
vigneshshanmugam: same

jahtalab:
@vigneshshanmugam , That's right, we probably should improve the truncation tests since they are only testing the mock models and the truncation function but not the real models.

Issue #249, opened 2019-04-11
Return a noop Transaction and Span instead of undefined if the agent is not active


jahtalab:
Currently if the agent is not active methods such as `startTransaction` and `startSpan` would return undefined. However this makes coding difficult since there's constant need to check for undefined. 

We can improve this by returning noop Transactions and Spans. But we need to mark these entities clearly as noop.

Frozen-byte:
When inactive, 
version 4.6.0 `startTransaction()` returned a `ApmBase` instance  
in 4.7.0 it returns `undefined`

Not sure what is the intended behavior here.

vigneshshanmugam:
When the agent is inactive

+ 4.6.0 - startTransaction should return Transaction class and not `ApmBase`? Could you please double check once? 

+ 4.7.0 - startTransaction should return `undefined` and it is the correct behaviour for now. 

Bug fix - #569

Frozen-byte:
You are totally right, was an `Transaction` instance.

vigneshshanmugam:
Oops, sorry for that. To be honest this was a bug that was always there and we did a quick fix as this could result in custom transactions being sent to the server when the agent was inactive which is not the expected behaviour. This issue has more details to it - https://github.com/elastic/apm-agent-rum-js/issues/566

I am curious on why would the user be setting the agent to `active: false` and want to do the custom transactions. Incase if this was the expected behaviour, then you would need to set the `instrument` flag to false which would completely disable all automatic instrumentations. Let me know what you think?

Frozen-byte:
In our React-App; we have not updated react-router (...economic reasons) and are forced to handle the Routing events with custom Events.
For all other "default" apiRequests the metrics are perfectly captured without any further hassle.

In development environment we disable RUM by default to keep the Database cleaner :upside_down_face:

PR #705, opened 2020-03-16
perf(rum-core): random number generator using crypto.getRandomValues

vigneshshanmugam:
+ part of #621 
+ use crypto.getRandomValues API and remove uuid dependency as it adds extra bytes and also provides polyfill which is not necessary since we removed support for IE 10. 
+ small wins on bundlesize. 

## Benchmarks
```
generateRandomIdOld: 219.953125ms
generateRandomIdNew: 213.82421875ms
```

**similar results**

#### Code
```js
var noOfTimes = 1e5

var byteToHex = []
for (var i = 0; i < 256; ++i) {
  byteToHex[i] = (i + 0x100).toString(16).substr(1)
}

function bytesToHexOld(buf, offset) {
  var i = offset || 0
  var bth = byteToHex
  // join used to fix memory issue caused by concatenation: https://bugs.chromium.org/p/v8/issues/detail?id=3175#c4
  return [
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]],
    bth[buf[i++]]
  ].join('')
}

function bytesToHexNew(buffer) {
  const hexOctets = []
  for (let i = 0; i < buffer.length; i++) {
    hexOctets.push(byteToHex[buffer[i]])
  }
  return hexOctets.join('')
}

var destination = new Uint8Array(16)

function rng() {
  if (typeof crypto != 'undefined' && crypto.getRandomValues) {
    return crypto.getRandomValues(destination)
  } else if (
    typeof msCrypto != 'undefined' &&
    typeof msCrypto.getRandomValues == 'function'
  ) {
    return msCrypto.getRandomValues(destination)
  }
  return destination
}

function generateRandomIdOld(length) {
  var id = bytesToHexOld(rng())
  return id.substr(0, length)
}

function generateRandomIdNew(length) {
  var id = bytesToHexNew(rng())
  return id.substr(0, length)
}

console.time('generateRandomIdOld')
for (let i = 0; i < noOfTimes; i++) {
  generateRandomIdOld(32)
}
console.timeEnd('generateRandomIdOld')

console.time('generateRandomIdNew')
for (let i = 0; i < noOfTimes; i++) {
  generateRandomIdNew(32)
}
console.timeEnd('generateRandomIdNew')

```

### DIFF
--- a/dev-utils/test-servers.js
+++ b/dev-utils/test-servers.js
@@ -27,6 +27,7 @@ const express = require('express')
 const serveIndex = require('serve-index')
 const proxy = require('express-http-proxy')
 const { join } = require('path')
+const useragent = require('useragent')
 const { runIntegrationTest } = require('./integration-test')
 const { DEFAULT_APM_SERVER_URL } = require('./test-config')
 
@@ -46,17 +47,25 @@ function startBackendAgentServer(port = 8003) {
   })
 
   function dTRespond(req, res) {
-    const header = req.headers['traceparent']
-    let payload = { noHeader: true }
-    if (header) {
-      const splited = header.split('-')
+    const traceHeader = req.headers['traceparent']
+    const browser = useragent.is(req.headers['user-agent'])
+    /**
+     * Unsupported browsers IE 10
+     */
+    let payload = { supported: true }
+    if (browser.ie && Number(browser.version) < 11) {
+      payload.supported = false
+    }
+    if (traceHeader) {
+      const splited = traceHeader.split('-')
       payload = {
         version: splited[0],
         traceId: splited[1],
         parentId: splited[2],
-        flags: splited[3]
+        flags: splited[3],
+        ...payload
       }
-      console.log('traceparent:', header)
+      console.log('traceparent:', traceHeader)
     }
 
     res.setHeader('Content-Type', 'application/json')
--- a/dev-utils/test-servers.js
+++ b/dev-utils/test-servers.js
@@ -27,7 +27,6 @@ const express = require('express')
 const serveIndex = require('serve-index')
 const proxy = require('express-http-proxy')
 const { join } = require('path')
-const useragent = require('useragent')
 const { runIntegrationTest } = require('./integration-test')
 const { DEFAULT_APM_SERVER_URL } = require('./test-config')
 
@@ -47,25 +46,17 @@ function startBackendAgentServer(port = 8003) {
   })
 
   function dTRespond(req, res) {
-    const traceHeader = req.headers['traceparent']
-    const browser = useragent.is(req.headers['user-agent'])
-    /**
-     * Unsupported browsers IE 10
-     */
-    let payload = { supported: true }
-    if (browser.ie && Number(browser.version) < 11) {
-      payload.supported = false
-    }
-    if (traceHeader) {
-      const splited = traceHeader.split('-')
+    const header = req.headers['traceparent']
+    let payload = { noHeader: true }
+    if (header) {
+      const splited = header.split('-')
       payload = {
         version: splited[0],
         traceId: splited[1],
         parentId: splited[2],
-        flags: splited[3],
-        ...payload
+        flags: splited[3]
       }
-      console.log('traceparent:', traceHeader)
+      console.log('traceparent:', header)
     }
 
     res.setHeader('Content-Type', 'application/json')
--- a/package-lock.json
+++ b/package-lock.json
@@ -21388,6 +21388,12 @@
             "glob": "^7.1.3"
           }
         },
+        "semver": {
+          "version": "5.7.1",
+          "resolved": "https://registry.npmjs.org/semver/-/semver-5.7.1.tgz",
+          "integrity": "sha512-sauaDf/PZdVgrLTNYHRtpXa1iRiKcaebiKQ1BJdpQlWH2lCvexQdX55snPFyK7QzpudqbCI0qXFfOasHdyNDGQ==",
+          "dev": true
+        },
         "source-map": {
           "version": "0.6.1",
           "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.6.1.tgz",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2736,13 +2736,6 @@
         "error-stack-parser": "^1.3.5",
         "opentracing": "^0.14.3",
         "promise-polyfill": "^8.1.3"
-      },
-      "dependencies": {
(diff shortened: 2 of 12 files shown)


packages/rum/test/e2e/utils.js
```diff
@@ -24,9 +24,9 @@
  */
 
 function checkDtInfo(payload) {
-  console.log('distributed tracing header value', payload)
-  if (typeof payload.traceId !== 'string') {
-    throw new Error('Wrong distributed tracing payload: ')
+  console.log('distributed tracing header value', JSON.stringify(payload))
+  if (payload.supported && typeof payload.traceId !== 'string') {
```
vigneshshanmugam: this is to handle errors thrown on failsafe E2E tests (IE 10)

vigneshshanmugam:
@jahtalab good for review now, lets get this in for 5.x as well.

dev-utils/test-servers.js
```diff
@@ -27,6 +27,7 @@ const express = require('express')
 const serveIndex = require('serve-index')
 const proxy = require('express-http-proxy')
 const { join } = require('path')
+const useragent = require('useragent')
```
jahtalab: I'm not sure if we need to solve this issue on the server side, seems like this is something we can check on the browser.

packages/rum-core/src/common/utils.js
```diff
@@ -34,33 +33,45 @@ function isCORSSupported() {
...
+ */
 var byteToHex = []
 for (var i = 0; i < 256; ++i) {
   byteToHex[i] = (i + 0x100).toString(16).substr(1)
 }
 
-function bytesToHex(buf, offset) {
-  var i = offset || 0
-  var bth = byteToHex
-  // join used to fix memory issue caused by concatenation: https://bugs.chromium.org/p/v8/issues/detail?id=3175#c4
```
vigneshshanmugam: The memory issue mentioned here is not relevant for us as we are dealing with 16 items in the array and not thousands as explained in the issue. Also, we are still using array.join for concatenation just to keep us safe from the above issue.

packages/rum-core/src/common/utils.js
```diff
@@ -34,33 +33,45 @@ function isCORSSupported() {
...
+function rng() {
+  if (typeof crypto != 'undefined' && crypto.getRandomValues) {
+    return crypto.getRandomValues(destination)
+  } else if (
+    typeof msCrypto != 'undefined' &&
+    typeof msCrypto.getRandomValues == 'function'
+  ) {
+    return msCrypto.getRandomValues(destination)
+  }
+  return destination
```
jahtalab: I think here we shouldn't return `destination` if we can't generate one.

packages/rum-core/src/common/utils.js
```diff
@@ -34,33 +33,45 @@ function isCORSSupported() {
...
+
+/**
+ * Random number generator using crypto.getRandomValues
+ * uses msCrypto for supporting IE 11
+ * https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
+ */
+const destination = new Uint8Array(16)
+
+function rng() {
+  if (typeof crypto != 'undefined' && crypto.getRandomValues) {
```
jahtalab: nit: maybe check the typeof `getRandomValues` here as well.

package-lock.json
```diff
@@ -2735,8 +2735,14 @@
       "requires": {
         "error-stack-parser": "^1.3.5",
         "opentracing": "^0.14.3",
-        "promise-polyfill": "^8.1.3",
-        "uuid": "^3.1.0"
+        "promise-polyfill": "^8.1.3"
+      },
+      "dependencies": {
+        "uuid": {
```
jahtalab: Should `uuid` be added here?

package-lock.json
```diff
@@ -21365,6 +21388,12 @@
             "glob": "^7.1.3"
           }
         },
+        "semver": {
```
jahtalab: I think this might be a mistake as well

vigneshshanmugam: I get this in my local whenever i do `npm install`, can you do a clean install and check if its coming for you?

packages/rum-core/src/common/utils.js
vigneshshanmugam: its better to do this instead of undefined as our logic would still work for creating spans and transactions. We check the validity of the header here 
https://github.com/elastic/apm-agent-rum-js/blob/1501b11d6820cfbbaaa424320f75142777de3255/packages/rum-core/src/common/utils.js#L97-L103

package-lock.json
vigneshshanmugam: its a mistake, let me fix it.

package-lock.json
jahtalab: Please remove this one, I will also clean to make sure it's not added later.

packages/rum-core/src/common/utils.js
jahtalab: But we use the `generateRandomId` in a few different places (e.g. error id), even for spans and transactions the `isDtHeaderValid` only checks it if we should add the header or not. but the ids of could still be invalid.

Having said that, I'm not against having it as is (it might actually be useful for #249) but I think we'll need to check ids at some point to not send transaction/spans/errors with invalid ids.

vigneshshanmugam: I guess, if the browser is not supported, we have to take another route and not even create/send any events. 

Yes, lets do this separately.

package-lock.json
vigneshshanmugam: There is a bug in master,  opened a new PR to fix this #713

package-lock.json
```diff
@@ -18331,6 +18330,14 @@
         "tough-cookie": "~2.4.3",
         "tunnel-agent": "^0.6.0",
         "uuid": "^3.3.2"
+      },
+      "dependencies": {
+        "uuid": {
```
jahtalab: There's still `uuid` here, we don't use it anywhere else so I think we should remove these.

vigneshshanmugam: This is correct, please check the source where its originating from.  Its a dependency of other module.

PR #709, opened 2020-03-17
ci(jenkins): enhance commit linting

v1v:
### What

- Extend the scope with the explicit list of scopes.
- Lint the PR title.

### Why

Ensure merge to master is not broken. and backward compatible with how we have been doing so far.


package.json
```diff
@@ -33,8 +33,7 @@
   },
   "commitlint": {
     "extends": [
-      "@commitlint/config-conventional",
-      "@commitlint/config-lerna-scopes"
+      "@commitlint/config-conventional"
```
vigneshshanmugam: If we are doing the config via `commitlint.config,js`, lets move everything there and remove the config from `package.json`

commitlint.config.js
```diff
@@ -0,0 +1,40 @@
...
+  utils: { getPackages }
+} = require('@commitlint/config-lerna-scopes')
+
+module.exports = {
+  extends: ['@commitlint/config-lerna-scopes'],
+  rules: {
+    'scope-enum': async ctx => [
+      2,
+      'always',
+      [...(await getPackages(ctx)), 'benchmarks', 'dev', 'jenkins', 'release']
```
vigneshshanmugam: would also add `deps` as its frequently used for updating package dependencies.

commitlint.config.js
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+
+// See https://gitmemory.com/issue/conventional-changelog/commitlint/629/497624131
+const {
+  utils: { getPackages }
+} = require('@commitlint/config-lerna-scopes')
+
+module.exports = {
+  extends: ['@commitlint/config-lerna-scopes'],
```
vigneshshanmugam: this is not required, as we are using the `getPackages` call manually and both does the same thing.

PR #713, opened 2020-03-20
chore(deps): remove unused deps and fix lockfile

vigneshshanmugam:
+ came from #709

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2495,17 +2495,6 @@
         "conventional-changelog-conventionalcommits": "4.2.1"
       }
     },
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -21365,6 +21320,12 @@
             "glob": "^7.1.3"
           }
         },
+        "semver": {
```
jahtalab: `semver` is still here, they are persistent 😄

vigneshshanmugam: Yes, it should be there. With out recent addition of `commitlint`, the pakcage-lock is wrong.

Issue #715, opened 2020-03-20
CI: The test process should adjust the pipeline according to changes in the PR


jahtalab:
Sometimes depending on the changes made in a PR, there is no need to run all the checks for it. As a first step, if a PR only contains changes to doc files then only the docs build should run and nothing else.

@elastic/observablt-robots , a similar change was recently introduced for kibana, can we also do this?

cachedout:
It seems like we should be able to achieve this with a combination of a dedicated docs build stage that [only runs when certain files are changed ](https://jenkins.io/doc/book/pipeline/syntax/#when) and a [post section in that stage ](https://jenkins.io/doc/pipeline/tour/post/)that bails out of the pipeline after it's done.

@kuisathaverat and @v1v do you have other/better ways to achieve this?

vigneshshanmugam:
Other systems solve this by following this pattern `[skip ci]`. At-least Travis supports this patter. Can we do something similar for Jenkins?

v1v:
We got something lighter for the apm-server, the build will happen always
but all the stages that are not required to be launched for the docs are
skipped.

In fact we got something similar but the other way around for the benchmark
in the apmagent rum. :)


Not sure how it has done for the Kibana, but the above approach could be
done pretty quickly.

What do you think?

Cheers

jahtalab:
@v1v , that sounds good! Are the skipped checks still visible in the PR?

v1v:
> @v1v , that sounds good! Are the skipped checks still visible in the PR?

Yes, with a gray color in the CI Pipeline, but they won't be visible in the GitHub PR checks

v1v:
> @kuisathaverat and @v1v do you have other/better ways to achieve this?

@cachedout, that's what we've done for other cases, see:
- https://github.com/elastic/apm-server/blob/174985dcebae11cf41c91be5fceb3a2622ca8d25/Jenkinsfile#L78-L79 to look for the changelist
- https://github.com/elastic/apm-server/blob/master/Jenkinsfile#L98-L104 for the when logic in each stage

PR #716, opened 2020-03-20
ci(jenkins): speed up builds with only docs changes

v1v:
### What

- Any builds for any branch or PR that only contains changes related to the asciidoc then it will skip all the stages but the `checkout`

### Why

Docs validation happens with the `elasticsearch-ci/docs` therefore there is no need to validate anything in our end.

### Issues

Closes https://github.com/elastic/apm-agent-rum-js/issues/715

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,14 +54,13 @@ pipeline {
             stash allowEmpty: true, name: 'source', useDefaultExcludes: false
             script {
               dir("${BASE_DIR}"){
-
                 // Look for changes related to the benchmark, if so then set the env variable.
-                def regexps =[
+                def patternList =[
                   '^packages/.*/test/benchmarks/.*',
                   '^scripts/benchmarks.js',
                   '^packages/rum-core/karma.bench.conf.js'
                 ]
-                env.BENCHMARK_UPDATED = isGitRegionMatch(patterns: regexps)
+                env.BENCHMARK_UPDATED = isGitRegionMatch(patterns: patternList)
 
                 // Skip all the stages except docs for PR's with asciidoc changes only
                 env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.asciidoc' ], comparator: 'regexp', shouldMatchAll: true)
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -52,15 +52,19 @@ pipeline {
...
+                // Look for changes related to the benchmark, if so then set the env variable.
                 def regexps =[
                   '^packages/.*/test/benchmarks/.*',
                   '^scripts/benchmarks.js',
                   '^packages/rum-core/karma.bench.conf.js'
                 ]
                 env.BENCHMARK_UPDATED = isGitRegionMatch(patterns: regexps)
+
+                // Skip all the stages except docs for PR's with asciidoc changes only
+                env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.asciidoc' ], comparator: 'regexp', shouldMatchAll: true)
```
v1v: `shouldMatchAll: true` will help to filter those changes that contain code changes.

vigneshshanmugam: do we need the `comparator` here?  Asking because the benchmarks regex didn't have it.

v1v: Good question, the `regexps` variable name is a bit misleading in the benchmark, the benchmark is not a regex but a glob. In other words,  `isGitRegionMatch` got the default comparator as glob rather than regexp. I'll change the variable name to ensure we don't get confused in the future.

Jenkinsfile
```diff
@@ -69,6 +72,10 @@ pipeline {
         Lint the code.
         */
         stage('Lint') {
+          when {
+            beforeAgent true
+            expression { return env.ONLY_DOCS == "false" }
```
cachedout: What if we run the docs build first (or close to first) and if we detected it was a docs-only build, we just exited the pipeline right after it run? Would that be less code to maintain or would it make the behavior too rigid? WDYT?

v1v:
@vigneshshanmugam and @cachedout  I'll explore your thoughts/comments

Jenkinsfile
vigneshshanmugam: Thanks Victor for the details. Sounds good.

PR #721, opened 2020-03-24
docs: add performance tuning docs [skip ci]

### LINKED ISSUE #567: Add Performance tuning to documentation
We should explain the [benchmarks](https://github.com/elastic/apm-agent-rum-js/issues/26) in a high level and in terms of the overhead the users can expect on their application.

Good example of docs: [Apm Server](https://www.elastic.co/guide/en/apm/server/current/sizing-guide.html), [Nodejs](https://www.elastic.co/guide/en/apm/agent/nodejs/current/performance-tuning.html)

vigneshshanmugam:
+ fix #567

### DIFF
--- a/docs/performance-tuning.asciidoc
+++ b/docs/performance-tuning.asciidoc
@@ -1,0 +1,52 @@
+[[performance-tuning]]
+== Performance Tuning
+
+There are different ways to optimize/tune the performance of the RUM agent. 
+Which options to adjust depends on whether you are optimizing for speed, memory
+usage, bandwidth or storage.
+
+
+[float]
+[[performance-sampling]]
+=== Sampling
+
+The first knob to reach for when tuning the performance of the agent is <<transaction-sample-rate,`transactionSampleRate`>>.
+Adjusting the sampling rate controls what percent of requests are traced.
+By default, the sample rate is set at `1.0`, meaning _all_ requests are traced
+and sent to the APM server. 
+
+The sample rate will impact all four performance categories,
+so simply turning down the sample rate is an easy way to improve performance.
+
+Here's an example of setting the sample rate to 20%:
+
+[source,js]
+----
+require('@elastic/apm-rum').init({
+  transactionSampleRate: 0.2
+})
+----
+
+[float]
+[[performance-breakdown-metrics]]
+=== Breakdown Metrics
+
+Breakdown metrics help visualize where your application is spending the majority of
+its time. The <<breakdown-metrics,`breakdownMetrics`>> config controls whether metrics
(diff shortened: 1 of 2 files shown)


docs/performance-tuning.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+Which options to adjust depends on whether you are optimizing for speed, memory
+usage, bandwidth or storage.
+
+
+[float]
+[[performance-sampling]]
+=== Sampling
+
+The first knob to reach for when tuning the performance of the agent is <<transaction-sample-rate,`transactionSampleRate`>>.
+Adjusting the sampling rate controls what percent of requests are traced.
```
jahtalab, line 14: nit: "percentage" or "ratio"

docs/performance-tuning.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+By default, the sample rate is set at `1.0`, meaning _all_ requests are traced
+and sent to the APM server.
```
jahtalab, line 16: We probably should explain what exactly is the difference between sample and unsampled transactions, e.g. that unsampled transactions will still be reported but without spans, etc.

docs/performance-tuning.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+The sample rate will impact all four performance categories,
+so simply turning down the sample rate is an easy way to improve performance.
+
+Here's an example of setting the sample rate to 20%:
+
+[source,js]
+----
+require('@elastic/apm-rum').init({
+  transactionSampleRate: 0.2
+})
```
jahtalab, line 27: I think it's better to use import here.

docs/performance-tuning.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+[[performance-central-config]]
+=== APM Agent Configuration 
+
+Activate agent configuration via Kibana to start fetching new configuration
+changes from the APM server during the agent initialization phase.
+
+Setting the config option <<central-config,`centralConfig`>> to `true`
+incurs the cost of one additional HTTP request when the agent is 
+initialized and generates more load to the APM server. It is recommended
+to disable this config when the instrumented application is under heavy load.
```
jahtalab, line 52: We should probably mention it's disabled by default because of these reasons.

docs/performance-tuning.asciidoc
bmorelli25: Maybe something like: The Agent will still record the overall time and result of unsampled transactions,
but not context information, labels, or spans.

PR #722, opened 2020-03-24
chore(benchmarks): more microbenchmarks to find hot path

### LINKED ISSUE #720: Enhance micro benchmarks to help with performance tuning
+ Our micro benchmarks does not help in finding our the hot paths of the code which is not good as we cannot recommend our users when there is a performance problem (CPU/Memory). 

+ Also to help with our performance tuning documentation, It its useful to have the benchmarks ready and understand the impact. 

doc - #567

vigneshshanmugam:
+ fix #720

### DIFF
--- a/packages/rum-core/test/benchmarks/capture-navigation.bench.js
+++ b/packages/rum-core/test/benchmarks/capture-navigation.bench.js
@@ -24,29 +24,29 @@
  */
 import { captureNavigation } from '../../src/performance-monitoring/capture-navigation'
 import Transaction from '../../src/performance-monitoring/transaction'
-import { PAGE_LOAD } from '../../src/common/constants'
+import { PAGE_LOAD, ROUTE_CHANGE } from '../../src/common/constants'
 
 suite('CaptureNavigation', () => {
   const options = {
-    managed: true,
     startTime: 0,
     transactionSampleRate: 1
   }
+  const endTime = 10000
 
-  benchmark('page-load transaction', () => {
+  benchmark('hard navigation', () => {
     const pageLoadTr = new Transaction('/index', PAGE_LOAD, options)
     pageLoadTr.captureTimings = true
-    pageLoadTr.end(5000)
+    pageLoadTr.end(endTime)
     captureNavigation(pageLoadTr)
   })
 
-  benchmark('other transaction', () => {
+  benchmark('soft navigation', () => {
     /**
      * Does not include navigation timing spans and agent marks
      */
-    const nonPageLoadTr = new Transaction('/index', 'custom', options)
+    const nonPageLoadTr = new Transaction('/index', ROUTE_CHANGE, options)
     nonPageLoadTr.captureTimings = true
-    nonPageLoadTr.end(5000)
+    nonPageLoadTr.end(endTime)
     captureNavigation(nonPageLoadTr)
   })
 })
--- a/packages/rum-core/test/benchmarks/capture-navigation.bench.js
+++ b/packages/rum-core/test/benchmarks/capture-navigation.bench.js
@@ -22,50 +22,31 @@
  * THE SOFTWARE.
  *
  */
-
-import { createServiceFactory } from '../../src/index'
+import { captureNavigation } from '../../src/performance-monitoring/capture-navigation'
+import Transaction from '../../src/performance-monitoring/transaction'
 import { PAGE_LOAD } from '../../src/common/constants'
 
-suite('TransactionService', () => {
-  const serviceFactory = createServiceFactory()
-  const transactionService = serviceFactory.getService('TransactionService')
+suite('CaptureNavigation', () => {
+  const options = {
+    managed: true,
+    startTime: 0,
+    transactionSampleRate: 1
+  }
 
-  benchmark('managed sampled transaction overhead', async () => {
-    const tr = transactionService.startTransaction('/index', PAGE_LOAD, {
-      managed: true,
-      startTime: 0,
-      transactionSampleRate: 1
-    })
-    await tr.end()
+  benchmark('page-load transaction', () => {
+    const pageLoadTr = new Transaction('/index', PAGE_LOAD, options)
+    pageLoadTr.captureTimings = true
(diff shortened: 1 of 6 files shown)


vigneshshanmugam:
@jahtalab Let me know if you feel there are other parts of code worth measuring.

packages/rum-core/test/benchmarks/index.js
```diff
@@ -25,19 +25,22 @@
...
-    lastSpanEnd = 10 + i
-    span.end(lastSpanEnd)
+    const span = tr.startSpan(`name`, `type`, {
+      startTime: getRandomInRange(0, i)
+    })
+    span.end(getRandomInRange(i, i + 1))
   }
 
-  tr.end(lastSpanEnd + 100)
+  tr.end(noOfSpans + 10)
```
jahtalab: This doesn't seem correct to me. end method should be called with a correct time, not with number of spans

packages/rum-core/test/benchmarks/capture-navigation.bench.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  benchmark('page-load transaction', () => {
+    const pageLoadTr = new Transaction('/index', PAGE_LOAD, options)
+    pageLoadTr.captureTimings = true
+    pageLoadTr.end(5000)
+    captureNavigation(pageLoadTr)
+  })
+
+  benchmark('other transaction', () => {
```
jahtalab: nit: use `custom transaction`

packages/rum-core/test/benchmarks/capture-navigation.bench.js
```diff
@@ -0,0 +1,52 @@
...
+ * THE SOFTWARE.
+ *
+ */
+import { captureNavigation } from '../../src/performance-monitoring/capture-navigation'
+import Transaction from '../../src/performance-monitoring/transaction'
+import { PAGE_LOAD } from '../../src/common/constants'
+
+suite('CaptureNavigation', () => {
+  const options = {
+    managed: true,
```
jahtalab: I think being managed or unmanaged doesn't affect these benchmarks, isn't that right?

packages/rum-core/test/benchmarks/capture-navigation.bench.js
vigneshshanmugam: We don't capture navigation for custom transactions. thats why i named it this way.

packages/rum-core/test/benchmarks/index.js
vigneshshanmugam: The logic is correct. `end` methods just needs a number so we can control the duration in our way and it would look like this 
```
const tr = generateTestTransaction(2)
tr.duration() // 12

// 2 created spans would be within this time range
```

packages/rum-core/test/benchmarks/capture-navigation.bench.js
vigneshshanmugam: Yes, kept it explicitly as `managed` since `page-load` is a managed transaction.

packages/rum-core/test/benchmarks/capture-navigation.bench.js
jahtalab: I think the title is not very descriptive, something that would indicated it doesn't capture navigation timing would be better.

packages/rum-core/test/benchmarks/index.js
jahtalab: the syntax is correct but logically it doesn't make sense to bound the end of the transactions to the number of spans. Actually I think the previous logic is simpler and works just fine without having to introducing randomness into this. Is there any reason you chose this algorithm?

packages/rum-core/test/benchmarks/capture-navigation.bench.js
jahtalab: But, `page-load` being managed doesn't play a part in this benchmark, we could have a unmanaged transaction with the type `page-load` and the result of this benchmark would be the same.

packages/rum-core/test/benchmarks/index.js
vigneshshanmugam: I needed randomness while testing the grouping logic `groupSmallContinuouslySimilarSpans`. The previous logic required a intermediate variable `lastSpanEnd` which can be eliminated since its simpler to use `noOfSpans` as lastSpanEnd is within the range since our span.end makes sure its within the time range. 

Either way its not super important, we can simply throw in a fixed number and end the transaction.

packages/rum-core/test/benchmarks/capture-navigation.bench.js
vigneshshanmugam: I can remove the flag.

packages/rum-core/test/benchmarks/capture-navigation.bench.js
vigneshshanmugam: I will change it to `non-page-load` as we till capture all timings(resource, user) except the navigation timing ones.

vigneshshanmugam: I have renamed these to `hard` and `soft` navigation as they are better than tying these to specific transactions

packages/rum-core/test/benchmarks/index.js
jahtalab: @vigneshshanmugam , Would you please elaborate on why randomness is need for testing `groupSmallContinuouslySimilarSpans`. If that's necessary I think we should at least add a comment here to explain why. At the moment it looks more like a mistake that we're using `noOfSpans + 10` as the end time of the transaction.

vigneshshanmugam: I am happy to hear if you have any ideas here. The randomness was required so we can group few spans lave rest as it is and to keep them differentiated. In any case, I am happy to tune the method if looks complicated.

vigneshshanmugam: We have discussed offline and agreed on removing the randomness and create a fixed number of groupable/ungroupable spans.

PR #724, opened 2020-03-25
fix(rum-core): filter API calls from resource spans

### LINKED ISSUE #723: API calls are added as resource timing spans in IE
+ `external` span types such as `xmlhttprequest` and `fetch` spans are added as resource timing spans due to a bug in our filtering logic.

vigneshshanmugam:
+ fix #723

### DIFF
--- a/packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
+++ b/packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
@@ -316,4 +316,24 @@ describe('Capture hard navigation', function() {
     expect(tr.marks.agent.firstContentfulPaint).toBeUndefined()
     unMock()
   })
+
+  it('should not add API calls as resource timing spans', function() {
+    const unMock = mockGetEntriesByType()
+    const tr = new Transaction('test', PAGE_LOAD, {
+      startTime: transactionStart
+    })
+    const apiSpan = tr.startSpan('GET http://ajax-filter.test', 'external.http')
+    apiSpan.end()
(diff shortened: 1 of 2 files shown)


packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
```diff
@@ -316,4 +316,24 @@ describe('Capture hard navigation', function() {
...
+      startTime: transactionStart
+    })
+    const apiSpan = tr.startSpan('GET http://ajax-filter.test', 'external.http')
+    apiSpan.end()
+    tr.captureTimings = true
+    tr.end(transactionEnd)
+    captureNavigation(tr)
+
+    const filteredRTSpan = tr.spans.filter(
+      span => span.name === 'http://ajax-filter.test'
```
jahtalab: Changing this check into something like `span.name.indexOf('http://ajax-filter.test')>-1` would be better, e.g. if there are changes to the resource timing span names that would affect this but might not necessarily break this test.

vigneshshanmugam: we have to test also the type in that case as it would match the API span name

vigneshshanmugam:
The build ran https://travis-ci.org/github/elastic/apm-agent-rum-js/builds/666835927, just the status is not getting reported for some reason.

vigneshshanmugam:
@jahtalab there is an incident from Travis https://www.traviscistatus.com/incidents/rx6fhs3wqcln

Issue #636, opened 2020-02-19
Capture all spans on route end in angular integration


vigneshshanmugam:
We have decided to use the Native promises when available to reduce the bundle size #629, but due to the way native promises are scheduled, we would end the transaction sooner and might not capture the spans all the time that is caused due to the route-change event. 

We have to fix this by doing something similar to #404 and fix the E2E test TODO.

PR #729, opened 2020-03-31
feat(rum-core): end transaction after scheduled browser frames

### LINKED ISSUE #404: ending automatic transactions in a graceful way
Our transaction end detection logic for all transactions type except `page-load` is a bit eager now and might fail to capture spans that happen in the frames that happen after the nextTick. 

During the route-change events, there are multiple ways users would be invoking external API calls, can be sync, concurrent/parallel or even lazy load some components based on the API calls. Right now our framework integrations rely on the internal lifecycle events of the respective frameworks and tries to end the transaction as soon as the component is mounted to the DOM. 

With this approach, we might be able to only capture only the API calls that are requested during the component mounting and might fail to capture calls that are done after the component is mounted to the DOM. 

### Solution 
+ Instead of closing the transaction as soon as the children components are mounted or when the route change navigation is done, We could actively wait and listen for few frames (for now we can listen for 5 successive ticks/animation frames) and decide if we want to close the transaction. 
+ If there are no calls within those 5 ticks, we close the transaction immediately and send it to the APM server. 
+ If there are any spans, assume a call happens on the 3rd frame, then we schedule another 5 frames and end the transaction accordingly. 
+ In order to reduce the noise in the transaction, our transaction would always get time shifted to the last active transaction. Ex - if there are 2 spans that happens in the 2nd tick after the transaction start, we end the transaction with the end time of the longest span and send it to the server. 
+ With our filtering logic, transaction without any spans will not be reported to the server.

vigneshshanmugam:
+ fix #404 #636 
+ Schedules `n` number of frames for every navigation and then ends the transaction once the browser has finished rendering the relevant changes inside the frame. 

### Pros
+ Configurable, We can easy tune it for different kinds of transaction and schedule different frames

### Cons
+ Will not work as expected when page is not visible as it depends on `rAF` and browser does not fire requestAnimationFrame unless the page is foregrounded again. Having a hard timeout would fix it. 
+ Might Introduce state inside transaction as we have to clear previous scheduled frames when we want to schedule few more again.  We might end up doing in another way, But that would mean we wont be able to use have two transactions running at same time. 

TODO
+ [ ] Add tests
+ [ ] Add code documentation

### DIFF
--- a/packages/rum-core/src/common/after-frame.js
+++ b/packages/rum-core/src/common/after-frame.js
@@ -23,38 +23,26 @@
  *
  */
 
-export function routeHooks(router, apm) {
-  let transaction
+export default class AfterFrame {
+  constructor() {
+    this.current = null
+  }
 
-  router.beforeEach((to, from, next) => {
-    const matched = to.matched || []
-    let path = to.path
-    /**
-     * Get the last matched route record which acts as stack when
-     * route changes are pushed and popped out of the stack
-     *
-     * Also account for the slug pattern on the routes, to.path always
-     * resolves the current slug param, but we need need to
-     * use the slug pattern for the transaction name
-     */
-    if (matched.length > 0) {
-      path = matched[matched.length - 1].path || path
-    }
-    transaction = apm.startTransaction(path, 'route-change', {
-      managed: true,
-      canReuse: true
+  next(times, callback) {
+    this.current = requestAnimationFrame(() => {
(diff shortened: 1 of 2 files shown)


vigneshshanmugam:
We have discussed this solution in detail and agreed on the other solution as its simpler and would work for most of the use cases. We might do this at somepoint,  if we really need this as it could help with capturing frames per second and time to interactivity.

PR #730, opened 2020-04-01
feat(rum-core): end spa navigations after browser frame

### LINKED ISSUE #404: ending automatic transactions in a graceful way

vigneshshanmugam:
+ Alternative solution to #729 
+ fixes #404 #636 
+ The solution is based on scheduling the transaction to be ended after the browser paints the new frame ( includes script, layout, style and paint). This ensures that if there is any network activity or if the underlying framework yields to the browser to do the work (style, paint). We include those timings as well which would help us measure most of the layout changes. 

There is a detailed [blog post ](https://nolanlawson.com/2018/09/25/accurately-measuring-layout-on-the-web/)explaining different ways to schedule a call after next frame. We are using the `rAF` + `setTimeout` technique as we want the call to get executed even if the tab is opened in invisible mode (cmd + click). 

### Pros
+ Simpler and Easier to reason about, don't need to keep the state on the frame running and scheduling problems
+ Gets called even if the current tab is invisible but with delays (due to the use of timers). `RAF` never gets called for invisible tabs. 

### Cons
+ We wont be able to measure if the user does `setTimeout(task, 1000)` since it could be scheduled by the browser in next to next frames and we only measure the next frame after the navigation. This should be okay as it is not a usual thing. 
+ we might increase the transaction duration. by `16.67ms` worst case due to the use of rAF, but we would be able to capture something the frameworks/user is doing during layout so might be worth it. 

Personal Opinion, I like this solution better than the other one due to its simplicity and solving for 99% of the use cases. Was able to work for most of the cases when the call is scheduled in
`setImmediate`, `setTimeout(call, 0)`, `requestAnimationFrame`, `requestIdleCallback`, `Promise.resolve`, `MessageChannel` during manual testing.


jahtalab:
- test for `afterFrame` to investigate behaviour on different browsers
- We also need test for react.lazy

vigneshshanmugam:
jenkins run the tests please

vigneshshanmugam:
Chrome 76 seems to have an issue with scheduling tasks. test passes on all browsers except for this one.

packages/rum-react/test/e2e/with-router/switch.e2e-spec.js
```diff
@@ -59,4 +58,32 @@ describe('Using Switch component of react router', function() {
...
+
+    const pageLoadTransaction = transactions[0]
+    expect(pageLoadTransaction.type).toBe('page-load')
+    expect(pageLoadTransaction.name).toBe('/notfound')
+
+    const routeTransaction = transactions[1]
+    expect(routeTransaction.name).toBe('/lazy')
+    expect(routeTransaction.type).toBe('route-change')
+
+    expect(routeTransaction.spans.length).toBe(1)
```
jahtalab, line 86: Shouldn't the request that fetches the separate JS bundle be captured from the resource timing?

vigneshshanmugam:
@jahtalab 
> Would you please comment about the Chrome 76 scheduling issue here?

Its not chrome 76 scheduling issue, I added more details with comments in the test itself. It's hard to know test this scenario for different browsers, thats why the test checks for 2 or 3 more, Also its not a common scenario - No one of the frameworks that I have tested(React, Vue, Angular, Preact) use this double timeouts. 

There is no underlying issue with our approach, instead if the user really uses this way of rendering something, Neither the frameworks not our library would be able to track it. Unless we monkey patch.

packages/rum-core/test/common/after-frame.spec.js
```diff
@@ -0,0 +1,70 @@
...
+      attributes: true
+    })
+    // Trigger mutation observer callback
+    document.body.setAttribute('test', 'test')
+  })
+
+  it('should be called after tasks', done => {
+    let count = 0
+    afterFrame(() => {
+      expect(count).toBeGreaterThanOrEqual(2)
```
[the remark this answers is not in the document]
vigneshshanmugam: There is a comment below that explains it. I can move this the `double timeout` test as a edge case test and add comments.

packages/rum-react/test/e2e/with-router/switch.e2e-spec.js
vigneshshanmugam: It wont, since use the pre bundled functional component, I will create a new lazy component and add it.

vigneshshanmugam:
Wrote a detailed comment on the double setTimeout tests https://github.com/elastic/apm-agent-rum-js/pull/730/files#diff-c6439d182e960bb8b3e580ba6c8e250aR61-R88 and also moved them to edge case test.

vigneshshanmugam:
Lazy component mount cannot be solved with this PR as our Route component wont be even kicked in(Transaction will not start until React fetches all underlying dependencies) when the dependencies are being fixed.  Also this PR is not for addressing that issue, instead measuring everything that happens before the browser renders the next frame in a better way. 

However I added tests when lazy components are fetched inside a component that is child of our  `ApmRoute` component. 

Lazy components can be still measured effectively by the technique already mentioned here - https://www.elastic.co/guide/en/apm/agent/rum-js/current/react-integration.html#_instrumenting_lazy_loaded_routes

PR #734, opened 2020-04-03
feat(rum-angular): support differential loading with CLI

### LINKED ISSUE #607: create new build targets compatible with Angular Packaging Format
+ Currently we expose `elastic/apm-rum-angular` with default target `IE11` which transpiles the native ES6 classes to ES5 compatible versions. When used in Angular projects bootstrapped with Angular CLI, the default target is `es2015/es6` and our `ApmErrorHandler` plugin wont work since native class is not compatible with transpiled one. 

```
Unhandled Promise rejection: Class constructor ErrorHandler cannot be invoked without 'new' ; Zone: <root> ; Task: Promise.then ; Value: TypeError: Class constructor ErrorHandler cannot be invoked without 'new'
    at new ApmErrorHandler (error-handler.js:8)
    at _createClass (core.js:30459)
    at _createProviderInstance (core.js:30426)
    at initNgModule (core.js:30332)
    at new NgModuleRef_ (core.js:31561)
    at createNgModuleRef (core.js:31544)
    at Object.debugCreateNgModuleRef [as createNgModuleRef] (core.js:44911)
    at NgModuleFactory_.create (core.js:46088)
    at core.js:40535
    at ZoneDelegate.invoke (zone-evergreen.js:359) TypeError: Class constructor ErrorHandler cannot be invoked without 'new'
    at new ApmErrorHandler (http://localhost:4200/vendor.js:108489:26)
    at _createClass (http://localhost:4200/vendor.js:71894:20)
    at _createProviderInstance (http://localhost:4200/vendor.js:71861:26)
    at initNgModule (http://localhost:4200/vendor.js:71767:32)
    at new NgModuleRef_ (http://localhost:4200/vendor.js:72877:9)
    at createNgModuleRef (http://localhost:4200/vendor.js:72860:12)
    at Object.debugCreateNgModuleRef [as createNgModuleRef] (http://localhost:4200/vendor.js:85070:12)
    at NgModuleFactory_.create (http://localhost:4200/vendor.js:86203:25)
    at http://localhost:4200/vendor.js:80903:64
    at ZoneDelegate.invoke (http://localhost:4200/polyfills.js:6074:26)
```

Easiest way to fix the problem is to add two new build targets, one for `es5` and one for `es6` depending on the [Angular Packaging Format](https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview)

More details on this issue #603

vigneshshanmugam:
+ fixes #607 

+ Angular CLI builds the application using the target `es2015` as described in here https://angular.io/guide/deployment#differential-builds, which means our builds with `module` target wont work since we always transpile using the lowest version `IE 11`

+ There is a strong reason why we use `IE 11` as the least target since we do not want our users to transpile the dependencies code(If users are looking for more size wins, its better to follow this path as well) and instead make it work out of the box.  

+ With this PR, All webpack users would still be unaffected as it uses `module` target and can perform Tree shaking and rest like before. Angular CLI users would make use of the newly specified target `es2015` and can utilize the differential loading technique (module/nomodule) as explained here -  https://angular.io/guide/deployment#configuring-differential-loading

```
Webpack (module) - ESM + ES5 
Angular CLI (es2015) - ESM + ES2015
```

If webpack users wants to utilize the differential loading and do not want transpiled ES5 version. They can configure `resolve.mainFields` to points to 
+ `source` - No transpilation done
+ `es2015` - Target browsers that supports ES Modules 

```js
...  resolve: {
        mainFields: ['source', 'es2015', 'module', 'main'],
      }
```

Since we don't have tests with Angular CLI, Manual tests are performed with Angular CLI and it works as expected with both `ApmService` and `ApmErrorHandler`.

### DIFF
--- a/packages/rum-angular/package.json
+++ b/packages/rum-angular/package.json
@@ -6,6 +6,7 @@
   "license": "MIT",
   "main": "dist/lib/index.js",
   "module": "dist/es/index.js",
+  "es2015": "dist/es2015/index.js",
   "source": "src/index.ts",
   "files": [
     "dist",
@@ -27,6 +28,7 @@
     "build": "run-p build:main build:module",
     "build:main": "BABEL_ENV=BROWSER_PROD npx babel -x '.ts' src -d dist/lib ",
     "build:module": "BABEL_ENV=BROWSER_ESM_PROD npx babel -x '.ts' src -d dist/es",
+    "build:es2015": "BABEL_ENV=BROWSER_ESM_ES2015 npx babel -x '.ts' src -d dist/es2015",
     "build:e2e": "npm run script buildE2eBundles packages/rum-angular/test/e2e",
     "test:unit": "npm run script runUnitTests packages/rum-angular",
     "test:e2e": "npm run script runE2eTests packages/rum-angular/wdio.conf.js",
--- a/dev-utils/build.js
+++ b/dev-utils/build.js
@@ -36,7 +36,8 @@ const BUNDLE_TYPES = {
   NODE_ESM_PROD: 'NODE_ESM_PROD',
   BROWSER_DEV: 'BROWSER_DEV',
   BROWSER_PROD: 'BROWSER_PROD',
-  BROWSER_ESM_PROD: 'BROWSER_ESM_PROD'
+  BROWSER_ESM_PROD: 'BROWSER_ESM_PROD',
+  BROWSER_ESM_ES2015: 'BROWSER_ESM_ES2015'
 }
 
 const {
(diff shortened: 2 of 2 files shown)


packages/rum-angular/package.json
```diff
@@ -27,6 +28,7 @@
     "build": "run-p build:main build:module",
```
chriscarpenter12, line 28: Why not add `build:es2015` here on `lerna run build`?

vigneshshanmugam: wow good catch, we used to have `build:*`, totally forgot. thanks for the review @chriscarpenter12  👍

vigneshshanmugam: added a new PR #736

PR #742, opened 2020-04-08
fix(rum-react): memoizing function component to avoid unmount

cauemarcondes:


vigneshshanmugam:
jenkins run the tests please

packages/rum-react/src/get-with-transaction.js
```diff
@@ -40,6 +41,45 @@ function isReactClassComponent(Component) {
   return !!(prototype && prototype.isReactComponent)
 }
 
+const getMemoizedApmComponent = memoize((Component, apm, name, type) => {
```
cauemarcondes, line 44: @dgieselaar WDYT about memoize a function that returns a Component? I'm not sure about it. I need to do it to avoid creating a new ApmComponent every time a change happens in the URL.

packages/rum-react/src/get-with-transaction.js
```diff
@@ -78,84 +117,52 @@ function getWithTransaction(apm) {
...
-             * components since the parent component is mounted after child
-             */
-            this.transaction = apm.startTransaction(name, type, {
-              managed: true,
-              canReuse: true
-            })
-          }
-
-          componentDidMount() {
+        ApmComponent = React.useCallback(
```
cauemarcondes, line 123: In case of hooks is available we can use useCallBack to memoize the component.

packages/rum-react/src/get-with-transaction.js
dgieselaar: Well, it's a memory leak I think. I would at least tie it to the lifecycle of a container (e.g. by using `getDerivedStateFromProps`). I think you also might be able to rewrite this to a component that takes `apm`, `name` and `type` as props, and then you don't need memoization at all.

PR #744, opened 2020-04-08
ci(jenkins): release automation followups

v1v:
- Workspace is not anymore a shallow cloned but contains the whole history. Otherwise, the changelogs will not contain all the changes.
- Send email notification when the release has been published.
- Update RELEASE.md docs with more details.
- Enable verbose output when to publish and version. See [here](https://github.com/lerna/lerna/blob/5e790e5d661ee95e0dbc2160e6509e6691b8b9e3/core/project/README.md#command-specific-configuration)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -9,10 +9,14 @@
       "allowBranch": ["master", "4.x"],
       "conventionalCommits": true,
       "message": "chore(release): publish",
-      "gitRemote": "upstream"
+      "gitRemote": "upstream",
+      "loglevel": "verbose"
     },
     "run": {
       "stream": true
+    },
+    "publish": {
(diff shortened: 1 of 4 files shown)


lerna.json
```diff
@@ -9,10 +9,14 @@
...
       "conventionalCommits": true,
       "message": "chore(release): publish",
-      "gitRemote": "upstream"
+      "gitRemote": "upstream",
+      "loglevel": "verbose"
     },
     "run": {
       "stream": true
+    },
+    "publish": {
```
vigneshshanmugam, line 18: `publish` internally calls `version` so the above `loglevel` should be enough

Issue #191, opened 2019-02-19
Ability to load the agent asynchronously


vigneshshanmugam:
+ We should allow loading the RUM agent asynchronously and also update the getting started document to load via different options. 
+ Should also figure out the pros and cons of each loading approaches and should document the same.

vigneshshanmugam:
Some notes from our discussion. 

### Solution 1
 Users can add a JS snippet in the head of the page which can both hook in our custom patch logic(fetch, XHR and errors) and also load our agent script dynamically by creating script element
```js
// Code to our minimal custom patching logic
const script = document.createElement('script')
script.src = 'elastic-apm-rum.js' // link to CDN hosted script
script.onload = () => {}
src.async = true
src.onerror = () => {}
src.crossorigin = 'anonymous'
```
##### Open questions
+  need to figure out how to deliver the asset
+ CSP violations
+  Even though the script is downloaded async, It will be executed synchronously. We can also look in to loading techniques through iframe.

PR #745, opened 2020-04-08
docs: add blocking/nonblocking pattern on usage

vigneshshanmugam:
+ As a result of #739 
+ Part of #191
+ Adds a blocking/non-blocking loading patter and also added some couple of caching header strategies users could use when hosting the bundle in the CDN.  
+ Also documented the caveat that DT might not work for calls that happened before the agent script is loaded.

### DIFF
--- a/docs/set-up.asciidoc
+++ b/docs/set-up.asciidoc
@@ -6,27 +6,82 @@ you must first {apm-server-ref-v}/configuration-rum.html[enable the RUM endpoint
 
 Once the APM Server endpoint has been configured, you can:
 
-* <<install-the-agent>>
+* <<usage>>
 * <<configuring-cors>>
 
-[[install-the-agent]]
-=== Install the Agent
+[[usage]]
+=== Usage
 
-There are three ways to use the APM RUM Agent:
+There are three ways to include the APM RUM Agent:
 
-* <<using-package-managers>>: Install the Agent as a dependency to your application.
+APM RUM Agent can be added to your web page in one of two ways
+
 * <<using-script-tags>>: Add a `<script>` tag to the HTML page.
-* <<using-production-build>>: Optimized bundles.
+* <<using-bundlers>>: Bundling the Agent package during the application's build process
 
 [float]
-[[using-package-managers]]
-=== Using Package Managers
+[[using-script-tags]]
+=== Using Script Tags
 
-Install the APM agent for JavaScript as a dependency to your application:
+==== Synchronous / Blocking Pattern
+
+Add a <script> tag to load the bundle and use the `elasticApm` global
+object to initialize the agent:
+
+[source,html]
+----
+<script src="https://<your-cdn-host>.com/path/to/elastic-apm-rum.umd.min-<version>.js" crossorigin></script>
+<script>
+  elasticApm.init({
+    serviceName: '<instrumeted-app>',
+    serverUrl: '<apm-server-url>',
+  })
+</script>
+----
+
+==== Asynchronous / Non-Blocking Pattern
+
+Loading the script asynchronously ensures the agent script will not block other
+resources on the page, however, it will stick block browsers `onload` event.
+
+[source,html]
+----
+<script>
+  ;(function(d, s, c) {
+    var j = d.createElement(s),
+      t = d.getElementsByTagName(s)[0]
+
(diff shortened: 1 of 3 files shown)


docs/set-up.asciidoc
```diff
@@ -6,34 +6,90 @@ you must first {apm-server-ref-v}/configuration-rum.html[enable the RUM endpoint
...
 Once the APM Server endpoint has been configured, you can:
 
-* <<install-the-agent>>
+* <<usage>>
 * <<configuring-cors>>
 
-[[install-the-agent]]
-=== Install the Agent
+[[usage]]
+=== Usage
```
bmorelli25, line 13: Let's try to keep these as actions. Here's a screenshot:
Screen Shot 2020-04-08 at 8 22 49 AM

Instead of "usage" replacing "install the agent", would any of these make sense:
* Import the Agent?
* Load the Agent?
* Initialize the Agent?
* Include the Agent?

docs/set-up.asciidoc
```diff
@@ -6,34 +6,90 @@ you must first {apm-server-ref-v}/configuration-rum.html[enable the RUM endpoint
...
 
-There are three ways to use the APM RUM Agent:
+There are three ways to include the APM RUM Agent:
+
+APM RUM Agent can be added to your web page in one of two ways
```
bmorelli25, line 17: "There are **three** ways..."
_followed by_
"...in one of **two** ways"

Let's get rid of the number to ease maintenance. How about: "There are multiple was to add the APM RUM Agent to your web page:"

docs/set-up.asciidoc
```diff
@@ -110,6 +146,8 @@ plugins: [
 ]
 ----
 
+NOTE: Currently the optimized(minified + gzipped) agent bundle size is about 16KiB.
```
bmorelli25, line 149:
```suggestion
NOTE: Currently the optimized (minified + gzipped) agent bundle size is about 16KiB.
```

docs/set-up.asciidoc
```diff
@@ -6,34 +6,90 @@ you must first {apm-server-ref-v}/configuration-rum.html[enable the RUM endpoint
...
+    t.parentNode.insertBefore(j, t)
+  })(document, 'script', {serviceName: '<instrumeted-app>', serverUrl: '<apm-server-url>'})
+</script>
+----
+
+Even though this is the recommended pattern, there is a caveat
+attached to it as a result of downloading and initializing the
+agent asynchronously. The distributed tracing would not work for requests that
+happened before the agent is initialized as the agent would not be available to
+patch the requests on time.
```
bmorelli25, line 65: Maybe something like this? It simplifies it a bit, but hopefully not too much:

```
Even though this is the recommended pattern, there is a caveat to be aware of.
Because the downloading and initializing of the agent happens asynchronously,
distributed tracing will not work for requests that occur before the agent is initialized.
```

docs/set-up.asciidoc
```diff
@@ -6,34 +6,90 @@ you must first {apm-server-ref-v}/configuration-rum.html[enable the RUM endpoint
...
 
-Install the APM agent for JavaScript as a dependency to your application:
+NOTE: Please download the latest version of RUM agent from https://github.com/elastic/apm-agent-rum-js/releases/latest[GitHub] or
+https://unpkg.com/@elastic/apm-rum/dist/bundles/elastic-apm-rum.umd.min.js[UNPKG]
+and host the file in your Server/CDN before deploying to production with proper
+versioning scheme and set far future `max-age` and `immutable` in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control[cache-control]
+header as the file never changes.
```
bmorelli25, line 71: This is a really long sentence.

How about:

```
NOTE: Please download the latest version of RUM agent from [GitHub] or [UNPKG]
and host the file in your Server/CDN before deploying to production.
Remember to use a proper versioning scheme and set a far future `max-age` and `immutable` in the [cache-control]
header, as the file never changes.
```

docs/set-up.asciidoc
vigneshshanmugam: Forgot to remove the first line, Yeah your suggestion seems to be better.

docs/set-up.asciidoc
vigneshshanmugam: Initialise/Import sounds good.

vigneshshanmugam:
@bmorelli25 I am merging it to do it along with a release.

PR #747, opened 2020-04-14
fix(rum-react): Don't recreate APM component when props are the same

cauemarcondes:
This fix avoids the `apmComponent` to be recreated every time a query param is changed in the URL and fully reload the page.

### DIFF
--- a/packages/rum-react/src/get-apm-route.js
+++ b/packages/rum-react/src/get-apm-route.js
@@ -23,17 +23,30 @@
  *
  */
 
-import React from 'react'
+import React, { Component } from 'react'
 import { Route } from 'react-router-dom'
 import { getWithTransaction } from './get-with-transaction'
 
 function getApmRoute(apm) {
   const withTransaction = getWithTransaction(apm)
 
-  return function ApmRoute(props) {
-    const { path, component } = props
-    const apmComponent = withTransaction(path, 'route-change')(component)
-    return <Route {...props} component={apmComponent} />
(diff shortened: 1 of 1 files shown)


packages/rum-react/src/get-apm-route.js
```diff
@@ -23,17 +23,30 @@
...
+        nextProps.component === this.props.component &&
+        nextProps.path === this.props.path
+      ) {
+        return false
+      }
+      return true
+    }
+    render() {
+      const { path, component, ...props } = this.props
+      const apmComponent = withTransaction(path, 'route-change')(component)
```
vigneshshanmugam, line 47: From quick review,  this does not seem correct as it would trigger countless renders as new components would be created. I will debug more and update the PR with more details.

dgieselaar:
Why was the other one closed? What changed?

cauemarcondes:
> Why was the other one closed? What changed?

Instead of memoizing the components returned by `withTransaction` I made the `ApmRoute` a class component and checked with `shouldComponentUpdate` if the props are the same. @vigneshshanmugam is going to make some tests with `getDerivedStateFromProps`

dgieselaar:
Ok, my first instinct would be to return `ApmComponent` from `getDerivedStateFromProps` when the appropriate props change, let's see what @vigneshshanmugam comes up with.

vigneshshanmugam:
Closing as a result of #748

PR #748, opened 2020-04-14
fix(rum-react): avoid rerendering on query param changes

vigneshshanmugam:
+ Alternative solution to the #747 
+ This avoids the problem that would trigger full page reload when the query param is changed by any of the child components of `<ApmRoute>`

### DIFF
--- a/packages/rum-react/src/get-apm-route.js
+++ b/packages/rum-react/src/get-apm-route.js
@@ -33,15 +33,26 @@ function getApmRoute(apm) {
   return class ApmRoute extends React.Component {
     constructor(props) {
       super(props)
-      const { path, component } = props
+      this.state = {}
+    }
+
+    static getDerivedStateFromProps(nextProps, prevState) {
+      const initial = prevState.apmComponent == null
+      const { path, component } = nextProps
+      const pathChanged = path != prevState.path
+
       /**
-       * Having the ApmComponent on state ensures that we capture the SPA
-       * navigation only when route changes/mount and not on re-renders
-       * Ex: updating query params from child components
+       * Should update the apmComponent state and re-render the component only on
+       * initial mount and on route change.
+       * Ex: Query param changes should not result in new apmComponent
        */
-      this.state = {
-        apmComponent: withTransaction(path, 'route-change')(component)
+      if (initial || pathChanged) {
+        return {
+          apmComponent: withTransaction(path, 'route-change')(component),
+          path
+        }
       }
+      return null
     }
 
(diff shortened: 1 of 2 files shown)


packages/rum-react/src/get-apm-route.js
```diff
@@ -30,10 +30,23 @@ import { getWithTransaction } from './get-with-transaction'
...
+    constructor(props) {
+      super(props)
+      const { path, component } = props
+      /**
+       * Having the ApmComponent on state ensures that we capture the SPA
+       * navigation only when route changes/mount and not on re-renders
+       * Ex: updating query params from child components
+       */
+      this.state = {
+        apmComponent: withTransaction(path, 'route-change')(component)
```
dgieselaar, line 43: what happens when path or component are updated?

vigneshshanmugam: The below test covers the path case https://github.com/elastic/apm-agent-rum-js/pull/748/files#diff-9cf03e1a91b2eaf59d727f74ebc3503fR134

Does it make sense?

dgieselaar: `path` as a component prop, I mean (same for `component`).

vigneshshanmugam: Can you give an example? That would mean, you will have different `ApmRoute` components. 

```jsx
<ApmRoute path="/home" component={MainComponent} />
<ApmRoute path="/func" component={FunctionalComponent} />
```

dgieselaar: Consider:

```
import React from 'react';

function App ( ) {
  const isLoggedIn = useLoggedIn();

  const path = isLoggedIn ? `/u/0` : '/';

  const component = isLoggedIn ? UserHome : AnonHome;

  return (<>
    <AppComponent path={path} component={component}/>
  </>)
}
```

Obviously a contrived example, but I think consumers _can_ use it in this way? Whether it's a good idea is a different discussion 😬

vigneshshanmugam: Good one. Not sure a super valid usecase for route components as its better to keep them static instead of dynamically changing the path and underlying component during each render. I will check if we can provide a solution for this as well.

vigneshshanmugam: I was wrong, I had a quick look. at `Switch` and it seems to do something similar. I changed the solution to use `getDerivedStateFromProps` which should fix this case as well. Thanks 👍

PR #749, opened 2020-04-15
feat(rum-core): use performance observer to capture FCP

### LINKED ISSUE #731: use PerformanceObserver to capture FCP
+ Since Performance Observer is widely supported and chrome is the only browser with support for PaintTiming API, we can move away from using `PerformanceTimeline` API and use the observer as we can get buffered events which is not possible via the timeline. Plus observer doesn't suffer from performance issues like the timeline.

We already use the Observer for `longtask` and `lcp`, so it would be easy to implement.

vigneshshanmugam:
+ fix #731


packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
```diff
@@ -295,28 +287,6 @@ describe('Capture hard navigation', function() {
...
-    const tr = new Transaction('test', PAGE_LOAD)
-    tr.captureTimings = true
-    captureNavigation(tr)
-    tr.end()
-
-    expect(tr.marks.agent.firstContentfulPaint).toBeGreaterThan(0)
-    unMock()
-  })
-
-  it('should handle missing first-contentful-paint on page load', function() {
```
vigneshshanmugam, line 309: missing tests are handled in perf-entry-recorder spec

PR #752, opened 2020-04-16
feat(rum-core): add network info for all transactions

### LINKED ISSUE #479: Need to support for Network Information.
would like to have a break down based on NW e.g 3G, 4G ,Wifi etc 
https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API

vigneshshanmugam:
+ fix #479 
+ All transactions like `page-load`, `route-change` and other auto instrumented transactions would include this information by default under transaction page metadata. 
+ Errors only include the previous page info and does not include the network info as its not super useful.

### DIFF
--- a/packages/rum-core/src/common/utils.js
+++ b/packages/rum-core/src/common/utils.js
@@ -200,11 +200,29 @@ function getTimeOrigin() {
   return PERF.timing.fetchStart
 }
 
-function getPageMetadata() {
+function getNetworkInformation(captureNetInfo) {
+  const connection = window.navigator && window.navigator.connection
+  if (!captureNetInfo || typeof connection !== 'object') {
+    return undefined
+  }
+  /**
+   * Ignoring `type` and `downlinkMax` as they are only
+   * supported on Chrome OS
+   */
+  return {
+    downlink: connection.downlink,
+    effective_type: connection.effectiveType,
+    rtt: connection.rtt,
+    save_data: !!connection.saveData
+  }
+}
+
+function getPageMetadata(captureNetInfo = true) {
   return {
     page: {
       referer: document.referrer,
-      url: window.location.href
+      url: window.location.href,
+      netinfo: getNetworkInformation(captureNetInfo)
     }
   }
 }
--- a/packages/rum-core/test/common/context.spec.js
+++ b/packages/rum-core/test/common/context.spec.js
@@ -182,12 +182,25 @@ describe('Context', () => {
         message: 'test'
       }
     }
-    addTransactionContext(transaction, configContext)
-    expect(transaction.context).toEqual({
+
+    const pageContext = {
       page: {
         referer: jasmine.any(String),
-        url: jasmine.any(String)
-      },
+        url: jasmine.any(String),
+        netinfo:
+          {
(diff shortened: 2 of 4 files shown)


packages/rum-core/src/common/utils.js
```diff
@@ -200,13 +200,34 @@ function getTimeOrigin() {
   return PERF.timing.fetchStart
 }
 
-function getPageMetadata() {
+function getNetworkInformation() {
+  const connection = navigator && navigator.connection
+
+  if (!(connection && typeof connection === 'object')) {
```
jahtalab, line 206: It's better to not negate the condition and just wrap the following return inside if.

packages/rum-core/src/common/utils.js
```diff
@@ -200,13 +200,34 @@ function getTimeOrigin() {
...
   return {
-    page: {
-      referer: document.referrer,
-      url: window.location.href
-    }
+    downlink: connection.downlink,
+    effective_type: connection.effectiveType,
+    rtt: connection.rtt,
+    save_data: !!connection.saveData
+  }
```
jahtalab, line 218: Do you plan to define these fields in apm server schema? if so maybe reference the PR.

Is there any reason not to include `connection.type`, it includes things like `cellular`, `wifi`, etc which is useful in categorising the network.

packages/rum-core/test/common/context.spec.js
```diff
@@ -211,7 +214,12 @@ describe('Context', () => {
       },
       ...userContext
     })
-
+    expect(pageloadTr.context.page.netinfo).toEqualOrUndefined({
```
jahtalab, line 217: IMO, using `toEqualOrUndefined` is not a good idea, instead we should detect if the feature that is required for this to work exist and we should add explicit checks for the outcomes. With this matcher we don't know whether the result was undefined because of a bug or because the api is not supported.

packages/rum-core/src/common/utils.js
vigneshshanmugam: Added the link to server schema. `type` is available only on chromeOS and its an exception.

packages/rum-core/test/common/context.spec.js
vigneshshanmugam: If i get it right, do you want to test it only for supported browsers? Like we do explicit check like 

```js
if (connection) {
  expect(measuredInfo).toBe(actualInfo)? 
}
```

I was trying to express this using the matcher, May be i missed something. We can also do explicit checks in tests if its confusing. 

````

jahtalab: correct! also check the else condition, i.e:
if (connection) {
  expect(measuredInfo).toBe(actualInfo)
} else {
 expect(measuredInfo).toBe(undefined)
}

vigneshshanmugam: Yeah sounds good 👍

vigneshshanmugam:
On hold till the spec is finalised on the APM Server- More details https://github.com/elastic/apm-server/pull/3918#issuecomment-651502077

Issue #755, opened 2020-04-19
suggestion: Lightweight errors-only agent


danieljuhl:
Have you thought about providing an additional entry point if all that is desired is to capture errors (globally and on-demand)?

We have several projects where timing is completely disabled, but I think the bundle size impact is mainly caused by the transaction/span/timing logic, and less by the error capturing.

We also want to split global error capturing from any other logic, as we need it at “root” level, to ensure all errors are captured, and not only when the website/apps full bundle has loaded. As the latter can miss important and critical errors, when something in the bundle fails completely.

jahtalab:
Thanks for your suggestion @danieljuhl !

We already have separated the code for error logging and performance monitoring which should facilitate this request. Although at this point we don't have dedicated entry point for each module.

PR #764, opened 2020-04-23
perf: refactor ServiceFactory to use constant service names

### LINKED ISSUE #238: (perf): reduce the bundle size by trimming the service factory and reduce the calls to get service.
Following the investigation that is done here https://github.com/elastic/apm-agent-rum-js/issues/159, we have the opportunity to improve the bundles size by reducing what service factory does and also reducing the number of calls to get service.

jahtalab:
Fixes #238

The gzipped bundle size is reduced from 16.78 KB on master to 16.65 KB.

### DIFF
--- a/packages/rum-angular/test/index.ts
+++ b/packages/rum-angular/test/index.ts
@@ -36,7 +36,7 @@ export function initializeApmService(service, config) {
   const { serverUrl } = globalConfig.agentConfig
   config.serverUrl = serverUrl
   const serverMock = new ApmServerMock(apmServer, globalConfig.useMocks)
-  apmBase.serviceFactory.registerServiceInstance('ApmServer', serverMock)
+  apmBase.serviceFactory.instances['ApmServer'] = serverMock
 
   return service.init(config)
 }
--- a/packages/rum-core/src/common/service-factory.js
+++ b/packages/rum-core/src/common/service-factory.js
@@ -32,22 +32,23 @@ import {
   LOGGING_SERVICE,
   APM_SERVER
 } from './constants'
+import { __DEV__ } from '../env'
+
+const serviceCreators = {
+  [CONFIG_SERVICE]: () => new ConfigService(),
+  [LOGGING_SERVICE]: () =>
+    new LoggingService({
+      prefix: '[Elastic APM] '
+    }),
+  [APM_SERVER]: factory => {
+    return new ApmServer(
+      ...factory.getService([CONFIG_SERVICE, LOGGING_SERVICE])
+    )
+  }
+}
 
 class ServiceFactory {
   constructor() {
-    this._creators = {
-      [CONFIG_SERVICE]: () => new ConfigService(),
-      [LOGGING_SERVICE]: () =>
-        new LoggingService({
-          prefix: '[Elastic APM] '
-        }),
-      [APM_SERVER]: () => {
-        return new ApmServer(
-          this.getService(CONFIG_SERVICE),
-          this.getService(LOGGING_SERVICE)
-        )
-      }
-    }
     this._instances = {}
     this.initialized = false
   }
@@ -75,24 +76,25 @@ class ServiceFactory {
     apmServer.init()
   }
 
-  registerServiceCreator(name, creator) {
-    this._creators[name] = creator
-  }
-
   registerServiceInstance(name, instance) {
     this._instances[name] = instance
   }
 
   getService(name) {
-    if (!this._instances[name]) {
-      if (typeof this._creators[name] === 'function') {
-        this._instances[name] = this._creators[name](this)
-      } else {
-        throw new Error('Can not get service, No creator for: ' + name)
+    if (typeof name === 'string') {
+      if (!this._instances[name]) {
+        if (typeof serviceCreators[name] === 'function') {
+          this._instances[name] = serviceCreators[name](this)
+        } else if (__DEV__) {
+          console.log('Can not get service, No creator for: ' + name)
+        }
       }
+      return this._instances[name]
+    } else if (name.length) {
+      // Assume name is array of strings
+      return name.map(n => this.getService(n))
     }
-    return this._instances[name]
   }
 }
 
-export default ServiceFactory
+export { serviceCreators, ServiceFactory }
--- a/packages/rum-core/src/common/service-factory.js
+++ b/packages/rum-core/src/common/service-factory.js
@@ -41,15 +41,17 @@ const serviceCreators = {
       prefix: '[Elastic APM] '
     }),
   [APM_SERVER]: factory => {
-    return new ApmServer(
-      ...factory.getService([CONFIG_SERVICE, LOGGING_SERVICE])
-    )
+    const [configService, loggingService] = factory.getService([
+      CONFIG_SERVICE,
+      LOGGING_SERVICE
+    ])
+    return new ApmServer(configService, loggingService)
   }
 }
 
 class ServiceFactory {
   constructor() {
-    this._instances = {}
+    this.instances = {}
     this.initialized = false
   }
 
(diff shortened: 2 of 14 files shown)


packages/rum-core/src/common/service-factory.js
```diff
@@ -73,27 +71,27 @@ class ServiceFactory {
...
 
   getService(name) {
-    if (!this._serviceInstances[name]) {
-      if (typeof this._serviceCreators[name] === 'function') {
-        this._serviceInstances[name] = this._serviceCreators[name](this)
+    if (!this._instances[name]) {
+      if (typeof this._creators[name] === 'function') {
+        this._instances[name] = this._creators[name](this)
       } else {
         throw new Error('Can not get service, No creator for: ' + name)
```
vigneshshanmugam, line 91: It good to not throw here, instead log as error in dev mode.

jahtalab:
@vigneshshanmugam , 

Re No. 1, Separating the modules like ErrorLogging and PerformanceMonitoring helps with the ability to accommodate requests like https://github.com/elastic/apm-agent-rum-js/issues/755. Having said that, we can still reduce the overhead by using service factory objects directly or other similar ideas. let's discuss it further on the sync.

Re. 2, I think the trade-off is between executing JS during load time or during init time, the lazying init with it's current form optimises for running most/all of the work in the init time. But we discuss further different ways to resolve the trade-off.

jahtalab:
We had a discussion:
- Refactor `registerServiceCreator`
- eliminate apm-server init

packages/rum-core/src/common/service-factory.js
```diff
@@ -32,22 +32,23 @@ import {
...
+  [APM_SERVER]: factory => {
+    return new ApmServer(
+      ...factory.getService([CONFIG_SERVICE, LOGGING_SERVICE])
+    )
+  }
+}
 
 class ServiceFactory {
   constructor() {
-    this._creators = {
```
vigneshshanmugam: Having the creators here is actually better and we are modifying the creators in another files seems like a sideeffect. Lets use the same patter and instead reduce the naming of instance methods to something like `addCreator` and `addInstance` instead of big `registerServiceInstance` etc.

packages/rum-core/src/error-logging/index.js
```diff
@@ -25,14 +25,17 @@
...
+import { serviceCreators } from '../common/service-factory'
+
+function registerServices() {
+  serviceCreators['ErrorLogging'] = serviceFactory => {
+    const deps = serviceFactory.getService([
+      APM_SERVER,
+      CONFIG_SERVICE,
+      'TransactionService'
+    ])
+    return new ErrorLogging(...deps)
```
vigneshshanmugam, line 37: Just check how spread `...deps` is transpiled - might be a reason why you are seeing a increase in bundle size.

packages/rum-core/src/common/service-factory.js
```diff
@@ -75,24 +76,25 @@ class ServiceFactory {
...
+    if (typeof name === 'string') {
+      if (!this._instances[name]) {
+        if (typeof serviceCreators[name] === 'function') {
+          this._instances[name] = serviceCreators[name](this)
+        } else if (__DEV__) {
+          console.log('Can not get service, No creator for: ' + name)
+        }
       }
+      return this._instances[name]
+    } else if (name.length) {
```
vigneshshanmugam, line 93: Lets use Array.isArray.

packages/rum-core/src/common/service-factory.js
jahtalab: I compared the size reduction for having creators on the class, it nearly had 0.06 KB reduction in size, so I kept it this way. 

You're right about the side-effect but I think we probably should accept some side-effect in favour of the bundle size reduction.

packages/rum-angular/test/index.ts
```diff
@@ -36,7 +36,7 @@ export function initializeApmService(service, config) {
   const { serverUrl } = globalConfig.agentConfig
   config.serverUrl = serverUrl
   const serverMock = new ApmServerMock(apmServer, globalConfig.useMocks)
-  apmBase.serviceFactory.registerServiceInstance('ApmServer', serverMock)
+  apmBase.serviceFactory.instances['ApmServer'] = serverMock
```
vigneshshanmugam, line 39: can you import and use service constants?

vigneshshanmugam, line 39: can you import and use service constants?

packages/rum-core/src/common/service-factory.js
```diff
@@ -73,28 +74,24 @@ class ServiceFactory {
...
-      if (typeof this._serviceCreators[name] === 'function') {
-        this._serviceInstances[name] = this._serviceCreators[name](this)
-      } else {
-        throw new Error('Can not get service, No creator for: ' + name)
+    if (typeof name === 'string') {
+      if (!this.instances[name]) {
+        if (typeof serviceCreators[name] === 'function') {
+          this.instances[name] = serviceCreators[name](this)
+        } else if (__DEV__) {
+          console.log('Can not get service, No creator for: ' + name)
```
vigneshshanmugam, line 87: nit: 'Cannot'

packages/rum-core/src/error-logging/index.js
```diff
@@ -24,15 +24,22 @@
...
+    ] = serviceFactory.getService([
+      APM_SERVER,
+      CONFIG_SERVICE,
+      'TransactionService'
+    ])
+    return new ErrorLogging(apmServer, configService, transactionService)
   }
 }
+
+export { ErrorLogging, registerServices }
```
vigneshshanmugam, line 45: Can we remove the `ErrorLogging` export as it's not used?

packages/rum-core/src/index.js
```diff
@@ -25,25 +25,30 @@
...
+  ERROR,
+  CONFIG_SERVICE,
+  LOGGING_SERVICE,
+  APM_SERVER
+} from './common/constants'
 import { getInstrumentationFlags } from './common/instrument'
 import afterFrame from './common/after-frame'
 import { createTracer } from './opentracing'
 
+registerPerfServices()
```
vigneshshanmugam, line 48: this feels like Side Effect. can we move this inside createServiceFactory itself? Atleast we know we register only when someone creates the factory.

packages/rum-core/src/performance-monitoring/index.js
```diff
@@ -25,27 +25,42 @@
...
-      var transactionService = serviceFactory.getService('TransactionService')
-      return new PerformanceMonitoring(
-        apmService,
-        configService,
-        loggingService,
-        transactionService
-      )
-    })
+function registerServices() {
+  serviceCreators['TransactionService'] = serviceFactory => {
```
vigneshshanmugam, line 37: Any reason why we didn't have TransactionService in constant?

packages/rum-core/src/performance-monitoring/index.js
```diff
@@ -25,27 +25,42 @@
...
+    return new PerformanceMonitoring(
+      apmServer,
+      configService,
+      loggingService,
+      transactionService
+    )
   }
 }
+
+export { PerformanceMonitoring, registerServices }
```
vigneshshanmugam, line 66: remove PerformanceMonitoring export as well?

packages/rum-core/src/common/service-factory.js
```diff
@@ -26,42 +26,43 @@
...
-import { CONFIG_CHANGE } from './constants'
+import {
+  CONFIG_CHANGE,
+  CONFIG_SERVICE,
+  LOGGING_SERVICE,
+  APM_SERVER
+} from './constants'
+import { __DEV__ } from '../env'
+
+const serviceCreators = {
```
vigneshshanmugam, line 37: This is exactly similar to file level instance. Remember you were suggesting aganist this idea? This would mean we have the same factory instance for all the tests? Previously `createServiceFactory` would give us a fresh instance every time.

jahtalab: Although, this is a file level object, it's not the same as I was suggesting.

The file level instance that I don't like is the one that contains our service instances, imagine if `ServiceFactory.instances` was a file level object, this is not a good pattern.

But serviceCreators are only functions that create the services, the only way this file level object can make problems is as if we start changing it during our tests.

jahtalab:
@vigneshshanmugam , I've rebased and addressed some of the comments. 

Re. using constant service names, actually don't have much effect on the gzipped files, so I don't think it makes sense to use them everywhere, but if there're multiple of the same service in a file that could be a good place to use constants.

Issue #656, opened 2020-02-26
Provide a bot to keep our dev dependencies updated


jahtalab:
Similar to the greenkeeper service or the dependent bot. One requirement is to recognise local dependencies, example pr:https://github.com/elastic/apm-agent-rum-js/pull/643

@elastic/observablt-robots , do you have any solution for this?

v1v:
To avoid any issues with the CLA, we will build a wrapper to create PRs on behalf of the service account we use in the CI.

An example of the PRs creation can be found here:
- https://github.com/v1v/apm-agent-rum-js/pulls?q=is%3Apr+is%3Aopen+label%3Adependencies

This is based on an draft jenkins pipeline library I just coded that uses:
- https://github.com/dependabot/dependabot-script 

![image.png](https://zube.io/files/elastic/cce3b9f2932346dc5ba732ef8a13b9c0-image.png)

In a nutshell, we will build a process in place that will help to the teams to control:
- When to run the dev dependencies update: (daily, weekly, monthly)
- Who to assign those issues.
- What package manager to apply. (npm_yarn, maven, go, gradle)

For such we will create a properties file under the .ci folder then the automation. will read them and apply the execution.

v1v:
What do you think?

v1v:
Regarding the approach to use dependabot, @watson just added quite valuable information for some other use cases:

- https://github.com/elastic/apm-agent-nodejs/pull/1727#issuecomment-618569553

Glad to say even if we use the dependabot adhoc thing, we could add more features to fullfil some of those requirements.

I'll gather all those details in another issue to discuss deeply with all the apm teams and agree what's the best approach.

vigneshshanmugam:
IMO, Using dependabot is a good idea and would already cut our work when updating dependencies. However as i expressed already in a thread, Dependabot doesn't seem to be working correctly on the Lerna mono-repos and we might need to tweak it a little bit to behave correctly. Other than that, I agree with your points mentioned here https://github.com/elastic/apm-agent-rum-js/issues/656#issuecomment-618408210

cachedout:
@v1v One question I think is worth asking might be if we can get a high-level view of dependencies across **all** the observability repos. Perhap Do you think that has value?

mdelapenya:
> @v1v One question I think is worth asking might be if we can get a high-level view of dependencies across **all** the observability repos. Perhap Do you think that has value?

[...]

PR #765, opened 2020-04-23
ci(dependabot): configuration file

v1v:
If we would like to enable the dependabot automation process to bump the version of the packages automatically, then it's required this particular file.

### How does it work?

There is a weekly pipeline that will care for all the projects with this particular file `ci/dependabot.yml` and if so it will run the validation and PR creations.

### For instance

https://github.com/v1v/apm-agent-rum-js/pulls?q=is%3Apr+is%3Alabel%3Adependencies

### Actions

- [ ] Verify if this particular dependabot is a good candidate to be enabled in this project.
- [ ] Assign?

### Issues

Closes https://github.com/elastic/apm-agent-rum-js/issues/656

### DIFF
--- a/.ci/dependabot.yml
+++ b/.ci/dependabot.yml
@@ -1,0 +1,5 @@
+### The list of GitHub accounts that will be used to assign the PRs to (no space and comma separated)
+assign: ''
+
+### The package manager to be used (no space and comma separated if more than one package)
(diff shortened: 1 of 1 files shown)


v1v:
I'll close this PR now for the time being, as we are still gathering details about what's the tool to be used.

PR #771, opened 2020-04-28
perf(rum-core): compress server payload for all events

### LINKED ISSUE #768: Payload Optimisation phase I
+ Part of #158 
+ APM Server has deployed the first phase of the Proposed payload optimisation technique which is Shortening the field names that are sent in the payload PR - https://github.com/elastic/apm-server/pull/3414
+ Removing duplicate fields - some fields inside spans like `transaction_id`,`trace_id`, `parent_id` inside spans
+ One exception with `parent_id` is instead of passing the whole id incase when spans are subspans of other spans inside same transaction. we send the index of the parent span inside `parent_idx` mapped to `pi` of the payload to avoid sending the full id. 
+ `start`, `duration` are represented as JSON numbers so we can send integers instead of floats. 
+ `sync` is optional. so we send only when it's true. 

We are still exploring what could be done with metricsets. Moving them inside transaction or leaving them in separate lines (marking transaction details as optional). 

Full mapping of all the fields is here - https://github.com/elastic/apm-server/blob/master/model/modeldecoder/field/rum_v3_mapping.go

vigneshshanmugam:
+ fix #768 
+ Adds a new config option `compressPayload` which is set to false by default and when enabled will use the new v3 spec for reporting the events. 
+ Compression is done after the payload to make sure we don't break the existing version and also let the user supplied filters work as expected
 + floats are converted to integers and this would be applied for all the previous apm server versions as well and would be applied even if the flag is set to false. Server schema is a JSON number so this validation is intact. 

### Test Scenario

No of Spans 
+ 50 JS files
+ 3 Custom spans (Parsing document, DOM interactive etc)
+ Metricsets/Breakdown - 4 for page-load

| Name of the test               | Payload Size (approx bytes) |
| :------------------------ | :-------------------------: |
| Current Release                |            31750            |
| After this PR                      |            17057            |

**Note: The flag would be removed and set to true by default from verison 6.x**

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -28,15 +28,23 @@ import throttle from './throttle'
 import NDJSON from './ndjson'
 import { XHR_IGNORE } from './patching/patch-utils'
 import { truncateModel, METADATA_MODEL } from './truncate'
-import { SERVER_URL_PREFIX, ERRORS, TRANSACTIONS } from './constants'
+import { ERRORS, TRANSACTIONS } from './constants'
 import { noop } from './utils'
 import { Promise } from './polyfills'
+import {
+  compressMetadata,
+  compressTransaction,
+  compressError
+} from './compress'
 import { __DEV__ } from '../env'
 
 /**
  * Throttling interval deafults to 60 seconds
  */
 const THROTTLE_INTERVAL = 60000
+
+const V2_PREFIX = '/intake/v2/rum/events'
+const V3_PREFIX = '/intake/v3/rum/events'
 
 class ApmServer {
   constructor(configService, loggingService) {
@@ -210,28 +218,42 @@ class ApmServer {
     this.throttleEvents({ [TRANSACTIONS]: transaction })
   }
 
-  ndjsonErrors(errors) {
-    return errors.map(error => NDJSON.stringify({ error }))
+  ndjsonErrors(errors, compress) {
+    const key = compress ? 'e' : 'error'
+    return errors.map(error =>
+      NDJSON.stringify({
+        [key]: compress ? compressError(error) : error
+      })
+    )
   }
 
   ndjsonMetricsets(metricsets) {
     return metricsets.map(metricset => NDJSON.stringify({ metricset })).join('')
   }
 
-  ndjsonTransactions(transactions) {
+  ndjsonTransactions(transactions, compress) {
+    const key = compress ? 'x' : 'transaction'
+
     return transactions.map(tr => {
-      let spans = ''
-      if (tr.spans) {
-        spans = tr.spans.map(span => NDJSON.stringify({ span })).join('')
-        delete tr.spans
-      }
-      let breakdowns = ''
-      if (tr.breakdown) {
-        breakdowns = this.ndjsonMetricsets(tr.breakdown)
-        delete tr.breakdown
+      let spans = '',
+        breakdowns = ''
+
+      if (!compress) {
+        if (tr.spans) {
+          spans = tr.spans.map(span => NDJSON.stringify({ span })).join('')
+          delete tr.spans
+        }
+        if (tr.breakdown) {
+          breakdowns = this.ndjsonMetricsets(tr.breakdown)
+          delete tr.breakdown
+        }
       }
 
-      return NDJSON.stringify({ transaction: tr }) + spans + breakdowns
+      return (
+        NDJSON.stringify({ [key]: compress ? compressTransaction(tr) : tr }) +
+        spans +
+        breakdowns
+      )
     })
   }
 
@@ -253,26 +275,34 @@ class ApmServer {
       return
     }
 
+    const cfg = this._configService
     const payload = {
       [TRANSACTIONS]: transactions,
       [ERRORS]: errors
     }
-    const filteredPayload = this._configService.applyFilters(payload)
+    const filteredPayload = cfg.applyFilters(payload)
     if (!filteredPayload) {
       this._loggingService.warn('Dropped payload due to filtering!')
       return
     }
 
+    const compress = cfg.get('compressPayload')
     let ndjson = []
     const metadata = this.createMetaData()
-    ndjson.push(NDJSON.stringify({ metadata }))
+    const metadataKey = compress ? 'm' : 'metadata'
+
+    ndjson.push(
+      NDJSON.stringify({
+        [metadataKey]: compress ? compressMetadata(metadata) : metadata
+      })
+    )
 
     ndjson = ndjson.concat(
-      this.ndjsonErrors(filteredPayload[ERRORS]),
-      this.ndjsonTransactions(filteredPayload[TRANSACTIONS])
+      this.ndjsonErrors(filteredPayload[ERRORS], compress),
+      this.ndjsonTransactions(filteredPayload[TRANSACTIONS], compress)
     )
     const ndjsonPayload = ndjson.join('')
-    const endPoint = this._configService.get('serverUrl') + SERVER_URL_PREFIX
+    const endPoint = cfg.get('serverUrl') + compress ? V3_PREFIX : V2_PREFIX
     return this._postJson(endPoint, ndjsonPayload)
   }
 }
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -287,9 +287,9 @@ class ApmServer {
 
     const apiVersion = cfg.get('apiVersion')
     /**
-     * when API version is >2, we can enable compression of the payload
+     * Enable payload compression only when API version is > 2
      */
-    const compress = apiVersion > 2 ? true : false
+    const compress = apiVersion > 2
 
     let ndjson = []
     const metadata = this.createMetaData()
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -302,7 +302,7 @@ class ApmServer {
       this.ndjsonTransactions(filteredPayload[TRANSACTIONS], compress)
     )
     const ndjsonPayload = ndjson.join('')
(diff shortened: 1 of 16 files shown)


packages/rum-core/src/common/apm-server.js
```diff
@@ -28,16 +28,24 @@ import throttle from './throttle'
...
 import { Promise } from './polyfills'
+import {
+  compressMetadata,
+  compressTransaction,
+  compressError
+} from './compress'
 import { __DEV__ } from '../env'
 
 /**
  * Throttling interval deafults to 60 seconds
```
jahtalab, line 42: nit: defaults

packages/rum-core/src/common/config-service.js
```diff
@@ -89,6 +89,7 @@ class Config {
       transactionSampleRate: 1.0,
       centralConfig: false,
       monitorLongtasks: true,
+      compressPayload: false,
```
jahtalab, line 92: Thinking about this again, having something like `apiVersion: 2` or similar would be better since we can include all the changes related to v3 before we release the next major.

vigneshshanmugam: Hmmm, I was thinking in the similar area. apiVersion also sounds good to me. I was thinking in terms of stack version so we can enable breakdown or other configs by default ?

vigneshshanmugam: We just need to agree on the config name and then we can merge this PR.

jahtalab: The stack version is also not a bad idea, although I'm leaning towards api version since we expect it to change less frequently and hopefully we can remove it with version 6.0

vigneshshanmugam: Sounds good, let's go with apiVersion. I will change it now.

packages/rum-core/src/common/apm-server.js
```diff
@@ -250,7 +269,9 @@ class ApmServer {
     }
     const transactions = []
     const errors = []
-    for (const event of events) {
+    for (let i = 0; i < events.length; i++) {
```
vigneshshanmugam, line 272: changed it to reduce unnecessary iterator polyfill when bundling.

packages/rum-core/src/common/apm-server.js
```diff
@@ -262,26 +283,39 @@ class ApmServer {
...
     if (!filteredPayload) {
       this._loggingService.warn('Dropped payload due to filtering!')
       return
     }
 
+    const apiVersion = cfg.get('apiVersion')
+    /**
+     * when API version is >2, we can enable compression of the payload
+     */
+    const compress = apiVersion > 2 ? true : false
```
jahtalab, line 301: nit: can be change to `compress = apiVersion>2`

packages/rum-core/src/common/compress.js
```diff
@@ -0,0 +1,294 @@
...
+        compressed.sc = value
+        break
+      case 'url':
+        compressed.url = value
+        break
+      case 'response':
+        compressed.r = compressResponse(value)
+        break
+    }
+  }
```
jahtalab, line 88: Why not just hard-code the keys instead of using a for loop?

packages/rum-core/src/common/compress.js
```diff
@@ -0,0 +1,294 @@
...
+            rc: service.resource
+          },
+          ad: value.address,
+          po: value.port
+        }
+        break
+      case 'custom':
+        compressed.cu = value
+        break
+    }
```
jahtalab, line 134: here also, the for loop seems unnecessary!

packages/rum-core/src/common/compress.js
```diff
@@ -0,0 +1,294 @@
...
+
+function compressMarks(marks) {
+  if (!marks) {
+    return null
+  }
+  const { navigationTiming, agent } = marks
+  const compressed = { nt: {}, a: {} }
+
+  COMPRESSED_NAV_TIMING_MARKS.forEach((mark, index) => {
+    const mapping = NAVIGATION_TIMING_MARKS[index]
```
jahtalab, line 147: Would you please move these two constants in the same file and add a comment that changing the order of one of them would affect the compression?

But a better solution would be to change `NAVIGATION_TIMING_MARKS` to be an object that contains the mapping and remove `COMPRESSED_NAV_TIMING_MARKS`

packages/rum-core/src/common/compress.js
```diff
@@ -0,0 +1,294 @@
...
+  }
+  return compressed
+}
+
+function compressMarks(marks) {
+  if (!marks) {
+    return null
+  }
+  const { navigationTiming, agent } = marks
+  const compressed = { nt: {}, a: {} }
```
jahtalab, line 144: You don't need the `a: {}` since you're setting it a bit after

packages/rum-core/src/common/compress.js
```diff
@@ -0,0 +1,294 @@
...
+     * Set the below fields only when they are truthy or exists
+     */
+    if (span.sync === true) {
+      spanData.sy = true
+    }
+    if (span.subtype) {
+      spanData.su = span.subtype
+    }
+    if (span.action) {
+      spanData.ac = span.action
```
jahtalab, line 215: Did we used to set empty string on these fields?

packages/rum-core/test/common/compess.spec.js
```diff
@@ -0,0 +1,248 @@
...
+    for (const key of Object.keys(originalFlat)) {
+      const keys = key.split(SEPARATOR)
+      const mappedKeys = getMappedKeys(keys)
+      const compressedKey = mappedKeys.join(SEPARATOR)
+      /**
+       * Some fields are made optional in the v3 spec so we check
+       * the condition only when the values are present in new scheme
+       */
+      if (compressedFlat[compressedKey] != null) {
+        expect(originalFlat[key]).toEqual(compressedFlat[compressedKey])
```
jahtalab, line 209: I'm not a fan of allowing for null values in general. This would mean that if by mistake we don't set a field it won't be captured.

packages/rum-core/src/common/compress.js
vigneshshanmugam: No its undefined by default.

packages/rum-core/test/common/compess.spec.js
vigneshshanmugam: Some values are optional in new spec. Its okay to not set a field as these tests are only used to check the compressed value equal to actual value. Not for any other purposes.

packages/rum-core/src/common/compress.js
vigneshshanmugam: Mapping of these two are not done as we cannot determine whether we have to use key or value depending on whether compression is supported or not at this part of the code. I will move them to same place and add comments.

packages/rum-core/src/common/compress.js
vigneshshanmugam: Yeah two patterns, either this or plan check for undefined at every key and add that. I went with this loop one. I don't have a strong preference though.

packages/rum-core/src/common/compress.js
```diff
@@ -0,0 +1,294 @@
...
+
+  return {
+    id: transaction.id,
+    tid: transaction.trace_id,
+    n: transaction.name,
+    t: transaction.type,
+    d: transaction.duration,
+    c: compressContext(transaction.context),
+    m: compressMarks(transaction.marks),
+    b: compressMetricsets(transaction.breakdown),
```
vigneshshanmugam, line 228: self note: change it to `me`

packages/rum-core/test/common/compess.spec.js
jahtalab: My concern is that if a field is not set by the compress algorithm by mistake it won't be caught in our tests. Isn't that correct?

The alternative to this is to hardcode those optional fields and expect a value for everything else.

vigneshshanmugam: We can check the length of compressed vs uncompressed keys to make sure we did not miss anything. I am not in favor of hardcoding as we have to keep track of those fields manually.

jahtalab: From what I understand from the algorithm checking the number of keys won't work since there're some fields that can be omitted based on their value. 

IMO, we need to update the tests every time these keys change anyway. 
Also, I could be wrong, but isn't there only a few fields that are optional, everything else is has to be mapped to some field. we can only hardcode those few optional fields

vigneshshanmugam: Each unmapped key like `metadata.service.name` would be mapped to `m.s.n` and the keys length for uncompressed and compressed would be the same since we have mapping defined for all of those. If its not, then it means we have a bug. 

The optional fields are the one we introduced now like 
```
transaction_id
parent_id
trace_id
sync
subtype
action
transaction.name and type in breakdown
```

These fields would be available in previous model but not on the compressed ones. 

We cannot hardcode and put these checks manually as the single algorithm looks for all of the events and I like to keep it this way. We already have a test which would fail if we don't send any fields that APM Server expects.

packages/rum-core/test/common/compress.spec.js
```diff
@@ -0,0 +1,272 @@
...
+    for (const key of Object.keys(originalFlat)) {
+      const keys = key.split(SEPARATOR)
+      const mappedKeys = getMappedKeys(keys)
+      const compressedKey = mappedKeys.join(SEPARATOR)
+      /**
+       * Some fields are made optional in the v3 spec so we check
+       * the condition only when the keys are not optional and
+       * values exist in the new compressed model
+       */
+      if (!isOptional(key) && compressedFlat[compressedKey] != null) {
```
jahtalab, line 232: I think this should be `!isOptional(key) || compressedFlat[compressedKey] != null`

vigneshshanmugam: damm. my bad 😃

vigneshshanmugam:
@jahtalab The Jenkins node was returning 500, I restarted it now.

PR #774, opened 2020-04-30
chore(dev): watch files when running karma dev

vigneshshanmugam:
+ Without this flag set to false, watch mode does not work correctly.

### DIFF
--- a/dev-utils/karma.js
+++ b/dev-utils/karma.js
@@ -68,7 +68,7 @@ const baseConfig = {
     'karma-sourcemap-loader'
   ],
   client: {
-    // clearContext: false, // make Jasmine Spec Runner output visible in browsers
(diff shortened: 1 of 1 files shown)


vigneshshanmugam:
@jahtalab Please test it, It seems to be the one that is affecting and I have spent some time yesterday figuring out but this the the closest I could get away with. 

Plus I added `clearContext: false` last time to test the output with html reporter and we dont even use it.

vigneshshanmugam:
To be honest, I tried debugging for sometime but couldn't find the the real reason or whether this was a recent change in karma itself. so just added a PR so we can rurn `karma:dev` again happily 😆

vigneshshanmugam:
Merging this PR as the Integration test issue is unrelated to the changes.

PR #787, opened 2020-05-13
feat: add first input delay span

jahtalab:
This PR includes the following:
- Adds First input delay metrics
- Refactor perf-entry-recorder
- Move bootstrap from `apm-rum` to `apm-rum-core`


packages/rum-core/test/performance-monitoring/metrics-polyfill.spec.js
```diff
@@ -0,0 +1,49 @@
...
+import {
+  polyfillFID,
+  metrics
+} from '../../src/performance-monitoring/metrics-polyfill'
+describe('metrics-polyfill', () => {
+  it('should polyfill first input delay', done => {
+    polyfillFID()
+    expect(metrics).toEqual({})
+
+    document.body.click()
```
vigneshshanmugam, line 35: Need more scenarios to account for a single final value instead of multiples.

jahtalab:
Meeting notes:
- Need to double check the visibility logic
- Investigate removing `require.cache`

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -93,6 +95,103 @@ function createLongTaskSpans(longtasks) {
...
+
+  eachEventType(addEventListener)
+}
+
+export function createFirstInputDelaySpan(list) {
+  const fidEntries = list.getEntriesByType(FIRST_INPUT)
+  let firstInput = fidEntries[0]
+
+  if (firstInput && !metrics.wasHidden) {
+    let { name, startTime, processingStart } = firstInput
```
vigneshshanmugam, line 184: We dont want to use the `target`? this would be useful in checking which element was responsible for the FID. 
https://wicg.github.io/event-timing/#dom-performanceeventtiming-target

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -215,6 +314,11 @@ export function captureObserverEntries(list, { capturePaint }) {
     result.spans.push(tbtSpan)
   }
 
+  const fidSpan = createFirstInputDelaySpan(list)
```
vigneshshanmugam, line 317: Lets follow the same semantics as other functions in this file. Please pass the entries here instead of the list. 

```
const fidEntries = list.getEntriesByType(FIRST_INPUT)
const fidSpan = createFirstInputDelaySpan(fidEntries)
```

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -26,12 +26,14 @@
...
   LARGEST_CONTENTFUL_PAINT,
-  FIRST_CONTENTFUL_PAINT
+  FIRST_CONTENTFUL_PAINT,
+  FIRST_INPUT
 } from '../common/constants'
 import { noop, PERF } from '../common/utils'
-import { metrics } from '../common/metrics'
 import Span from './span'
 
+export const metrics = {}
```
vigneshshanmugam, line 35: need pre initialisation of FCP here. 
```js
export const metrics = {
   fcp: 0
}
```

We use this to check TBT https://github.com/elastic/apm-agent-rum-js/blob/cac9bfc0bbd14ad52e998ae95c537591765929b9/packages/rum-core/src/performance-monitoring/perf-entry-recorder.js#L111

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -269,3 +373,21 @@ export class PerfEntryRecorder {
...
 }
+
+export function bootstrapMetrics() {
+  metrics.wasHidden = document.visibilityState === 'hidden'
+
+  window.addEventListener(
+    'visibilitychange',
+    () => {
+      if (document.visibilityState === 'hidden') {
+        metrics.wasHidden = true
```
vigneshshanmugam, line 384: Did you check about the use of timestamp?

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -269,3 +373,21 @@ export class PerfEntryRecorder {
...
+      }
+    },
+    { capture: true, once: true }
+  )
+
+  if (!isPerfEntryTypeSupported(FIRST_INPUT)) {
+    polyfillFID()
```
vigneshshanmugam, line 391: I think this polyfill wont work for libraries and works only for users who can include it on the HTML head document.

packages/rum-core/test/performance-monitoring/metrics.spec.js
```diff
@@ -0,0 +1,252 @@
...
+  metrics
+} from '../../src/performance-monitoring/metrics'
+import { LARGEST_CONTENTFUL_PAINT, LONG_TASK } from '../../src/common/constants'
+import {
+  mockObserverEntryTypes,
+  mockObserverEntryNames
+} from '../utils/globals-mock'
+import longtaskEntries from '../fixtures/longtask-entries'
+
+polyfillFID()
```
vigneshshanmugam, line 40: Include it on `polyfills.js`

packages/rum/test/node/index-export.spec.js
```diff
@@ -41,8 +41,10 @@ describe('apm base', () => {
     /**
      * Delete module cache and run bootstrap again
      */
-    delete require.cache[require.resolve('@elastic/apm-rum')]
-    delete require.cache[require.resolve('@elastic/apm-rum/dist/lib/bootstrap')]
+    let cache = require.cache
```
vigneshshanmugam, line 44: you could just do delete require.cache or set it to empty object instead of looping over

packages/rum/test/specs/index.spec.js
```diff
@@ -36,6 +36,10 @@ describe('index', function() {
   beforeEach(function() {
     originalTimeout = jasmine.DEFAULT_TIMEOUT_INTERVAL
     jasmine.DEFAULT_TIMEOUT_INTERVAL = 120000
+    let cache = require.cache
```
vigneshshanmugam, line 39: same as above.

packages/rum-core/src/performance-monitoring/metrics.js
jahtalab: We decided to use visibiltyState instead of responsiveness in a separate PR.

packages/rum/test/node/index-export.spec.js
jahtalab: @vigneshshanmugam , setting require.cache won't work, since it seems the same object ref is kept somewhere else so we have to delete its keys.

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -95,95 +97,13 @@ function createLongTaskSpans(longtasks) {
...
   let firstInput = fidEntries[0]
 
   if (firstInput && !metrics.wasHidden) {
-    let { name, startTime, processingStart } = firstInput
+    const { name, startTime, processingStart } = firstInput
 
-    let span = new Span('FirstInputDelay', FIRST_INPUT, { startTime })
+    const span = new Span('First Input Delay', FIRST_INPUT, { startTime })
     span.addContext({
       custom: { eventType: name }
```
vigneshshanmugam, line 108: Server intake schema doesn't have camelCased fields. Can we follow the convention and use `eventtype` instead?

packages/rum-core/src/bootstrap.js
```diff
@@ -23,19 +23,22 @@
...
-export default function bootstrap() {
+export function bootstrap() {
   if (alreadyBootstrap) {
     return enabled
   }
   alreadyBootstrap = true
 
   if (isPlatformSupported()) {
     patchAll()
+    bootstrapMetrics()
```
vigneshshanmugam, line 41: nit: can you name it something like `observeHidden` or `observeVisibility`? bootstrap doesn't even come close to what we are doing in that function.

packages/rum/test/node/index-export.spec.js
vigneshshanmugam: I am removing this test in my next PR #796 so lets not spend more time. I am good with it.

packages/rum-core/src/bootstrap.js
jahtalab: I think we should start using bootstrap (or something similar) to indicate that this function has side-effects!

vigneshshanmugam: IMO, This function doesn't have side effects, we don't add anything to global, Adding listeners are not side effecty. If you refer to `metrics` its on module scope not on global.

jahtalab: Why adding an event listener is not a side-effect? IMO, it's adding a global event therefore it's a side-effect.

packages/rum-core/src/performance-monitoring/metrics.js
jahtalab: I removed it for now, let's add it when there's a use-case for it and also the proper way would be to add it to top-level context instead of custom and define it in schema as well.

packages/rum-core/src/bootstrap.js
vigneshshanmugam: Observers are not side effect. Anyway its not a hard concern for this PR.

vigneshshanmugam:
@jahtalab Side note, please update the description of the PR.

packages/rum-core/src/bootstrap.js
jahtalab: I can't say I agree with that, as an example consider the case in which bootstrap is called twice, in that case two event listeners are going to be registered which is obviously not a correct logic. the bootstrap function indicates that the operation within it should be executed once per application.

PR #792, opened 2020-05-20
fix(rum): allow setting labels before initializing

### LINKED ISSUE #780: Refactor ConfigService to be able to set configs even before calling init
We can remove the `defaults` property and use its value to initialise `config` property on construction.

vigneshshanmugam:
+ fix #780 
+ User configs would be merged with the actual config instead of having `defaults`. 
+ Added a test case and removed the `config.init` from places where unnecessary.

### DIFF
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -64,8 +64,7 @@ function getDataAttributesFromNode(node) {
 
 class Config {
   constructor() {
-    this.config = {}
-    this.defaults = {
+    this.config = {
       serviceName: '',
       serviceVersion: '',
       environment: '',
@@ -180,7 +179,7 @@ class Config {
       properties.serverUrl = properties.serverUrl.replace(/\/+$/, '')
     }
 
-    this.config = merge({}, this.defaults, this.config, properties)
+    this.config = merge({}, this.config, properties)
     this.events.send(CONFIG_CHANGE, [this.config])
   }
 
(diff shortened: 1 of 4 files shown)


vigneshshanmugam:
@jahtalab Test failure is unrelated, Ready to be reviewed

packages/rum-core/src/common/config-service.js
```diff
@@ -180,7 +179,7 @@ class Config {
       properties.serverUrl = properties.serverUrl.replace(/\/+$/, '')
     }
 
-    this.config = merge({}, this.defaults, this.config, properties)
+    this.config = merge({}, this.config, properties)
```
jahtalab, line 182: I don't think we still to pass a new object here, we can simply merge into `this.config`

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -308,6 +308,21 @@ describe('ApmBase', function() {
...
+    }
+    apmBase.addLabels(labels)
+    apmBase.init({
+      serviceName,
+      serverUrl,
+      disableInstrumentations: [PAGE_LOAD]
+    })
+    const configService = serviceFactory.getService('ConfigService')
+    expect(configService.get('context.tags')).toEqual(labels)
+  })
```
jahtalab, line 324: I think having this test here is fine but since this change is in config service, the test should be in it's spec file. there can be two test if you think that's justified.

vigneshshanmugam: There is already a test in config service, so I added this test to ensure setting labels before init.

PR #796, opened 2020-05-25
fix(rum): use single instance of apm across all packages

### LINKED ISSUE #791: bundle should not replace `elasticApm` if its present on window
Currently our `ApmBase` bundle if loaded multiple times will create a new instance of `ApmBase` and replace the existing one, this is a bug and creates problems if the users are instrumenting the webpage using our UMD bundle and integration in other pages. As a result of replacing the existing one, we have to initialise again otherwise the integration would be disabled.

+ This is one of the common cases in Kibana as we can use `@elastic/apm-rum` bundle for the initial load and use Integrations like `ApmRoute` for react specific plugins.

vigneshshanmugam:
+ fixes #791 
+ Ensures the following when the agent script is loaded twice or agent script is used along with other framework integrations
  -  `apm` instance is singleton 
  - does not patch the `window.elasticApm` multiple times
  - does not patch XHR, Fetch, History multiple times.

### DIFF
--- a/packages/rum-core/src/common/bootstrap.js
+++ b/packages/rum-core/src/common/bootstrap.js
@@ -23,24 +23,17 @@
  *
  */
 
-import { isPlatformSupported } from './common/utils'
-import { patchAll } from './common/patching'
-import { bootstrapMetrics } from './performance-monitoring/metrics'
+import { isPlatformSupported, isBrowser } from './utils'
+import { patchAll } from './patching'
+import { bootstrapMetrics } from '../performance-monitoring/metrics'
 
-let alreadyBootstrap = false
 let enabled = false
-
 export function bootstrap() {
-  if (alreadyBootstrap) {
-    return enabled
-  }
-  alreadyBootstrap = true
-
   if (isPlatformSupported()) {
     patchAll()
     bootstrapMetrics()
     enabled = true
-  } else if (typeof window !== 'undefined') {
+  } else if (isBrowser) {
     /**
      * Print this error message only on the browser console
      * on unsupported browser versions
--- a/packages/rum/test/specs/index.spec.js
+++ b/packages/rum/test/specs/index.spec.js
@@ -46,22 +46,6 @@ describe('index', function() {
     jasmine.DEFAULT_TIMEOUT_INTERVAL = originalTimeout
   })
 
-  it('should log only on browser environments', () => {
-    // Pass unsupported check
-    const nowFn = window.performance.now
-    window.performance.now = undefined
-
-    spyOn(console, 'log')
-
-    require('../../src/')
-
-    expect(console.log).toHaveBeenCalledWith(
-      '[Elastic APM] platform is not supported!'
-    )
-
-    window.performance.now = nowFn
-  })
-
   it('should init ApmBase', function(done) {
     var apmServer = apmBase.serviceFactory.getService('ApmServer')
     if (globalConfig.useMocks) {
--- a/packages/rum/test/specs/index.spec.js
+++ b/packages/rum/test/specs/index.spec.js
@@ -104,6 +104,13 @@ describe('index', function() {
     }
   })
 
+  it('should return same instance when loaded multiple times', () => {
+    require('../../src/')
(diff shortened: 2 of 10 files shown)


vigneshshanmugam:
@jahtalab Can i get your review before I do the rebase, that wont have a huge change as it would be simply point to rum-core.

vigneshshanmugam:
> agent is loaded in the page twice

This is a common problem and we have to deal with it as you said. There are couple of options. 

### 1. Exposing `apmBase, apm` from `@elastic/apm-rum-*`(react/angular/vue) packages 

This would let users to use to initialise the agent from the imports of a single package and would remove the need to import `@elastic/apm-rum` for initialisation purpose alone.  This is already done in `VuePlugin` and `AngularService` but `react` package doesn't expose it. 


### 2. Relying on `window.elasticApm` in framework integrations

This has huge disadvantages. 

+ We expect the users to load `apmbase` before the framework package which would result in `window.elasticApm` not being found if any of the modules was downloaded `async` and we cannot dynamically import if the package isn't found. 

Looking at both options, I would say **No.1** is an easy solution and best approach. I have created an issue stating the above reasons #799.


In any case, this PR is not related to how exports are used. Instead it fixes the issue if the user imported our packages by mistake multiple times and our library has to deal with it.

packages/rum/src/index.js
```diff
@@ -23,18 +23,31 @@
  *
  */
 
-import bootstrap from './bootstrap'
+import { bootstrap, isBrowser } from './bootstrap'
```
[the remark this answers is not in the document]
vigneshshanmugam: Sure.

packages/rum-core/src/common/bootstrap.js
```diff
@@ -23,24 +23,17 @@
...
  */
 
-import { isPlatformSupported } from './common/utils'
-import { patchAll } from './common/patching'
-import { bootstrapMetrics } from './performance-monitoring/metrics'
+import { isPlatformSupported, isBrowser } from './utils'
+import { patchAll } from './patching'
+import { bootstrapMetrics } from '../performance-monitoring/metrics'
 
-let alreadyBootstrap = false
```
vigneshshanmugam, line 30: This logic is redundant as its handled while creating `apmBase` singleton instance.

packages/rum/test/specs/index.spec.js
```diff
@@ -46,22 +46,6 @@ describe('index', function() {
     jasmine.DEFAULT_TIMEOUT_INTERVAL = originalTimeout
   })
 
-  it('should log only on browser environments', () => {
```
jahtalab, line 49: What do you think about changing this test to require `index.js` twice and test for the right condition

vigneshshanmugam: the test is moved to `bootstrap.spec.js`. Why do we need another test here?

jahtalab: Just as a smoke test, and also, to test the actual scenario of including the agent twice which the bootstrap.spec.js doesn't test

vigneshshanmugam: Ahh got it. yes will do 👍

packages/rum/src/index.js
```diff
@@ -23,17 +23,34 @@
...
 import ApmBase from './apm-base'
 
-const enabled = bootstrap()
-const serviceFactory = createServiceFactory()
-const apmBase = new ApmBase(serviceFactory, !enabled)
+/**
+ * Use a single instance of ApmBase across all instance of the agent
+ * including the instanes used in framework specific integrations
+ */
+function getApmBase() {
```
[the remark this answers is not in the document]
vigneshshanmugam: Just a personal preference. It looked a bit cleaner than having to reassign `apmBase` like this
```js
let apmBase

if (isBrowser && window.elasticApm) {
  apmBase = window.elasticApm
} else {
  const enabled = bootstrap()
  const serviceFactory = createServiceFactory()
  apmBase = new ApmBase(serviceFactory, !enabled)

  if (isBrowser) {
    window.elasticApm = apmBase
  }
}
export {apmBase}

```

PR #798, opened 2020-05-26
fix(rum-core): consider user defined type of high precedence

### LINKED ISSUE #758: React withTransaction not redefining transaction type
I'm using `withTransaction` from the react package, but its behaving strangely. I'm still getting `route-change` events, but not the custom `component` events that should be happening (I'm following the [elastic docs](https://www.elastic.co/guide/en/apm/agent/rum-js/5.x/react-integration.html#_instrumenting_individual_react_components) to create the custom `component` transactions). When I turn logging to debug, it says the transaction type is redefined at first, but it doesn't appear to stick once the transaction is actually done? The custom name does appear to stick

image

vigneshshanmugam:
+ fix #758 
+ User defined type for managed transaction should be considered a high priority as it would let users to filter these transactions in the UI in a better way. This applies only to managed transaction and custom transactions is not touched in this PR. 
+ Also moved and improved the logging after the redefining as it gives false information as mentioned in the linked issue.

### DIFF
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -143,29 +143,20 @@ class TransactionService {
       const currentTypeOrder = TRANSACTION_TYPE_ORDER.indexOf(tr.type)
       const redefineTypeOrder = TRANSACTION_TYPE_ORDER.indexOf(type)
 
-      if (currentTypeOrder >= 0) {
-        /**
-         * If the redefined type is not present in the predefined order, that implies
-         * it's a user defined type and it is of higher precedence
-         */
-        if (redefineTypeOrder === -1) {
-          redefineType = type
-        } else if (
-          /**
-           * Update type based on precedence defined in TRANSACTION_TYPE_ORDER.
-           * If either orders don't exist we also don't redefine the type.
-           */
-          redefineTypeOrder >= 0 &&
-          redefineTypeOrder < currentTypeOrder
-        ) {
-          redefineType = type
-        }
+      /**
+       * Update type based on precedence defined in TRANSACTION_TYPE_ORDER.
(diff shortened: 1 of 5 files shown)


vigneshshanmugam:
@jahtalab Integration test failure is unrelated, Good to be reviewed.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -139,28 +139,36 @@ class TransactionService {
...
+        } else if (
+          /**
+           * Update type based on precedence defined in TRANSACTION_TYPE_ORDER.
+           * If either orders don't exist we also don't redefine the type.
+           */
+          redefineTypeOrder >= 0 &&
+          redefineTypeOrder < currentTypeOrder
+        ) {
+          redefineType = type
+        }
```
jahtalab, line 162: I think this if can be written as follows and it would have the same effect:

```suggestion
        if (redefineTypeOrder < currentTypeOrder) {
          redefineType = type
        }
```

vigneshshanmugam: try harding `if else` 😅

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -139,28 +139,27 @@ class TransactionService {
...
+       */
+      if (currentTypeOrder >= 0 && redefineTypeOrder < currentTypeOrder) {
+        redefineType = type
+      }
       if (__DEV__) {
         this._logger.debug(
           `redefining transaction(${tr.id}, ${tr.name}, ${tr.type})`,
           'to',
-          `(${name}, ${type})`,
+          `(${name || tr.name}, ${redefineType})`,
```
vigneshshanmugam, line 159: sometimes, the name is undefined so it's better to log the redefined name without any changes.

PR #800, opened 2020-05-27
feat(rum-react): use correct path when route is path array

### LINKED ISSUE #702: Support path array
For `path`, react-router accepts "Any valid URL path or array of paths that path-to-regexp@^1.7.0 understands.".  But `ApmRoute` expects a string.

I suppose supporting an array of paths would require checking which path actually matched, which is a bit of a pain the way that `ApmRoute` is currently doing things.

I think it would be better if there was another component which just created a transaction while mounted, e.g.

```
<Route>
   <ApmTransaction name={path} type="route-change">
      ...
   </ApmTransaction>
</Route>
```

This would make it easier to use Apm in other routing systems besides react-router as well as helping solve this current issue.

vigneshshanmugam:
+ fix #702 
+ Uses the current path from router history if the path is provided as an array. Since the component is re rendered on other path mount, we wont create transaction as its the correct behaviour.


package.json
```diff
@@ -149,8 +149,8 @@
     "puppeteer": "^1.20.0",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
-    "react-router": "^4.2.0",
-    "react-router-dom": "^4.2.2",
+    "react-router": "^5.2.0",
```
vigneshshanmugam, line 152: path array is supported from version 5, so updated to latest to test. We dont use any breaking changes so should be good here.

vigneshshanmugam:
@jahtalab We will not be able to capture the current matched route in the `ApmRoute` component using any react lifecycle or hook since the path matching algorithm relies inside the rendering of `Route` component 
https://github.com/ReactTraining/react-router/blob/21a62e55c0d6196002bd4ab5b3350514976928cf/packages/react-router/modules/Route.js#L38-L42

So the only way I found was to compute the matched path inside our higher order component `ApmComponent` which has access to the passed in props to the component being mounted. 
https://github.com/ReactTraining/react-router/blob/21a62e55c0d6196002bd4ab5b3350514976928cf/packages/react-router/modules/Route.js#L64

Hope this explains the reason for doing it inside `ApmComponent` and not inside `ApmRoute`. Let me know if you need more details or any ideas.  

> I also considered if this would be useful for withTransaction alone, but I don't think there's a valid use-case for that. 

This is not the use case i was trying to address. So we can eliminate this requirement.

jahtalab:
Thanks for the background @vigneshshanmugam , My suggestion is to extend the functionality of `withTransaciton` to be able to change the transaction created through that without leaking the logic from ApmRoute into withTransaction. An example would be, if withTransaction accept a third argument as a callback which has the transaction and props, i.e.:

`withTransaction('name','type',(transaction, props)=>{transaction.name = ...})(Component)`

This has the added advantage that users can add keys from props to transaction labels if needed.

vigneshshanmugam:
> This has the added advantage that users can add keys from props to transaction labels if needed.

Good suggestion. I like the approach, Let me change the PR to account for it.

packages/rum-react/src/get-apm-route.js
```diff
@@ -48,7 +61,16 @@ function getApmRoute(apm) {
        */
       if (initial || pathChanged) {
         return {
-          apmComponent: withTransaction(path, 'route-change')(component),
+          apmComponent: withTransaction(
+            path,
```
jahtalab, line 65: Check the path for array here as well.

packages/rum-react/test/specs/get-apm-route.spec.js
```diff
@@ -105,6 +105,39 @@ describe('ApmRoute', function() {
...
+
+    const transactionService = serviceFactory.getService('TransactionService')
+    const dummyTr = {
+      name: 'test',
+      detectFinish: () => {}
+    }
+    spyOn(transactionService, 'startTransaction').and.returnValue(dummyTr)
+
+    const Home = () => 'home'
+    const Topics = () => 'Topics'
```
jahtalab, line 119: Maybe change one of these into a class to test both types

packages/rum-react/src/get-apm-route.js
vigneshshanmugam: Its not useful as it will always be array if its passed as array and `match` would always be null.

PR #803, opened 2020-05-29
fix(rum-core): capture tbt after all task entries are observed

### LINKED ISSUE #802: capture TBT after all long tasks are captured for page-load
+ The current TBT calculation has a bug as our observer would run whenever the entry is found and we could potentially capture TBT multiple times whenever there are long tasks are triggered during the time frame of page-load. 

One easy solution is to take records of the observed entries and act on it post page-load.

vigneshshanmugam:
+ fix #802 
+ On very long task entry, we populate the `module level metrics` constant and update the duration and use it once the page load is done to capture the TBT as spans. 
+ We have to do something similar for `CLS` as well.

### DIFF
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -33,9 +33,14 @@ import { noop, PERF } from '../common/utils'
 import Span from './span'
 
 export const metrics = {
-  fcp: 0
+  fcp: 0,
+  tbt: {
+    start: Infinity,
+    duration: 0
+  }
 }
 
+const LONG_TASK_THRESHOLD = 50
 /**
  * Create Spans for the long task entries
  * Spec - https://w3c.github.io/longtasks/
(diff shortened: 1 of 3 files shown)


vigneshshanmugam:
Test failures are related to `apm-server` 8.0.0 snapshot.

vigneshshanmugam:
>  In the original issue you mention the buffering issue with long tasks but set buffered: false for long tasks, would please clarify the issue a bit?

I have updated the description of the issue, It does not matter if buffered is true or false, this double capture exists still. 

> In any case, I think we might have to change how we capture the PerformanceObserver events with regards to buffering.

Current implementation takes buffering in to account as well, Buffering would only just produce more events, It will not change the order event arrives so I dont follow you on this. In any case, lets tackle it in separate PR. 

> Also, WDYT about adding a test for the specific scenario that might result in capturing tbt multiple times?

It was quite hard to simulate it, I will give it a try and update the PR.

PR #813, opened 2020-06-11
feat: use page visibilityState for browser responsiveness check

### LINKED ISSUE #295: replace browserresponsive check in favor of page visibility checks
+ Right now we have a scheduled timer running in the background when the agent is started and we check the browser being responsive before the transaction is sent to the APM server.

+ Also that check s run for page-load transaction which we don't even use it for discarding the transaction. It's effectively not useful in many cases. 

+ Instead of this expensive algo, we can simply use the https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API to look for the visibility state and ignore the transaction.

jahtalab:
This PR adds document visibilityState to check for browser responsiveness. The bootstrap function is added to keep all side-effects in one place.

- [x] Remove visibility check in `metrics.js`

Closes #295


packages/rum-core/src/performance-monitoring/bootstrap.js
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  window.addEventListener(
+    'visibilitychange',
+    () => {
+      let { lastHidden } = state
+      if (document.visibilityState === 'hidden') {
+        state.lastHidden = {
+          start: performance.now()
```
vigneshshanmugam, line 41: IMO, we can just use this time and rely on it to determine if the page was ever backgrounded and discard the transaction. Do you see any problems with this?

packages/rum-core/src/performance-monitoring/bootstrap.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+      } else if (lastHidden) {
+        lastHidden.duration = lastHidden.start - performance.now()
```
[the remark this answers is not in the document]
jahtalab: I was planning to add a threshold for the background duration but I decided against it. I've removed this now.

packages/rum-core/src/performance-monitoring/bootstrap.js
```diff
@@ -0,0 +1,49 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+export let state = {}
+
+export function bootstrap() {
```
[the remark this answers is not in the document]
jahtalab: Merged with the main bootstrap file.

packages/rum-core/src/performance-monitoring/bootstrap.js
jahtalab: I'm not sure what exactly do you mean with this, can you elaborate?

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -105,7 +105,7 @@ function createLongTaskSpans(longtasks) {
 export function createFirstInputDelaySpan(fidEntries) {
   let firstInput = fidEntries[0]
 
-  if (firstInput && !metrics.wasHidden) {
+  if (firstInput) {
```
vigneshshanmugam, line 108: Are we removing this check? since we are discarding the transaction anyway incase its hidden?

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -258,6 +231,15 @@ class TransactionService {
      * all the time
      */
     this.recorder.stop()
+    let { lastHidden } = state
+    if (lastHidden && lastHidden.start > tr._start) {
+      if (__DEV__) {
+        this._logger.debug(
+          `transaction(${tr.id}, ${name}, ${type}) is ignored due to page visibilityState.`
```
vigneshshanmugam, line 238: can we follow similar convention like before? 
```suggestion
          `transaction(${tr.id}, ${name}, ${type}) was discarded! Page was hidden during the lifetime of transaction.`
```

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -258,6 +231,15 @@ class TransactionService {
...
+        )
+      }
+      return
```
vigneshshanmugam, line 241: Is this correct? Don't we expect a promise for the transaction end?

packages/rum-core/src/state.js
```diff
@@ -26,3 +26,5 @@
 const __DEV__ = process.env.NODE_ENV !== 'production'
 
 export { __DEV__ }
+
+export let state = {}
```
vigneshshanmugam, line 30: nit: const

packages/rum-core/src/bootstrap.js
```diff
@@ -43,3 +43,23 @@ export function bootstrap() {
 
   return enabled
 }
+
+export function bootstrapPerf() {
+  if (document.visibilityState === 'hidden') {
+    state.lastHidden = {
```
vigneshshanmugam, line 49: can you change it to `state.lastHiddenTime` ? Would be easier to understand and we don't end up with another depth.

packages/rum-core/src/performance-monitoring/bootstrap.js
vigneshshanmugam: I forgot what it was, been a long time 😅

packages/rum-core/src/performance-monitoring/metrics.js
jahtalab: the firstInput could still be undefined if the array is empty.

vigneshshanmugam: I meant about the `wasHidden` check.

PR #825, opened 2020-06-24
feat(rum-core): capture XHR/Fetch spans using resource timing

### LINKED ISSUE #740: capture Fetch/XHR calls using ResourceTiming API
+ If the agent is loaded asynchronously, we might not be able to patch the XHR and Fetch calls on time and we might end up losing some Spans. We can always safely rely on to Resource Timing API's and create the spans for the missed API calls.

vigneshshanmugam:
+ fix #740 
+ This PR addresses the problem when the agent is downloaded asynchronously and we miss the XHR and fetch spans that happened before the patch agent script kicks in. However, we wont be able to propagate the trace header and distributed tracing for these cases wont work. 
+ Span names & types are not in sync with the patched ones as we want to distinguish them and would make it clear for the users that distributed tracing for the spans captured via `resource.*` wont work. 
```
// resource
name - 'http://example.com/fetch'
type - 'resource.fetch'

// patched
name - 'GET http://example.com/fetch'
type - 'external.http'
```
+ Ignore the Intake API endpoint using resource name, as we could end up creating lots of spans for every events beacon and we cannot rely on bootstrap time and start time (browser bugs) for this endpoint alone. 

+ Comparison logic is modified to take two things in to account
   1. Full URL's including query params as the web page may call the same endpoint with different queries and we have to capture them as separate spans
   2. Span start time as same request with same query params could happen before our patch kicked in and we have to capture them as different spans. 

### TODO
- [x] handle when `/example request` happened before agent script kicks in and another `/example` request caught via patching. These are two separate requests and we must create two spans instead of one. 
- [x] Add more tests for the filtering logic with relative and absolute URL's. 
- [x] Ignore intake API from spans


packages/rum-core/src/performance-monitoring/capture-navigation.js
```diff
@@ -119,60 +121,67 @@ function createResourceTimingSpan(resourceTimingEntry) {
...
-       * Since IE does not support initiatorType in Resource timing entry,
-       * We have to manually filter the API calls from creating duplicate Spans
-       *
-       * Skip span creation if initiatorType is other than known types specified as part of RESOURCE_INITIATOR_TYPES
-       * The reason being, there are other types like embed, video, audio, navigation etc
-       *
-       * Check the below webplatform test to know more
-       * https://github.com/web-platform-tests/wpt/blob/b0020d5df18998609b38786878f7a0b92cc680aa/resource-timing/resource_initiator_types.html#L93
-       */
-      if (initiatorType != null) {
```
vigneshshanmugam, line 155: Removed this logic as we removed support for IE 10.

packages/rum-core/src/performance-monitoring/capture-navigation.js
```diff
@@ -356,11 +382,11 @@ function captureNavigation(transaction) {
      * Capture resource timing information as spans
      */
     const resourceEntries = PERF.getEntriesByType(RESOURCE)
-    const apiCalls = getApiSpanNames(transaction)
+    const apiSpans = getAPISpans(transaction)
```
vigneshshanmugam, line 385: TODO: Skip this filtering logic for transactions other than `page-load` as we know our script would have kicked in and patched the necessary methods.

vigneshshanmugam:
@jahtalab Nice suggestion 👍, I had the similar thought initially when i started working on RUM on why we were using filterUrls instead of patch time. I thought the reason we didn't go for that solution was due to two reasons. 
1. We cannot rely on Resource timing startTime since it can be buggy in browsers. 
2. Users can disable instrumentation of XHR and Fetch. 

But as you said, It's simpler solution without doing the `URL and for loop` dance. ? I am happy with modifying the PR. WDYT?

packages/rum-core/src/state.js
```diff
@@ -24,7 +24,11 @@
  */
 
 const __DEV__ = process.env.NODE_ENV !== 'production'
+const state = {
+  //  Time when agent code is executed and patching of modules happens
+  patchedTime: null,
+  // Time when the document is last backgrounded
+  lastHiddenStart: Number.MIN_SAFE_INTEGER
```
vigneshshanmugam, line 31: @jahtalab I added this check as we don't want to compare undefined with number values. Just an extra guard.

packages/rum-core/src/common/patching/index.js
```diff
@@ -46,10 +47,10 @@ function patchAll() {
     patchHistory(function(event, task) {
       patchEventHandler.send(HISTORY, [event, task])
     })
-
     patchEventTarget(function(event, task) {
       patchEventHandler.send(EVENT_TARGET, [event, task])
     })
+    state.patchedTime = now()
```
jahtalab, line 53: Let's move this to bootstrap and use `bootstrapTime`, it's more generic that way.

vigneshshanmugam: Sure, will do that.

packages/rum-core/src/performance-monitoring/capture-navigation.js
```diff
@@ -119,60 +120,57 @@ function createResourceTimingSpan(resourceTimingEntry) {
...
+function isCapturedByPatching(resourceStartTime, requestPatchTime) {
+  return requestPatchTime != null && resourceStartTime > requestPatchTime
+}
+
+/**
+ * Check if the given url matches APM Server's Intake
+ * API endpoint and ignore it from Spans
+ */
+function isIntakeAPIEndpoint(url) {
+  return /intake\/v\d+\/rum\/events/.test(url)
```
jahtalab, line 137: Can you comment (on the PR) why `isIntakeAPIEndpoint` is necessary given that before bootstrap time there can not be any api request since we haven't bootstrapped?

vigneshshanmugam:
@jahtalab Its just travis. The tests passed though- https://travis-ci.org/github/elastic/apm-agent-rum-js/builds/703572041. I am going to ignore it.

Issue #56, opened 2018-09-21
Handling huge cardinality of page load transaction names


jahtalab:
A little bit of background on the issue, we used included the page url as transaction name (without the query string) by default and we had the same problem even without the query string there are a lot applications that simply include ids in their url. 

Currently the default transaction name is `unknown` and not the page url so all transactions are grouped by default under `unknown` but there is an API to let users set the initial page load transaction name and it can be set to the page url by the user which probably is the easiest way to set the page name. This can create a large number of transaction names.

Some solutions:
* Implementing heuristics that tries to detects ids in the URL (I've made a POC) on the agent

* Implementing a grouping algorithm on the Kibana side.

alvarolobato:
I think we could do a two step approach here:

1. Improve the API and add an additional method other than `setInitialPageLoadName()` that would directly take the URL and strip the query string. We could instead provide a helper method to do it or instruct the user how to do it, but I would prefer having an specific function for it.

1. Allow a way to initialize an list of pattern matches that the user could define in order to strip the parameters embedded in the URL. I would try to use simple matching patterns and avoid using regex for simplicity. This matching could be done either in the agent or the server but it seems that a single matching per pageload in the agent is not a big deal and we potentially save resources on the server.

sqren:
>  I would try to use simple matching patterns and avoid using regex for simplicity.

I like minimatch for these usecases
https://github.com/isaacs/minimatch#minimatch

jahtalab:
We had a meeting around sampling and high cardinality:
- We discussed storage and network traffic reduction
- For storage we should look into aggregation and trimming data (e.g. removing spans for older transactions)
- For network traffic
   - we still need to have sampling in some form or another
   - we discussed providing config options to let the user decide which transactions are important (this can be provided through central config)
   - Another idea is to crawl the website and discover the urls and let the user choose in the UI
- High cardinality issue
   - We will provide a config option to let the user specify the url pattern (this can be configured 
 in central config or in apm-server)
   - we discussed a heuristic based solution ([POC](https://gist.github.com/jahtalab/6ef468fd4a1c8fdd2a72516def37a924))
   - we also discussed using machine learning to categorise url sections (I will do a POC on this)


cc @axw , @drewpost @vigneshshanmugam

PR #827, opened 2020-06-26
feat(rum): categorize transactions based on current url

### LINKED ISSUE #255: Provide a default name for page load transactions

### LINKED ISSUE #297: use pathname as transaction name for route change
+ While patching the history API, we use `title` from the `history.pushState` as the transaction name. Instead we need to use `path` as the default transaction name.

### Scenario (SPA)

1. go to `www.example.com` 
2. navigate to www.example.com/a , history.pushState fires with these parameters (state, title, url) -> ({}, null, '/a')

**Note: Title is also ignored by some browsers(FF)**

Right now our transactions are named as **UNKNOWN** which does not give much benefit for the user. Instead we could have transaction name as **/a**. To handle cardinality problems, We can remove query params and other hashes from the URL. We can add the full URL to the context object. 

If you want to test real website. 
1. Go to  https://www.flipkart.com/
2.  paste this JS in the console
```js
 const nativePushState = history.pushState
  if (typeof nativePushState === 'function') {
    history.pushState = function(state, title, url) {
      const task = {
          state,
          title,
          url
      }
      console.log(task)
      nativePushState.apply(this, arguments)
    }
  }
```
3. Navigate to any link from the menu bar. Check the console for the URL

/cc @jahtalab @roncohen

vigneshshanmugam:
+ fix #255 
+ fix #297 
+ solution for #56 
+ PR solves the huge cardinality problems by using a heuristic based approach to shorten the page URL to a shorten version which would help name the transactions in a user understandable way than having it as `UNKNOWN`. 
+ The slugify algorithm uses a modified version of [prior work](https://gist.github.com/jahtalab/6ef468fd4a1c8fdd2a72516def37a924) done by @jahtalab. 
+ How it works. 
    1. Depth of the path tree is always set to `2` which means `/a/b/c/d` would become `/a/b/*`. The reasoning for depth to be  2 and not 1 is because most websites uses locale as the first depth in the URL tree so we want to categorize them better upfront than just printing locales in transaction. 
    2. Trailing slashes does not create a new depth. 
    3. Path using Digits and path followed by digits are redacted. ex - `/1234/product` -> `/:id/*`
    4. Special characters more than max length of 2 are redacted.
    5. Mix of upper case and lower case characters are redacted 

Please see the [test cases](https://github.com/elastic/apm-agent-rum-js/blob/625d80deb11967bafc094de67f747c10270eddc8/packages/rum-core/test/common/slugify.spec.js) for more in-depth overview. 

The algorithm seems to be work for most of the websites i have tested by using a crawler. The setup is [here](https://github.com/vigneshshanmugam/slug-urls/) if you want to play around.


packages/rum-core/src/common/slugify.js
```diff
@@ -0,0 +1,99 @@
...
+      redactedParts.push(redactString)
+      redcatedBefore = true
+      continue
+    }
+
+    part && redactedParts.push(part)
+  }
+
+  const redacted =
+    origin +
```
vigneshshanmugam, line 91: I am still unsure if we want to have `origin` in the transaction name or not. The reason why I have it here is because there are cases where the user wants to have a single `serviceName` for the website but deployed across multiple domains. 

Let me know if you have any ideas here /cc @jahtalab @drewpost

packages/rum-core/src/common/slugify.js
```diff
@@ -0,0 +1,99 @@
...
+    if (
+      numberOfDigits > 3 ||
+      (part.length > 3 && numberOfDigits / part.length >= 0.3)
+    ) {
+      redactedParts.push(redactString)
+      redcatedBefore = true
+      continue
+    }
+
+    const numberofUpperCase = (part.match(upperCaseRegex) || []).length
```
vigneshshanmugam, line 73: I am going to change this `uppercase` and `lowercase` handling part with shorter version.

packages/rum-core/src/common/slugify.js
```diff
@@ -0,0 +1,99 @@
...
+
+  const redactString = ':id'
+  const end = '*'
+  const specialCharsRegex = /\W|_/g
+  const digitsRegex = /[0-9]/g
+  const lowerCaseRegex = /[a-z]/g
+  const upperCaseRegex = /[A-Z]/g
+
+  var redactedParts = []
+  var redcatedBefore = false
```
axw, line 44:
```suggestion
  var redactedBefore = false
```

vigneshshanmugam:
@axw Good suggestion, I am just worried about one thing though. If it creates too much cardinality in transaction name and our metrics based approach in APM UI might suffer due to the consequence of this change.

packages/rum-core/src/common/slugify.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Converts URL path tree in to slug based on tree depth
+ */
+export default function slugifyUrl(urlStr, depth = 2) {
+  const parsedUrl = new Url(urlStr)
+  const { origin, query, path } = parsedUrl
+  const pathParts = path.substring(1).split('/')
+
+  const redactString = ':id'
+  const end = '*'
```
jahtalab, line 37: I think the indicator for multiple paths is `**`

packages/rum-core/src/common/slugify.js
```diff
@@ -0,0 +1,99 @@
...
+  const specialCharsRegex = /\W|_/g
+  const digitsRegex = /[0-9]/g
+  const lowerCaseRegex = /[a-z]/g
+  const upperCaseRegex = /[A-Z]/g
+
+  var redactedParts = []
```
jahtalab, line 43: nit: const

packages/rum-core/src/common/slugify.js
jahtalab: IMO it's best to remove the origin and advice people to create multiple services.

jahtalab:
@axw , We've discussed the approach here and it seems there are some benefits on putting this logic in APM server. Since the logic here is used in aggregation and that should be tied to the stack version. WDYT?

axw:
@vigneshshanmugam

> Good suggestion, I am just worried about one thing though. If it creates too much cardinality in transaction name and our metrics based approach in APM UI might suffer due to the consequence of this change.

Fair enough. I'm a bit worried that people who can't use this approach, with a fixed depth of 2 (or 3), will revert to using the whole path. I guess we can deal with that if it later though, if it does become a problem.

@jahtalab 

> We've discussed the approach here and it seems there are some benefits on putting this logic in APM server. Since the logic here is used in aggregation and that should be tied to the stack version. WDYT?

What are those benefits? Performance? I'd be interested to know what the overhead of this is; it's only for page loads, so doesn't seem like it would be significant. On the other hand, pushing it to the server means concentrating the performance cost, which might end up very expensive when your application has many users.

To me it feels simpler overall to set it in the agent, as it's just another agent as far as the server is concerned. Setting it in the agent provides greater flexibility, whereas the server would have only limited configuration; we wouldn't be adding arbitrary script execution. I'd also be a bit wary of creating an expectation of transaction name overriding for non-RUM, which won't work because of client-side breakdown aggregations.

jahtalab:
@axw , the main benefit of having this on the Apm server is alignment with the stack version release, in other words since this is used by the apm server for aggregation, any changes to this logic will affect the results that are stored in ES and therefore should be released with proper versioning (e.g. if there's a breaking change). If this doesn't seem like an issue to you, I'm ok with moving forward with the current implementation.

The performance on the agent side is not a big concern (although we will verify this with numbers)

axw:
Transaction names have always been used for grouping and aggregation (in the UI), so I don't see a need to tie this to a stack version. I'm happy to continue with the current approach.

vigneshshanmugam:
> Fair enough. I'm a bit worried that people who can't use this approach, with a fixed depth of 2 (or 3), will revert to using the whole path. I guess we can deal with that if it later though, if it does become a problem.

Yeah we can detect the whole path usage and can apply the same `redact` logic and provide a warning in the agent itself. 

> The performance on the agent side is not a big concern (although we will verify this with numbers)

I will post the numbers. But last time i checked, it wasn't hitting us hard.

vigneshshanmugam:
### Benchmark numbers
```
Unique transaction count -  ~650000
Unique transaction Names generated - 72
```

Here are the results
```js
{ 
  mean: 0.004329631194898993,
  median: 0.001528,
  p75: 2.633728,
  p90: 2.633728,
  p99: 2.633728 
}

```

The impact is super minimal and it will not affect the agent code that much as we still would ask the user to set the `pageLoadTransactionName` if they need more granular control.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -263,7 +264,7 @@ class TransactionService {
           const pageLoadTransactionName = this._config.get(
             'pageLoadTransactionName'
           )
-          if (name === NAME_UNKNOWN && pageLoadTransactionName) {
```
vigneshshanmugam, line 266: history pushState test fails due to this change. Do we want this logic to exist here @hamid?

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -274,6 +275,13 @@ class TransactionService {
             tr.spans.push(createTotalBlockingTimeSpan(metrics.tbt))
           }
         }
+        /**
+         * Categorize the transaction based on the current location
+         */
+        if (tr.name === NAME_UNKNOWN) {
+          tr.name = slugifyUrl(window.location.href)
+        }
```
jahtalab, line 283: There might be a race condition here, since we execute this in the promise, there's a possibility that we capture the url for the next route if the route change happens before this point.

packages/rum-core/src/common/slugify.js
jahtalab: Are you still planning to do this?

packages/rum-core/test/common/slugify.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+  it('handle trailing slash', () => {
+    validate('/a/', '/a')
+    validate('/a', '/a')
+    validate('/', '/')
+  })
+
+  it('handle query param', () => {
+    validate('/a/b/?id=hello', '/a/b?{query}')
+    validate('/?foo=bar&bar=baz', '/?{query}')
```
jahtalab, line 46: Should we remove the query string altogether? I think it's not useful in the transaction name.

vigneshshanmugam: Hmm, but we need a differentiation between search landing page and search query page in ecommerce context.

```
/search
vs
/search?{query}
```

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: But that would be redefined and we would still end up capturing the correct paths provided by the `route-change` transactions. So i guess we are safe here.

vigneshshanmugam:
@jahtalab I thought about it, Utils file in itself is so huge and putting this functionality there would increase it more and makes it bit harder  to glimpse through. I am in favor of keeping it separate file as its more cleaner.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: Not sure if that helps, since the href will be changed by the routing framework.

vigneshshanmugam: For all the instrumented frameworks, this does not create any issues as you see from the E2E tests. However for non-supported frameworks or routers, we will end up capturing `href` and its anyways a better solution than having `UNKNOWN`. Do you agree?

vigneshshanmugam:
I will do the docs change on a separate PR.

PR #831, opened 2020-07-01
feat(rum-core): populate destination info for NT spans

### LINKED ISSUE #829: Populate destination information for navigation timing spans
+ When the backend generates [dynamic HTML](https://www.elastic.co/guide/en/apm/agent/java/current/public-api.html#api-ensure-parent-id) and injects the RUM agent, the RUM agent would need to populate the destination information for at least one of navigation timing span (`Request and Receiving Document`) to connect all the services in the Service Maps UI. 

step_1_transaction
step_1_service_map

vigneshshanmugam:
+ fix #829 
+ This PR populates the destination information for span that is responsible for Initial HTML generation which comes under page-load transaction. So an example service request page from `www.example.com` would have a service map linked to `www.example.com` since the HTML fro the service is provided from that origin. This is in addition to other static resources like `JS, CSS, Images etc`. This would solve for both static and dynamically generated HTML's.

### DIFF
--- a/packages/rum-core/test/common/context.spec.js
+++ b/packages/rum-core/test/common/context.spec.js
@@ -165,6 +165,25 @@ describe('Context', () => {
     })
   })
 
+  it('should add navigation timing span context', () => {
+    const url = 'https://example.com'
+    const span = new Span(url, 'hard-navigation')
+    span.end()
+    addSpanContext(span, { url })
+
+    expect(span.context).toEqual({
+      destination: {
+        service: {
+          name: 'https://example.com',
+          resource: 'example.com:443',
+          type: 'hard-navigation'
+        },
+        address: 'example.com',
+        port: 443
+      }
+    })
+  })
+
   it('should enrich transaction with context info based on type', () => {
     const transaction = new Transaction('test', 'custom')
     const trContext = { tags: { tag1: 'tag1' } }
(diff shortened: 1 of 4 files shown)


packages/rum-core/test/common/context.spec.js
```diff
@@ -165,6 +165,25 @@ describe('Context', () => {
...
+    const span = new Span(url, 'hard-navigation')
+    span.end()
+    addSpanContext(span, { url })
+
+    expect(span.context).toEqual({
+      destination: {
+        service: {
+          name: 'https://example.com',
+          resource: 'example.com:443',
+          type: 'hard-navigation'
```
jahtalab, line 179: What is the reason for using a separate type here? Would it make a difference in the UI?

vigneshshanmugam: I kept it similar to the `span.type`. I went over the UI code to see the difference. The UI relies on the `span.type` field for the icons and not the `span.destination.service.type`. So if we need the same treatment as for Resources, We would need to change our Navigation timing spans `type` and `subtype`. 

Icon logic - https://github.com/elastic/kibana/blob/e70fcc708e727ba6c2b0d4145b136ef3d75a42a5/x-pack/plugins/apm/public/components/app/ServiceMap/icons.ts#L94-L104

Resource spans get `globe` icon - https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/public/components/app/ServiceMap/icons/globe.svg

PR #833, opened 2020-07-03
add apiVersion config option to typescript

marksmithson:
The apiVersion config option was added in 5.2.0 https://www.elastic.co/guide/en/apm/agent/rum-js/current/release-notes-5.x.html as part of this issue https://github.com/elastic/apm-agent-rum-js/issues/768.

However the option was not added to index.d.ts making it inaccessible when using typescript

### DIFF
--- a/packages/rum/src/index.d.ts
+++ b/packages/rum/src/index.d.ts
@@ -105,7 +105,7 @@ declare class Span extends BaseSpan {
 }
 
 interface AgentConfigOptions {
-  apiVersion?: 2|3
+  apiVersion?: 2 | 3
   serviceName?: string
   serverUrl?: string
   serviceVersion?: string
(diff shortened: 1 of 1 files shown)


marksmithson:
agreement now signed

packages/rum/src/index.d.ts
```diff
@@ -105,6 +105,7 @@ declare class Span extends BaseSpan {
 }
 
 interface AgentConfigOptions {
+  apiVersion?: number
```
vigneshshanmugam, line 108: Thanks @marksmithson for the PR. Can you restrict this to either it can be number 2 or 3? Right now the agent sends the events to Intake API `v2` by default and only if its set to 3, we enable payload compresion. 

Also i feel by restricting to 2 or 3, We know we can drop the other one when doing major version changes.

marksmithson:
fixed linting issue `npm run lint` completes without errors.

PR #834, opened 2020-07-03
docs: add default categorisation for page load transaction

vigneshshanmugam:
+ Based on the PR #827 , We no longer name the transaction to `UNKNOWN` instead use our own algorithm to group the transactions. 
+ I removed the custom page load transaction section and added it as part of the config as our logic got changed and we already do the proposed code changes by default in agent itself.


vigneshshanmugam:
@bmorelli25 Can you help me with setting up redirects for `custom-transaction-name` to `page-load-transaction-name`?

docs/configuration.asciidoc
```diff
@@ -192,8 +192,29 @@ This option overrides the ID of the span that is generated for receiving the ini
...
+This option sets the name for the page load transaction. By default, Transaction names for Hard and Soft navigations are 
+inferred by the agent based on the current URL. Let's look at the below example
+
+[source,js]
+----
+// Blog Posts
+https://www.elastic.co/blog/reflections-on-three-years-in-the-elastic-public-sector
+https://www.elastic.co/blog/say-heya-to-the-elastic-search-awards
+https://www.elastic.co/blog/and-the-winner-of-the-elasticon-2018-training-subscription-drawing-is
+
```
jahtalab, line 204: Maybe add the inferred transaction name here e.g. Transaction: `/blog/:id`

docs/configuration.asciidoc
```diff
@@ -192,8 +192,29 @@ This option overrides the ID of the span that is generated for receiving the ini
...
+// Documentation
+https://www.elastic.co/guide/en/elastic-stack/current/index.html
+https://www.elastic.co/guide/en/apm/get-started/current/index.html
+https://www.elastic.co/guide/en/infrastructure/guide/current/index.html
+----
+
+The page load transactions for the above URL's would be inferred automatically and transaction names will be categorized
+as `/blog/:id` and `/guide/en/*` by the agent. A common pattern to name the transactions would be to use the current
+URL(window.location.href). However, it creates too many unique transactions(blog titles, query strings, etc) 
+and would be less useful when visualizing the traces in UI.
```
jahtalab, line 214: Add this "A common pattern to name the transactions would be to use the current
URL(window.location.href). However, it creates too many unique transactions(blog titles, query strings, etc) 
and would be less useful when visualizing the traces in UI." to the top, it is more like an introduction for the feature.

docs/configuration.asciidoc
```diff
@@ -192,8 +192,29 @@ This option overrides the ID of the span that is generated for receiving the ini
...
+
+If the default transaction names are not helpful, please set `pageLoadTransactionName` to something
+meaningful that groups transactions under the same categories(blog, guide, etc) and avoid using the current URL.
```
jahtalab, line 217: Space after parentheses, it's not a function 😄

jahtalab:
@vigneshshanmugam , I also think it would make more sense to move the explanation for slugify to `custom-transaction-name` page instead of the config page.

docs/configuration.asciidoc
```diff
@@ -192,8 +192,29 @@ This option overrides the ID of the span that is generated for receiving the ini
 
 * *Type:* String
 
-This option sets the name for the page load transaction.
-See the <<custom-transaction-name,custom initial page load transaction names>> documentation for strategies on using this method.
+This option sets the name for the page load transaction. By default, Transaction names for Hard and Soft navigations are 
+inferred by the agent based on the current URL. Let's look at the below example
+
+[source,js]
```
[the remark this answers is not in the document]
vigneshshanmugam: Just copy pasted from before, I will change it to html?

docs/configuration.asciidoc
vigneshshanmugam: Wow, Is that true? Are you saying I have been doing this all wrong? 🤣

vigneshshanmugam:
@jahtalab I am. trying to reduce the context switch for the users when it comes to features. Our docs has already bloated so much and I don't find it good at the moment. Let's talk about this in our sync.

docs/configuration.asciidoc
vigneshshanmugam: Yeah good point.

docs/configuration.asciidoc
bmorelli25: And don't forget the period after etc., too. In American English (the language of our docs), abbreviations are always followed by a period. If it helps, think of the abbreviation for doctor: "Dr. Smith entered the room." When et cetera is abbreviated, it gets a period too: "...categories (blog, guide, etc.) and..."

vigneshshanmugam: TIL :D Thank you both 👍

vigneshshanmugam: The weird thing is, I ran it over Grammarly and it didn't complain 🤷

vigneshshanmugam:
@bmorelli25 We decided to keep it under the advanced section so please ignore the redirect part. We are good with current structure.

docs/advanced-topics.asciidoc
```diff
@@ -41,6 +41,45 @@ RUM agent publishes the type definitions for the `@elastic/apm-rum` package via
 If you are importing the package on a TypeScript codebase, you will get automatic code completion, hover info
 and method signature information on supported platforms (Visual Studio Code, etc)
 
+[[custom-transaction-name]]
+=== Custom page load transaction names
+
+A common pattern to name the transactions would be to use the current URL(window.location.href).
```
bmorelli25, line 47: Just noticed this

```suggestion
A common pattern to name the transactions would be to use the current URL (window.location.href).
```

docs/advanced-topics.asciidoc
```diff
@@ -41,6 +41,45 @@ RUM agent publishes the type definitions for the `@elastic/apm-rum` package via
...
+// Documentation - '/guide/en/*'
+https://www.elastic.co/guide/en/elastic-stack/current/index.html
+https://www.elastic.co/guide/en/apm/get-started/current/index.html
+https://www.elastic.co/guide/en/infrastructure/guide/current/index.html
+----
+
+The page load transaction names for the above URL's would be inferred automatically and categorized
+as `/blog/:id` and `/guide/en/*` by the agent. The grouping logic in the agent works by recursively traversing the
+URL path tree until the depth of 2 and converts them to wildcard or slugged matches based on the number of digits, special characters,
+the mix of upper and lowercase characters in the path.
```
jahtalab, line 70:
```suggestion
URL path tree until the depth of 2 and converting them to wildcard or slugged matches based on the number of digits, special characters,
the mix of upper and lowercase characters in the path. The algorithm is based on heuristics that are derived from common patterns in URLs and therefore, it might not correctly identify matches in some cases.
```

jahtalab: Just to set the expectation from the algorithm, I think we should add this part ^

docs/advanced-topics.asciidoc
```diff
@@ -41,6 +41,45 @@ RUM agent publishes the type definitions for the `@elastic/apm-rum` package via
...
+
+If the inferred transaction names are not helpful, please set <<page-load-transaction-name,`pageLoadTransactionName`>> configuration to something
+meaningful that groups transactions under the same categories (blog, guide, etc.) and avoid using the current URL at all costs.
```
jahtalab, line 73:
```suggestion
meaningful that groups transactions under the same categories (blog, guide, etc.) and avoid using the full URL at all costs.
```

docs/advanced-topics.asciidoc
vigneshshanmugam: Sure, I will add your suggestions.

docs/advanced-topics.asciidoc
```diff
@@ -41,6 +41,45 @@ RUM agent publishes the type definitions for the `@elastic/apm-rum` package via
...
+// Documentation - '/guide/en/*'
+https://www.elastic.co/guide/en/elastic-stack/current/index.html
+https://www.elastic.co/guide/en/apm/get-started/current/index.html
+https://www.elastic.co/guide/en/infrastructure/guide/current/index.html
+----
+
+The page load transaction names for the above URL's would be inferred automatically and categorized
+as `/blog/:id` and `/guide/en/*` by the agent. The grouping logic in the agent works by recursively traversing the
+URL path tree until the depth of 2 and converting them to wildcard or slugged matches based on the number of digits, special characters,
+the mix of upper and lowercase characters in the path. The algorithm is based on heuristics that are derived from common patterns in URLs and therefore, it might not correctly identify matches in some cases.
```
vigneshshanmugam, line 70:
```suggestion
the mix of upper and lowercase characters in the path. The algorithm uses heuristics that are derived from common patterns in URL's and therefore, it might not correctly identify matches in some cases.
```

PR #838, opened 2020-07-07
feat: add CLS, FID and TBT to experience field

### LINKED ISSUE #795: capture Cumulative Layout Shift as span
What - FID is a metric that measures a page's responsiveness during load. As such, it only focuses on input events from discrete actions like clicks, taps, and key presses.
SPEC - https://github.com/WICG/layout-instability/blob/master/README.md
Blog - https://web.dev/cls/

jahtalab:
This PR adds CLS, FID and TBT to the ['experience' field](https://github.com/elastic/apm-server/issues/3972)

+ fix #795 

I think we should find a better place for storing metrics such as TBT and CLS.
Furthermore, we have too many places to store RUM specific data (page, marks, etc.) and I think we can take this opportunity to merge these together. I'm going to make a proposal for this on apm-server.

### DIFF
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -164,6 +164,12 @@ export function calculateTotalBlockingTime(longtaskEntries) {
   })
 }
 
+/**
+ * Calculates Cumulative Layout Shift using
+ * 'layout-shift' entries captured through PerformanceObserver.
+ * https://developer.mozilla.org/en-US/docs/Web/API/LayoutShift
+ * https://web.dev/cls/
+ */
 export function calculateCumulativeLayoutShift(clsEntries) {
   clsEntries.forEach(entry => {
     if (!entry.hadRecentInput) {
@@ -243,6 +249,7 @@ export function captureObserverEntries(list, { capturePaint }) {
   const fidEntries = list.getEntriesByType(FIRST_INPUT)
   const fidSpan = createFirstInputDelaySpan(fidEntries)
   if (fidSpan) {
+    metrics.fid = fidSpan._start
     result.spans.push(fidSpan)
   }
 
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -27,7 +27,8 @@ import {
   LONG_TASK,
   LARGEST_CONTENTFUL_PAINT,
   FIRST_CONTENTFUL_PAINT,
-  FIRST_INPUT
+  FIRST_INPUT,
+  LAYOUT_SHIFT
 } from '../common/constants'
 import { noop, PERF } from '../common/utils'
 import Span from './span'
@@ -37,7 +38,8 @@ export const metrics = {
   tbt: {
     start: Infinity,
     duration: 0
-  }
+  },
+  cls: 0
 }
 
 const LONG_TASK_THRESHOLD = 50
@@ -162,6 +164,14 @@ export function calculateTotalBlockingTime(longtaskEntries) {
   })
 }
 
+export function calculateCumulativeLayoutShift(clsEntries) {
+  clsEntries.forEach(entry => {
+    if (!entry.hadRecentInput) {
+      metrics.cls += entry.value
+    }
+  })
+}
+
 /**
  * Captures all the observed entries as Spans and Marks depending on the
  * observed entry types and returns in the format
@@ -237,6 +247,8 @@ export function captureObserverEntries(list, { capturePaint }) {
   }
 
   calculateTotalBlockingTime(longtaskEntries)
+  const clsEntries = list.getEntriesByType(LAYOUT_SHIFT)
+  calculateCumulativeLayoutShift(clsEntries)
 
   return result
 }
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -34,6 +34,7 @@ import { noop, PERF } from '../common/utils'
 import Span from './span'
 
 export const metrics = {
+  fid: 0,
   fcp: 0,
   tbt: {
     start: Infinity,
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -31,7 +31,13 @@ import {
   metrics,
   createTotalBlockingTimeSpan
 } from './metrics'
-import { extend, getEarliestSpan, getLatestNonXHRSpan } from '../common/utils'
+import {
+  extend,
+  getEarliestSpan,
+  getLatestNonXHRSpan,
+  isUndefined,
+  isPerfTypeSupported
+} from '../common/utils'
 import { captureNavigation } from './capture-navigation'
 import {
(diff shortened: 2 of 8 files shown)


vigneshshanmugam:
@jahtalab I did a proposal before sometime about the new metric- see here https://github.com/elastic/apm/issues/267

jahtalab:
@vigneshshanmugam , I think `events` is not a good fit for metrics, these are just values without start or end, e.g. cls.

Also, having events as proposed in elastic/apm#267 is redundant IMO since we have spans that does the same thing.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -277,8 +279,14 @@ class TransactionService {
...
-          if (tr.captureTimings && metrics.tbt.duration > 0) {
-            tr.spans.push(createTotalBlockingTimeSpan(metrics.tbt))
+          if (tr.captureTimings) {
+            if (metrics.tbt.duration > 0) {
+              tr.spans.push(createTotalBlockingTimeSpan(metrics.tbt))
+            }
+            /**
+             * TODO: define the proper place to store cls.
+             */
+            tr.addContext({ cls: metrics.cls })
```
vigneshshanmugam, line 289: shouldn't we need to put it under `custom` as it would potentially break in future versions if at all schema is specd

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -277,8 +279,14 @@ class TransactionService {
...
            */
-          if (tr.captureTimings && metrics.tbt.duration > 0) {
-            tr.spans.push(createTotalBlockingTimeSpan(metrics.tbt))
+          if (tr.captureTimings) {
+            if (metrics.tbt.duration > 0) {
+              tr.spans.push(createTotalBlockingTimeSpan(metrics.tbt))
+            }
+            /**
+             * TODO: define the proper place to store cls.
+             */
```
vigneshshanmugam, line 288: We can report only non 0 values, it's wasteful to report when there is no layout shift.

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -237,6 +247,8 @@ export function captureObserverEntries(list, { capturePaint }) {
   }
 
   calculateTotalBlockingTime(longtaskEntries)
+  const clsEntries = list.getEntriesByType(LAYOUT_SHIFT)
```
vigneshshanmugam, line 250: Please add some doc comments upfront and also link to spec. Helps a lot while going over the code.

packages/rum-core/test/performance-monitoring/metrics.spec.js
```diff
@@ -245,5 +246,48 @@ describe('Metrics', () => {
         })
       )
     })
+
+    it('should calculate Cumulative Layout Shift', () => {
+      metrics.cls = 0
```
vigneshshanmugam, line 251: can you move this to `beforeEach` function.

vigneshshanmugam:
I raised that proposal based on metrics like TBT, FID which has start & end, needed to be on the timeline as opposed to marks  which are like FCP, LCP, etc. 

Yes, It is indeed close to Span but needing separate distinction in the UI, so I just picked a name as `events` and not going for Span as our UI does not have any way of showing duration events as part of Span(like how chrome devtools shows colour coded DNS, TCP entries). I feel having them as Span is not clear right now and not a good DX. In anycase, please open up the issue in APM server and we can close mine as well if yours gets accepted.

vigneshshanmugam:
@v1v Any clue on why the size report doesn't show diff? https://github.com/elastic/apm-agent-rum-js/pull/838#issuecomment-654891142

v1v:
@vigneshshanmugam , it seems the [latest master build](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Fapm-agent-rum-mbp/detail/master/293/pipeline) did not have the file with the bundlesize.json data, see the [logs](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Fapm-agent-rum-mbp%2FPR-838/detail/PR-838/1/pipeline/98#step-138-log-1)

This is a corner case when a PR with only docs gets merged into the master branch, then the bundlesize.json won't be generated.

I need to figure out if we can do something about this

vigneshshanmugam:
Thanks @v1v for the background. Hmmm, that makes sense. May be we can generate on last master vs current PR and ignore the last `bundlesize json` file being present. WDYT?

v1v:
The easy approach will be to enable the linting even for those PRs that are docs based. If the builds are not too slow then I'd say that's the less hacky approach. What do you think?

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -37,7 +38,8 @@ export const metrics = {
   tbt: {
     start: Infinity,
     duration: 0
-  }
+  },
+  cls: 0
```
jahtalab, line 42: change to `undefined`

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: We're going to differentiate between supported and unsupported browsers but still send 0 if supported!

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam, line 289: Now thats spec is kind of finalised, do you want to move it under experience and we can merge this PR?

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -280,13 +286,23 @@ class TransactionService {
...
+              tr.spans.push(createTotalBlockingTimeSpan(tbt))
+            }
+
+            tr.experience = {}
+            if (isPerfTypeSupported(LONG_TASK)) {
+              tr.experience.tbt = tbt.duration
+            }
+
+            if (isPerfTypeSupported(LAYOUT_SHIFT)) {
+              tr.experience.cls = cls
```
vigneshshanmugam, line 300: can you declare these metrics in the metrics object prior to setting it? Two things
1. makes it easier to track
2. object is in same shape, performance optimisation in engines.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -280,13 +286,23 @@ class TransactionService {
...
-            if (metrics.tbt.duration > 0) {
-              tr.spans.push(createTotalBlockingTimeSpan(metrics.tbt))
+            const { cls, fid, tbt } = metrics
+            if (tbt.duration > 0) {
+              tr.spans.push(createTotalBlockingTimeSpan(tbt))
+            }
+
+            tr.experience = {}
+            if (isPerfTypeSupported(LONG_TASK)) {
+              tr.experience.tbt = tbt.duration
```
vigneshshanmugam, line 296: Don't we need to do put this inside `tbt.duration > 0` condition?

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: Let's add fid to the metrics

PR #844, opened 2020-07-14
feat: userContext in vue plugin

yooouuri:
This makes it possible to set the userContext trough the Vue plugin.


yooouuri:
Ah did not know this was possible, then my PR is obsolete.

Thanks for pointing out!

PR #847, opened 2020-07-20
chore: add deps to run integration tests

vigneshshanmugam:
+ Integration tests are not able to run properly due to the same issue reported here for puppeteer images https://github.com/elastic/apm-integration-testing/issues/870
+ Did the same workaround as https://github.com/elastic/apm-integration-testing/issues/872


vigneshshanmugam:
The Integration tests on RUM passes without any issues, Stopped the build as its unnecessary to run saucelabs tests for this change.

PR #856, opened 2020-07-30
chore(benchmarks): update script to use new playwright API

vigneshshanmugam:
+ updates playwright to the latest version
+ fix deprecated methods, cache is disabled by default for every run in playwright itself.

### DIFF
--- a/.ci/docker/node-playwright/Dockerfile
+++ b/.ci/docker/node-playwright/Dockerfile
@@ -11,25 +11,39 @@ RUN apt-get -qq update && apt-get -qq install -y curl && \
 RUN apt-get -qq install -y git --no-install-recommends
 
 # 3. Install WebKit dependencies
-RUN apt-get -qq install -y libwoff1 \
-                       libopus0 \
-                       libwebp6 \
-                       libwebpdemux2 \
-                       libenchant1c2a \
-                       libgudev-1.0-0 \
-                       libsecret-1-0 \
-                       libhyphen0 \
-                       libgdk-pixbuf2.0-0 \
-                       libegl1 \
-                       libnotify4 \
-                       libxslt1.1 \
-                       libevent-2.1-6 \
-                       libgles2 \
-                       libvpx5
+RUN apt-get -qq install -y  --no-install-recommends \
+    libwoff1 \
+    libopus0 \
+    libwebp6 \
+    libwebpdemux2 \
+    libenchant1c2a \
+    libgudev-1.0-0 \
+    libsecret-1-0 \
+    libhyphen0 \
+    libgdk-pixbuf2.0-0 \
+    libegl1 \
+    libnotify4 \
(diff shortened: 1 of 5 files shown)


packages/rum/test/benchmarks/profiler.js
```diff
@@ -51,14 +51,9 @@ function gatherRawMetrics(browser, url) {
...
     let client = null
     try {
-      client = await browser.pageTarget(page).createCDPSession()
+      client = await context.newCDPSession(page)
     } catch (_) {}
 
-    /**
-     * Disable cache for each run
-     */
-    await page.setCacheEnabled(false)
```
jahtalab, line 60: Is the cache disabled by default in the new version?

vigneshshanmugam: Yes see this https://github.com/microsoft/playwright/pull/1231

.ci/docker/node-playwright/Dockerfile
```diff
@@ -11,25 +11,39 @@ RUN apt-get -qq update && apt-get -qq install -y curl && \
...
+    libevent-2.1-6 \
+    libgles2 \
+    libvpx5 \
+    libxcomposite1 \
+    libatk1.0-0 \
+    libatk-bridge2.0-0 \
+    libepoxy0 \
+    libgtk-3-0 \
+    libharfbuzz-icu0 \
+    libgstreamer-gl1.0-0 \
```
vigneshshanmugam, line 36: @v1v Can you help me here? I did install the required deps but the last and first run reported the same errors while testing on local, I cannot seem to find any error. Do you think it might have to do with caching? 

```
 /app/.cache/ms-playwright/webkit-1301/minibrowser-wpe/MiniBrowser: error while loading shared libraries: libgstreamer-1.0.so.0: cannot open shared object file: No such file or directory
```

https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/PR-856/2/

v1v: As you said, it seems the caching is causing this issue:
- https://github.com/elastic/apm-agent-rum-js/blob/f871eb5778ac1aa41c7d20e6bd040b17913b2ecc/dev-utils/docker-compose.yml#L162

I'd say it's required to use `--build` as part of https://github.com/elastic/apm-agent-rum-js/blob/master/.ci/scripts/benchmarks.sh#L3

Let me add that particular flag to your PR and see if it helps

v1v: See https://github.com/elastic/apm-agent-rum-js/pull/856/commits/aeea4b9e4732c0406b3595b41ec0f7c2becb8f14

vigneshshanmugam: Neat, that was my guess as well. Thanks Victor for the help 👍

PR #857, opened 2020-07-30
feat(rum-core): keep name and change message in rejection reason

clyfish:
When reason is an error object, call logErrorEvent with reason directly to keep `reason.name`.

Change unhandled rejection reason message prefix to `Uncaught (in promise) `, because chrome devtools shows unhandled promise rejection as `Uncaught (in promise) Error: message`.

And don't ignore the reason, when reason is a plain object, because to expose the error is better than to ignore it.


clyfish:
`npm run bundlesize` failed

Master branch:
```
@elastic/apm-rum:  FAIL  ./dist/bundles/elastic-apm-rum.umd.min.js: 18.05KB > maxSize 16KB (gzip)
@elastic/apm-rum:  FAIL  ./dist/bundles/elastic-apm-opentracing.umd.min.js: 19.51KB > maxSize 18KB (gzip)
```

This branch:
```
@elastic/apm-rum:  FAIL  ./dist/bundles/elastic-apm-rum.umd.min.js: 18.04KB > maxSize 16KB (gzip)
@elastic/apm-rum:  FAIL  ./dist/bundles/elastic-apm-opentracing.umd.min.js: 19.49KB > maxSize 18KB (gzip)
```

clyfish:
Without this PR, unhandled promise rejection with a Error object lose the type of error.

For example:

```
Promise.resolve(42).then(() => {
  const a = null
  console.log(a.b)
})
```

Master branch reports the error as:

```
error:
  exception:
    message: "Unhandled promise rejection: Cannot read property 'b' of null"
    type: "Unhandled promise rejection"
```

With this PR:

```
error:
  exception:
    message: "Uncaught (in promise) TypeError: Cannot read property 'b' of null"
    type: "TypeError"
```

The type of the error(TypeError) is lost on the master branch.

@vigneshshanmugam @jahtalab Please review this PR.

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -171,22 +171,24 @@ class ErrorLogging {
...
-    const prefix = 'Unhandled promise rejection: '
-    const { reason } = promiseRejectionEvent
+    const prefix = 'Uncaught (in promise) '
+    const { reason = '<no reason specified>' } = promiseRejectionEvent
+    let errorEvent
 
-    if (reason == null) {
-      this.logError(prefix + '<no reason specified>')
-    } else if (typeof reason.message === 'string') {
+    if (typeof reason.name === 'string' && typeof reason.message === 'string') {
```
vigneshshanmugam, line 178: name shouldn't be a required field here. It should work for error like objects as well. I would do this

```js
const name = reason.name || ''

this.logErrorEvent({
  error: reason,
  message: `${prefix}${name}: ${reason.message}`
})
```

vigneshshanmugam, line 178: name shouldn't be a required field here. It should work for error like objects as well. I would do this

```js
const name = reason.name ? reason.name + ':' : ''

this.logErrorEvent({
  error: reason,
  message: `${prefix}${name} ${reason.message}`
})
```

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -171,22 +171,24 @@ class ErrorLogging {
...
-      })
-    } else if (typeof reason !== 'object') {
-      this.logError(prefix + reason)
+      errorEvent = {
+        error: reason,
+        message: `${prefix}${reason.name}: ${reason.message}`
+      }
+    } else {
+      errorEvent = {
+        message: `${prefix}${reason}`
```
vigneshshanmugam, line 188: If we are going to support for all reason types, there should be a normalisation for these types.  Having `[object Object]` is not helpful.

packages/rum-core/src/error-logging/error-logging.js
vigneshshanmugam: By doing this, it would also work correctly for `Promise.reject({message: 'testerror'})`

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -171,22 +171,24 @@ class ErrorLogging {
   }
 
   logPromiseEvent(promiseRejectionEvent) {
-    const prefix = 'Unhandled promise rejection: '
-    const { reason } = promiseRejectionEvent
+    const prefix = 'Uncaught (in promise) '
```
vigneshshanmugam, line 174: Not so much opposed to naming the message in this format. Chrome and Firefox follows the new way, while Safari does with old format.

clyfish:
> Thanks for the PR @clyfish Added some comments.

Thank you @vigneshshanmugam , please review again.

packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -172,21 +172,25 @@ class ErrorLogging {
...
-      })
-    } else if (typeof reason !== 'object') {
-      this.logError(prefix + reason)
+      const name = reason.name ? reason.name + ': ' : ''
+      errorEvent = {
+        error: reason,
+        message: `${prefix}${name}${reason.message}`
+      }
+    } else {
+      reason = typeof reason === 'object' ? '<object>' : reason
```
vigneshshanmugam, line 188: Is it common to throw with any other types ? `function`?

clyfish:
@vigneshshanmugam Added assertion for the `type`

packages/rum-core/.prettierrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
vigneshshanmugam: can you remove this file as its unrelated to the PR? We have our prettier rules defined here https://github.com/elastic/apm-agent-rum-js/blob/7b02f14f626e41eb14172064afb4d51076ee379a/.eslintrc.js#L40

clyfish: It's added by mistake.
I've already removed it.

vigneshshanmugam:
Integration test failure is unrelated.

PR #862, opened 2020-08-05
chore(jenkins): increase bundlesize

cachedout:
This increases the bundle size for apm-rum and apm-opentracing. It should resolve errors seen here:

https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-rum%2Fapm-agent-rum-mbp/detail/master/308/pipeline/78/

### DIFF
--- a/packages/rum/package.json
+++ b/packages/rum/package.json
@@ -51,11 +51,11 @@
   "bundlesize": [
     {
       "path": "./dist/bundles/elastic-apm-rum*.min.js",
-      "maxSize": "20 kB"
+      "maxSize": "18.5 kB"
     },
     {
       "path": "./dist/bundles/elastic-apm-opentracing*.min.js",
-      "maxSize": "22kB"
+      "maxSize": "20kB"
     }
   ]
 }
--- a/packages/rum/package.json
+++ b/packages/rum/package.json
@@ -51,11 +51,11 @@
   "bundlesize": [
     {
       "path": "./dist/bundles/elastic-apm-rum*.min.js",
-      "maxSize": "16 kB"
+      "maxSize": "20 kB"
     },
     {
       "path": "./dist/bundles/elastic-apm-opentracing*.min.js",
-      "maxSize": "18kB"
+      "maxSize": "22kB"
     }
   ]
 }


packages/rum/package.json
```diff
@@ -51,11 +51,11 @@
   "bundlesize": [
     {
       "path": "./dist/bundles/elastic-apm-rum*.min.js",
-      "maxSize": "16 kB"
+      "maxSize": "20 kB"
```
vigneshshanmugam, line 54: Pretty hesitant to increase it by a lot, Can we keep it to `18.5kB` for this one and `opentracing ` as `20kB`?

vigneshshanmugam:
@cachedout can you rename the commit message as well? we follow conventional commits guidelines and these are the allowed ones https://github.com/elastic/apm-agent-rum-js/blob/7b02f14f626e41eb14172064afb4d51076ee379a/package.json#L43-L55

packages/rum/package.json
mdelapenya: I'd like to understand more about these checks at the test level: are they critical enough to have the limits hardcoded in the package.json file? How often do those limits change (== how often would the tests fail after resizing the bundle)?

vigneshshanmugam: It depends on features/bug fixes that goes inside the agent code that is being shipped to the users so if we are increasing it by a lot we would have to catch it during our tests and make sure the change is intended as per the plan. 

If we did merge a PR with the increased code size, then the RUM team would need to work on the optimization phase if its worth it or increase the maxSize to fix the test failures. 

Does that make sense?

mdelapenya: What about defining acceptable thresholds for increments? Something like increasing the bundle a small % of the size of the released one 🤔

vigneshshanmugam: We already have this check https://github.com/elastic/apm-agent-rum-js/pull/863#issuecomment-669099572 which shows us the thresholds, so i guess we are good here.

cachedout:
@vigneshshanmugam Sure thing. Commit messages amended and I modified the bundle size values per your suggestion. Thanks!

PR #863, opened 2020-08-05
fix(rum): use logLevel config when agent is inactive

### LINKED ISSUE #861: logLevel is ignored when agent is inactive
Currently the agent logs the message `RUM is inactive` even when the agents logLevel is set to `error` level. This is a bug and agent should use the logLevel and use it correctly. 

```js
init({
  active: false,
  logLevel: "error"
})
```

vigneshshanmugam:
+ fix #861 
+ The first logLevel update happens during the LoggingService creation phase in both active and also inactive phase.
+ The second logLevel update happens if users call `apm.config({})` or config is updated from remote config for the active phase.

### DIFF
--- a/packages/rum-core/src/common/logging-service.js
+++ b/packages/rum-core/src/common/logging-service.js
@@ -28,7 +28,7 @@ import { noop } from './utils'
 class LoggingService {
   constructor(spec = {}) {
     this.levels = ['trace', 'debug', 'info', 'warn', 'error']
-    this.level = spec.level || 'info'
+    this.level = spec.level || 'warn'
     this.prefix = spec.prefix || ''
 
     this.resetLogMethods()
@@ -39,6 +39,9 @@ class LoggingService {
   }
 
   setLevel(level) {
+    if (level === this.level) {
+      return
+    }
     this.level = level
     this.resetLogMethods()
   }
--- a/packages/rum-core/src/common/logging-service.js
+++ b/packages/rum-core/src/common/logging-service.js
@@ -44,30 +44,18 @@ class LoggingService {
   }
 
   resetLogMethods() {
-    var loggingService = this
-    this.levels.forEach(function(level) {
-      loggingService[level] = loggingService.shouldLog(level) ? log : noop
+    this.levels.forEach(level => {
+      this[level] = this.shouldLog(level) ? log : noop
 
       function log() {
-        var prefix = loggingService.prefix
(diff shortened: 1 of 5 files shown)


packages/rum/src/apm-base.js
```diff
@@ -85,7 +82,12 @@ export default class ApmBase {
         }
       } else {
         this._disable = true
-        loggingService.info('RUM agent is inactive')
+        /**
+         * Creation of LoggingService should happen only after config update.
+         * This order gurantees that we use the proper logLevel
+         */
+        const loggingService = this.serviceFactory.getService(LOGGING_SERVICE)
+        loggingService.warn('RUM agent is inactive')
```
vigneshshanmugam, line 90: moved from `info` to `warn` as our default level is warn.

packages/rum/src/apm-base.js
```diff
@@ -85,7 +82,12 @@ export default class ApmBase {
         }
       } else {
         this._disable = true
-        loggingService.info('RUM agent is inactive')
+        /**
+         * Creation of LoggingService should happen only after config update.
+         * This order gurantees that we use the proper logLevel
```
jahtalab, line 87: I think it is cleaner to just call `loggingService.setLevel` in the init function instead of deciding on the order of service creation.

packages/rum/src/apm-base.js
jahtalab: `warn` makes more sense in this case anyway :+1:

vigneshshanmugam:
@jahtalab I understand the concern and its valid, Only reason i went with this was it avoids calling multiple `resetLogMethods` during instance creation https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/src/common/service-factory.js#L39-L42 and setting levels https://github.com/elastic/apm-agent-rum-js/blob/3bee46ca626cc720235212b2e208327cb8c84bcc/packages/rum-core/src/common/service-factory.js#L72. 

We can work around the order issue if we set logLevel on `init`, but cant avoid multiple resets calls. Thoughts?

packages/rum-core/src/common/logging-service.js
```diff
@@ -39,35 +39,26 @@ class LoggingService {
   }
 
   setLevel(level) {
+    if (level === this.level) {
```
vigneshshanmugam, line 42: this avoids the problem i mentioned in this comment

PR #866, opened 2020-08-11
feat(rum-core): provide api to block managed transactions 

### LINKED ISSUE #841: Expose the task API
Currently we have the task API, but we need to document and test it properly.

vigneshshanmugam:
+ fixes #841 
+ Improves upon the work that is already on the original PR #194 
+ Adds an option `blocking` to the `startSpan` method that keeps the transaction open and allow controlling the behaviour of all managed transactions created by the agent itself through auto instrumentation or through user. 
+ Added `transaction.block(true)` method which can be used to control the closing of managed transactions. 
+ Added documentation on the custom transactions part with unmanaged and managed transactions.

### DIFF
--- a/docs/transaction-api.asciidoc
+++ b/docs/transaction-api.asciidoc
@@ -64,57 +64,51 @@ Arguments:
 
 * `options` - The following options are supported:
 
-** `blocked` - Blocks the associated transaction from ending till this span is ended. Blocked spans
-    automatically creates a task. Defaults to false
+** `blocking` - Blocks the associated transaction from ending till this span is ended. Blocked spans
+    automatically internally creates a task. Defaults to false
 
 ** `parentId` - Parent id associated with the new span. Defaults to current transaction id
 
 ** `sync` - Denotes if the span is synchronous or asynchronous.
 
-When a span is started it will measure the time until <<span-end,`span.end()`>> is called.
+When a span is started it will measure the time until <<span-end,`span.end()`>> is called. Blocked
+spans allow users to control the early closing of <<custom-managed-transactions, managed transactions>> in some cases.
 
-See <<span-api,Span API>> docs for details on how to use custom spans.
+See also the <<span-api,Span API>> docs for details on how to use custom spans.
 
 [float]
-[[transaction-add-task]]
-==== `transaction.addTask([taskId])`
+[[transaction-block]]
+==== `transaction.block(flag)`
 
 [source,js]
 ----
-const taskId = transaction.addTask('async-task')
+transaction.block(true)
 ----
 
-Starts a new task for the current transaction and returns the new task id. The transaction does not end till all the tasks are
-removed from the transacion.
+Blocks the currently running auto-instrumented or custom transactions from getting automatically closed once the associated spans are ended.
 
 Arguments:
 
-* `name` - Optional. The name of the task (string). Defaults to a random id.
+* `flag` - Boolean. Defaults to false.
 
-Tasks are pretty new and its available only on the RUM agent. Most of the time, the users shouldn't need to create tasks as
-custom transactions created by the user ends only when `transaction.end` method is called. However in the case of managed transactions like page-load, route-change and other user created ones by setting `managed: true` flag, the transaction might get closed automatically prior to the user's expectation once all the associated tasks are completed. Please check the <<custom-managed-transactions, custom transactions>> guide to understand how you can control the behaviour of transaction using tasks.
-
-
-NOTE: This API will be removed in the next major version, Please use the `startSpan` method with `blocking: true`.
-
-
-[float]
-[[transaction-remove-task]]
-==== `transaction.removeTask(taskId)`
+Most of the time, users shouldn't need to block the running transaction as the agent tries to account for all of the async network activity and capture
+the required information. The block API is useful when the user feels the auto instrumented code missed some important activity and ends up prematurely
+closing the transaction and the page contains lots of asynchronous tasks that cannot be automatically instrumented (eg: web workers).
 
 [source,js]
 ----
-transaction.removeTask('async-task')
+// get the current active autoinstrumented transaction (page-load, route-change, etc.)
+const activeTransaction = apm.getCurrentTransaction()
+
+if (activeTransaction) {
+  activeTransaction.block(true)
+}
+
+// Perform some more async tasks and unblock it
+activeTransaction.block(false)
 ----
 
-Removes the task if its present from the associated transaction and also checks if the transaction can be ended when there are no pending tasks to be completed.
-
-Arguments:
-
-* `name` - Required. The name of the task (string).
-
-
-NOTE: This API will be removed in the next major version, Please use the `startSpan` method with `blocking: true`.
+One can also achieve a similar action by creating blocked spans. Check the <<custom-managed-transactions, managed transactions>> guide to learn more.
 
 
 [float]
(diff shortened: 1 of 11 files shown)


packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -266,9 +266,9 @@ export default class PerformanceMonitoring {
         })
       }
 
-      const span = transactionService.startSpan(spanName, 'external.http')
-      const taskId = transactionService.addTask()
-
+      const span = transactionService.startSpan(spanName, 'external.http', {
+        blocked: true
```
vigneshshanmugam, line 270: might change it to `task` as it's confusing with `sync` flag.

docs/transaction-api.asciidoc
```diff
@@ -64,14 +64,58 @@ Arguments:
...
-** `sync` - Denotes if the span is blocking (sync) or non-blocking(async). 
+** `sync` - Denotes if the span is synchronous or asynchronous.
 
 When a span is started it will measure the time until <<span-end,`span.end()`>> is called.
 
 See <<span-api,Span API>> docs for details on how to use custom spans.
 
+[float]
+[[transaction-add-task]]
+==== `transaction.addTask([taskId])`
```
vigneshshanmugam, line 80: I am in favor of not documenting `addTask` and `removeTask` and instead guide users to use the `blocking` span API so we can reduce the number of concepts and also reduce confusion.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: or we can keep it to `blocking`.

docs/transaction-api.asciidoc
vigneshshanmugam: Thoughts @jahtalab

vigneshshanmugam:
Test failure is related to incorrect sauce connect version. Not sure whats happening, have to dig deeper.

vigneshshanmugam:
Saucelabs issue fixed by #868

docs/transaction-api.asciidoc
```diff
@@ -64,14 +64,58 @@ Arguments:
...
+
+Arguments:
+
+* `name` - Optional. The name of the task (string). Defaults to a random id.
+
+Tasks are pretty new and its available only on the RUM agent. Most of the time, the users shouldn't need to create tasks as
+custom transactions created by the user ends only when `transaction.end` method is called. However in the case of managed transactions like page-load, route-change and other user created ones by setting `managed: true` flag, the transaction might get closed automatically prior to the user's expectation once all the associated tasks are completed. Please check the <<custom-managed-transactions, custom transactions>> guide to understand how you can control the behaviour of transaction using tasks.
+
+
+NOTE: This API will be removed in the next major version, Please use the `startSpan` method with `blocking: true`.
```
jahtalab, line 98: Let's remove this for now, we can add if we decide to remove the task API.

docs/transaction-api.asciidoc
```diff
@@ -64,14 +64,58 @@ Arguments:
...
+----
+const taskId = transaction.addTask('async-task')
+----
+
+Starts a new task for the current transaction and returns the new task id. The transaction does not end till all the tasks are
+removed from the transacion.
+
+Arguments:
+
+* `name` - Optional. The name of the task (string). Defaults to a random id.
```
jahtalab, line 92: nit: The id is not random, but rather incremental!

packages/rum-core/src/performance-monitoring/performance-monitoring.js
jahtalab: I think using `withTask` might be a better option.

packages/rum-core/src/performance-monitoring/transaction.js
```diff
@@ -138,21 +143,16 @@ class Transaction extends SpanBase {
...
 
   addTask(taskId) {
-    if (typeof taskId === 'undefined') {
-      taskId = 'task' + this.nextAutoTaskId++
-    }
-    if (this._scheduledTasks.indexOf(taskId) == -1) {
-      this._scheduledTasks.push(taskId)
-      return taskId
+    if (!taskId) {
+      taskId = 'task-' + generateRandomId(16)
```
jahtalab, line 147: I think we should just use incremental ids here. using random Ids seems like an overkill.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
jahtalab: Also, let's make sure that this option is logged in the debug mode. This helps us debug if a transaction never ends.

packages/rum-core/src/performance-monitoring/transaction.js
vigneshshanmugam: This happens only in corner case when user doesn't provide any valid id. Its not overkill IMO, better than not having to store instance level variables.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: Thinking about it more, we want to stick to few concepts as possible and with this option setting to `task` or `withTask`, seems like we are teaching yet another concept to user. Lets think of a way to hide the task concept itself in general and provide another way to deal with it.

docs/transaction-api.asciidoc
vigneshshanmugam: Check my comment here, https://github.com/elastic/apm-agent-rum-js/pull/866#discussion_r469817535

I agree with you that users might not need to measure operations always, But WDYT about an alternative idea where the users will do

```
transaction.block(true)

transaction.block(false)
```

With this
+ we remove a new concept called Task
+ Easier for us to manage as well, We will internally still keep tasks and we can also use this signal for managed transactions.

vigneshshanmugam:
Aborted the build as it blocked on other PR #872

vigneshshanmugam:
@bmorelli25 Please provide feedback on the docs, I will address them on separate PR as i need these changes for my other PR.

docs/transaction-api.asciidoc
```diff
@@ -64,13 +64,51 @@ Arguments:
...
+Blocks the currently running auto-instrumented or custom transactions from getting automatically closed once the associated spans are ended.
 
-When a span is started it will measure the time until <<span-end,`span.end()`>> is called.
+Arguments:
+
+* `flag` - Boolean. Defaults to false.
+
+Most of the time, users shouldn't need to block the running transaction as the agent tries to account for all of the async network activity and capture
+the required information. The block API is useful when the user feels the auto instrumented code missed some important activity and ends up prematurely
+closing the transaction and the page contains lots of asynchronous tasks that cannot be automatically instrumented (eg: web workers).
```
bmorelli25, line 96: This is a bit of a run-on sentence.
```
The block API is useful when the user feels the auto instrumented code missed some important activity and ends
up prematurely closing the transaction and the page contains lots of asynchronous tasks that cannot be
automatically instrumented (eg: web workers).
```

What about this instead?
```
The block API is useful when you think auto instrumented code misses important activity and prematurely ends a transaction.
An example of this might be a page that contains lots of asynchronous tasks that cannot be automatically instrumented, like web workers.
```

docs/transaction-api.asciidoc
```diff
@@ -110,8 +148,8 @@ Defining too many unique fields in an index is a condition that can lead to a
 transaction.end()
 ----
 
-Ends the transaction. If the transaction has already ended, nothing happens.
-
+Ends the transaction. Calling `transaction.end` on an active transaction would end all of the underlying spans and marks them as `truncated`.
+If the transaction has already ended, nothing happens.
```
bmorelli25, line 152: Let's change this to present tense.

```
Ends the transaction. Calling `transaction.end` on an active transaction ends all of the underlying spans and marks them as `truncated`.
If the transaction has already ended, nothing happens.

PR #869, opened 2020-08-13
feat(rum-core): support ignore error

xincmm:
* The purpose of this pull request is to support ignoring some errors. In my project, I can safely ignore some errors, such as [https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded](https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded),but apm agent will report all errors.
I noticed that apm agent config has ignoreTransactions, Hope to add ignoreErrors.

### DIFF
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -75,6 +75,7 @@ class Config {
       logLevel: 'warn',
       breakdownMetrics: false,
       ignoreTransactions: [],
+      ignoreErrors: [],
       eventsLimit: 80,
       queueLimit: -1,
       flushInterval: 500,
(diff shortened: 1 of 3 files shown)


xincmm:
@vigneshshanmugam Thank you for your reply! Yes, I am currently using this method. but through filter, apm agent will still make a request to apm server（api：`/intake/v2/rum/events`）.

vigneshshanmugam:
Thats correct, You can set `logLevel: error` to get rid of those warnings. I understand the problem, but then we would have to introduce every other option for transactions, spans and errors. 

Transaction was added as it would let users control before the filter API was available. Would like to hear thoughts from @jahtalab as well.

xincmm:
@jahtalab Thanks. I will close this issue 😊

PR #883, opened 2020-09-01
perf(rum-core): avoid creating options when transaction exists

vigneshshanmugam:
+ `apm.startSpan` was creating options object even when there was an active transaction present which is wasteful to do in the first place. 
+ This PR improves the performance of `startSpan`method and skips options creation when there is already an active transaction.

### DIFF
--- a/packages/rum-core/test/performance-monitoring/transaction-service.spec.js
+++ b/packages/rum-core/test/performance-monitoring/transaction-service.spec.js
@@ -24,7 +24,6 @@
  */
 
 import TransactionService from '../../src/performance-monitoring/transaction-service'
-import Transaction from '../../src/performance-monitoring/transaction'
 import Span from '../../src/performance-monitoring/span'
 import Config from '../../src/common/config-service'
 import LoggingService from '../../src/common/logging-service'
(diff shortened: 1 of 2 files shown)


jahtalab (commented):
The `ensureCurrentTransaction` is used in other places in transaction service. I think the change logic breaks some assumptions there. The assumption is that `ensureCurrentTransaction` either returns current transaction or if there is none it creates one, but the current logic always creates one. But it seems that in those places we need a new transaction anyway. so changing name of this method would be better to represent it always sets the current transactions.

vigneshshanmugam:
I did verify at all the places, we have guards everywhere using `getCurrentTransaction` so I will change the method name to reflect the behaviour if it makes it clear.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -351,14 +357,6 @@ describe('TransactionService', function() {
     })
   })
 
-  it('should ensureCurrentTransaction once per startTransaction', function() {
```
jahtalab, line 354: Would you please revert this removal? we used to have a bug that we called the created transaction multiple times and that would result in loosing spans.

vigneshshanmugam: Done

jahtalab (approved):

PR #890, opened 2020-09-08
feat(rum): add missing methods to types

fralonra:
Add typings for methods: `isActive`, `fetchCentralConfig`, `config`.

Since none of them has underscore at the beginning of their function names, I believed that they ought to be public methods, and should be typed.

### DIFF
--- a/packages/rum/src/index.d.ts
+++ b/packages/rum/src/index.d.ts
@@ -39,7 +39,6 @@ declare module '@elastic/apm-rum' {
      */
     isEnabled(): boolean
     isActive(): boolean
-    fetchCentralConfig(): Promise<AgentConfigOptions | undefined>
     observe(name: TransactionEvents, callback: (tr: Transaction) => void): void
     config(config: AgentConfigOptions): void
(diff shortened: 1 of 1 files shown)


packages/rum/src/index.d.ts
```diff
@@ -38,7 +38,10 @@ declare module '@elastic/apm-rum' {
      * undocumented, might be removed in future versions
      */
     isEnabled(): boolean
+    isActive(): boolean
+    fetchCentralConfig(): Promise<AgentConfigOptions | undefined>
```
jahtalab, line 42: `fetchCentralConfig` is not meant to be used as an API. We should probably indicated that in the method comment but for now would you please remove this line.

jahtalab (changes requested):
Thanks @fralonra , Just one comment!

packages/rum/src/index.d.ts
fralonra: Done

jahtalab (approved):

PR #900, opened 2020-09-22
feat(rum-core): add longtask to experience field

### LINKED ISSUE #895: Investigate solutions for querying longtasks in the RUM dashboard
Currently querying longtaks to provide aggregation in the RUM dashboard includes some expensive ES queries. The query involves getting all `page-load` transactions and then querying all longtaks spans for those transactions.

To help with this the RUM agent can aggregate longtask metrics in the transaction document:
```
{
    experience: {
        longtask: {
            duration: 1,
            max: 1,
            count: 1
        }
    }
}
```

- [x] Request the change from the APM server -> https://github.com/elastic/apm-server/issues/4226

jahtalab:
Fixes #895

### DIFF
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -297,7 +297,7 @@ class TransactionService {
             if (longtask.count > 0) {
               tr.experience.longtask = {
                 count: longtask.count,
-                duration: longtask.duration,
+                sum: longtask.duration,
                 max: longtask.max
               }
             }
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -276,7 +276,7 @@ class TransactionService {
            * and once performance observer is disconnected
            */
           if (tr.captureTimings) {
-            const { cls, fid, tbt } = metrics
+            const { cls, fid, tbt, longtask } = metrics
             if (tbt.duration > 0) {
               tr.spans.push(createTotalBlockingTimeSpan(tbt))
             }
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -293,6 +293,28 @@ class TransactionService {
...
+                longtaskCount++
+                const dur = span.duration()
+                longtaskDuration += dur
+                if (dur > longtaskMax) {
+                  longtaskMax = dur
+                }
+              }
+            }
+            if (longtaskCount > 0) {
+              tr.experience.longtask = {
```
vigneshshanmugam, line 312: These have to be shortened fields. For the new proposals, we tend to follow the shortened formats. So we need something like

```js
tr.experience.lt = {
     d: longtaskDuration, // duration is already in mapping
     ct: longtaskCount,
     lo: longestLongTask
}
```

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -293,6 +293,28 @@ class TransactionService {
             if (fid > 0) {
               tr.experience.fid = fid
             }
+
+            let longtaskCount = 0
+            let longtaskDuration = 0
+            let longtaskMax = 0
+            for (let i = 0; i < tr.spans.length; i++) {
```
vigneshshanmugam, line 300: Even though this works, The approach doesn't feel right to me personally as we are looping over 100's of spans on an average page even for browsers that does not support long tasks.

vigneshshanmugam (changes requested):
IMO, the better approach would be to have this logic as part of `metrics.js`and extend our https://github.com/elastic/apm-agent-rum-js/blob/f8fbb1e65bc1b67bf40c6dfb9c1e455f3b46c061/packages/rum-core/src/performance-monitoring/metrics.js#L36-L44 object and add the longtask timing information like how we are doing for TBT calculation.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: I think we should use the shorthand if the fields have well-defined abbreviations like `cls` for others it seems to me it's a diminishing return.

vigneshshanmugam: Not really,  its not just for optimisation. We have to keep it consistent across the spec. Having all other fields as short-hand and having these ones with full form doesn't make any sense. 

I am leaning towards having them as short hand fields, If you are still not convinced, let's hear it from the server team.

vigneshshanmugam (approved):
Look good 👍

PR #904, opened 2020-09-28
feat: add span and transaction outcome

### LINKED ISSUE #876: Add outcome to transactions and spans
See details - https://github.com/elastic/apm/pull/299

jahtalab:
This PR add outcome to Transactions and Spans. The outcome is only added for `http-request` transactions since based on the definition of outcome we can only infer that for these type of transactions. For all other transactions the outcome is `unknown` which is set by the APM Server. See the original issue for more detail.

Fixes #876

### DIFF
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -308,9 +308,13 @@ export default class PerformanceMonitoring {
           status = target.status
         }
 
-        let outcome = OUTCOME_SUCCESS
-        if (status >= 400 || (status == 0 && data.status != 'abort')) {
-          outcome = OUTCOME_FAILURE
+        let outcome = null
+        if (data.status != 'abort') {
+          if (status >= 400 || status == 0) {
+            outcome = OUTCOME_FAILURE
+          } else {
+            outcome = OUTCOME_SUCCESS
+          }
         }
         span.outcome = outcome
         const tr = transactionService.getCurrentTransaction()
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -309,7 +309,7 @@ export default class PerformanceMonitoring {
         }
 
         let outcome = OUTCOME_SUCCESS
-        if (status >= 400 || status == 0) {
+        if (status >= 400 || (status == 0 && data.status != 'abort')) {
           outcome = OUTCOME_FAILURE
         }
         span.outcome = outcome
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -403,7 +403,8 @@ export default class PerformanceMonitoring {
         sync: span.sync,
         start: parseInt(span._start - transactionStart),
         duration: span.duration(),
-        context: span.context
+        context: span.context,
+        outcome: span.outcome
       }
       return truncateModel(SPAN_MODEL, spanData)
     })
@@ -422,7 +423,8 @@ export default class PerformanceMonitoring {
         started: spans.length
       },
       sampled: transaction.sampled,
-      experience: transaction.experience
+      experience: transaction.experience,
+      outcome: transaction.outcome
     }
     return truncateModel(TRANSACTION_MODEL, transactionData)
   }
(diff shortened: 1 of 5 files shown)


vigneshshanmugam (changes requested):
I am not sure if the PR fixes the original issue, Few concerns
+ Both Transaction and Span model doesn't have `outcome` defined. 
+ We are not sending the captured `outcome` payload for both transaction and spans to the APM server. Can you add a test for this case? 
+ The `outcome` field is not compressed. 
+ Is the reasoning behind setting `outcome` for `HTTP` transactions was due to the fact APM server have `unknown` as default for other transactions? If Yes, Can you add it to the issue description / code comments?

jahtalab:
Thanks @vigneshshanmugam , I probably should've open this as draft. In any case I've made some changes, please have another look.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -298,7 +300,24 @@ export default class PerformanceMonitoring {
...
+        const { span, response, target } = data
+        let status
+        if (response) {
+          status = response.status
+        } else {
+          status = target.status
+        }
+
+        let outcome = OUTCOME_SUCCESS
+        if (status >= 400 || status == 0) {
```
vigneshshanmugam, line 312: It's >= 400 for spans (includes client errors) and >= 500 for transactions (does not include client errors)

We have to account this before setting outcome for transactions.

vigneshshanmugam, line 312: Do we know if the status is 0 for timeouts and aborts as well? I don't remember the values.

vigneshshanmugam (commented):
Thanks @jahtalab Added few comments, Also we might need few test cases

+ one with status code 400, checking different outcome for span and transaction
+ Sending the outcome with valid values to apm-server, may be add test on `apm-server.spec.js`?

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -298,7 +300,24 @@ export default class PerformanceMonitoring {
     } else if (event === INVOKE) {
       const data = task.data
       if (data && data.span) {
-        transactionService.endSpan(data.span, data)
+        const { span, response, target } = data
```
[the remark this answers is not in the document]
jahtalab: I wanted to do a refactoring of our xhr and fetch patch, created an issue for it: https://github.com/elastic/apm-agent-rum-js/issues/907

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -298,7 +300,24 @@ export default class PerformanceMonitoring {
...
+          status = target.status
+        }
+
+        let outcome = OUTCOME_SUCCESS
+        if (status >= 400 || status == 0) {
+          outcome = OUTCOME_FAILURE
+        }
+        span.outcome = outcome
+        const tr = transactionService.getCurrentTransaction()
+        if (tr && tr.type === HTTP_REQUEST_TYPE) {
```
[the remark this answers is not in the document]
jahtalab: The current transaction can actually change before a task is invoked so this ensures we only set it for http requests

packages/rum-core/src/performance-monitoring/performance-monitoring.js
jahtalab: That's only for backend transactions which respond to a request not for outgoing transactions.

jahtalab: That's correct, I've added abort as an exception to failure but timeout should be considered failure.

jahtalab:
@vigneshshanmugam , The different outcome for Transaction in spec is for backend transactions responding to requests not for outgoing http, like in our case. I've added the payload to be sent to apm-server.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
vigneshshanmugam: I had a conversation with Felix offline that we decided on
Timeout -> `failure`
Abort -> `unknown`

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -308,9 +308,13 @@ export default class PerformanceMonitoring {
           status = target.status
         }
 
-        let outcome = OUTCOME_SUCCESS
-        if (status >= 400 || (status == 0 && data.status != 'abort')) {
-          outcome = OUTCOME_FAILURE
+        let outcome = null
```
vigneshshanmugam, line 311: Better to keep it `undefined` so it would be removed in JSON.Stringify? WDYT?

vigneshshanmugam (approved):
Looks good

PR #909, opened 2020-09-29
ci(jenkins): release input parameters

v1v:
## What

- Simplify the release process by filling only the required checkbox (`release`)
- Other input parameters will be ignored when release happens from the master branch.
- Update docs
- Remove unused parameter `parallel_test`

## Why

Every commit in the master branch runs the benchmark and saucelabs stages by default. Therefore, since the release happens from a particular commit in the master branch then it's not required to run those stages anymore.


vigneshshanmugam (approved):
Looks good 👍

jahtalab (approved):
Thanks @v1v

PR #914, opened 2020-10-07
chore: enable stale bot for issues

### LINKED ISSUE #912: Provide a bot to mark stale issues as such and close after some time
This bot would help with triaging issues by adding a stale label to the ones that haven't had any activities for a while and closing after some time.

The bot should only consider this for issues that don't have a label or have the `question` as label.

cc @elastic/observablt-robots

v1v:
## What

Enable [stale bot](https://github.com/probot/stale) to notify and then close the Issues that haven't had any activity for 120 days, a notification will be sent after 100 days and if nothing else happens they will get closed 20 days later.

## How it works

- Support `issues` only as stated
- It uses the `Stalled` label to categorise the ones that have been informed without any activity in 120 days.
- It looks for the issues that got the label `discuss` and `question`.
- Those `Stalled` issues will be closed if no further activity after 20 days.
- Those issues without a label are excluded from the bot. (There is no support for such). The workaround will be as simple as label them with `Stalled` then the bot will close them after 20 days.

### Issues

Closes #912 
Relates to https://github.com/elastic/beats/pull/19157

### Questions

~@cachedout what do we do to enable the botelastic to work with this repo?~ I'll ask infra to enable it

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,19 +1,19 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 
 # Number of days of inactivity before an Issue or Pull Request becomes stale
-daysUntilStale: 335
+daysUntilStale: 100
 
 # Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
 # Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
-daysUntilClose: 30
+daysUntilClose: 20
 
 # Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
-onlyLabels: []
+onlyLabels:
+  - question
+  - discuss
 
 # Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
-exemptLabels:
-- question
-- discuss
+exemptLabels: []
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: true
@@ -29,7 +29,7 @@ staleLabel: Stalled
 
 # Comment to post when marking as stale. Set to `false` to disable
 markComment: >
-  This issue has been automatically marked as stale because it has not had
+  This PR/Issue has been automatically marked as stale because it has not had
   recent activity. It will be closed if no further activity occurs. Thank you
   for your contributions.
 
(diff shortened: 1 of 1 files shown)


mdelapenya (approved):

.github/stale.yml
```diff
@@ -0,0 +1,84 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 335
```
vigneshshanmugam, line 4: This feels really long IMO, how about 3-6 months? Cleanup sooner also ensures we are picking and working on backlogs at some point. just a personal preference though? @jahtalab thoughts?

v1v: Sure thing, I just took what we got in place as a template, happy to change it :)

cachedout (approved):

.github/stale.yml
jahtalab: I agree, we should make a decision about an issue within 3-6 months (possibly even sooner than that but to reduce noise let's set it at 4 month). We either close the issue or add an appropriate label (e.g. `bug`, `enhancement`) to it.

.github/stale.yml
```diff
@@ -0,0 +1,84 @@
...
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 30
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+- question
+- discuss
```
jahtalab, line 16: I think these two sets of labels are should be reversed.

jahtalab (changes requested):
Thanks @v1v, 

I think the condition for the which issues should be considered stale is not correct. We want issues with `question`, `discuss` labels to be considered stale. Also issues with no label should be considered stale as well.

But issues with any other label should be exempt from being stale.

.github/stale.yml
v1v: Aha! I actually read the description in the issue incorrectly! Thanks for the heads up :)

kuisathaverat (approved):

vigneshshanmugam (approved):
LGTM

PR #917, opened 2020-10-11
Link Breakdown Metrics to Configuration Docs

arhum602:


vigneshshanmugam (changes requested):
Thanks for the PR @arhum602 , If you are trying to help with this issuue #917, Its the other way around. 

We have to link the explanation of breakdown metrics which is in `breakdown-metrics.asciidoc` to this `breakdownMetrics` configuration `<configuration.asciidoc>`.

PR #935, opened 2020-11-20
chore(jenkins): do not build test images

mdelapenya:
## What is this PR doing?
It removes the build commands, because we are going to consume the already published images for Puppeteer and Playwright.

Besides that, we are defining a `.node-version` file under the `dev-utils` directory to have one source of truth when building the docker images for tests using Puppeteer and Playwright.

Finally, we are adding to packer cache the test Docker images.

## Why is it important?
We were building the images with every CI build, and with this change, we will speed up the build in around 2-4 minutes. Although the images are cached on CI, we are building the images too because there could be changes in the PR affecting the build process, but reusing the existing layers.

## Other considerations
When updating the Node JS version for the images, it's required to build them from the CI first, as explained in the Dockerfile in a comment:

> When changing Node version, please build the Docker image for RUM here:
>   https://apm-ci.elastic.co/job/apm-shared/job/apm-docker-images-pipeline/build?delay=0sec
> Remember checking the `rum` parameter


package.json
```diff
@@ -52,7 +52,8 @@
           "rum-angular",
           "rum-core",
           "rum-react",
-          "rum-vue"
+          "rum-vue",
+          "testing"
```
mdelapenya, line 56: Please let me know if this new scope is acceptable

cachedout (approved):

v1v (changes requested):
I'd rather prefer to build if there are changes though, in other words, can we get the benefits of building with some cache layouts, can we verify if https://apm-ci.elastic.co/job/apm-shared/job/apm-docker-images-pipeline doesn't remote the docker layouts?

mdelapenya:
> I'd rather prefer to build if there are changes though, in other words, can we get the benefits of building with some cache layouts, can we verify if https://apm-ci.elastic.co/job/apm-shared/job/apm-docker-images-pipeline doesn't remote the docker layouts?

@v1v would reverting 5dbb768 and 053d073 be enough?

v1v:
I think https://github.com/elastic/apm-agent-rum-js/commit/5dbb768b4f78f362e72642873c157556cc9ea903#diff-2b49327f27d20a1006de264510d2b13b76497bdc41d4a2354f346c13308c010dL3 might be required to be changed, since `--force-rm` will avoid the caching

dev-utils/docker-compose.yml
```diff
@@ -125,10 +125,6 @@ services:
           - wait-for-apm-server
 
   node-puppeteer:
-    build:
-      context: ../.ci/docker/node-puppeteer
-      args:
-        - NODEJS_VERSION=12
```
v1v, line 131: will this avoid the building of those docker images?

mdelapenya: That's correct, it was not only the script but the compose file. Submitting a fix now

.ci/packer_cache.sh
```diff
@@ -0,0 +1,14 @@
...
+
+readonly NODEJS_VERSION="12"
+
+DOCKER_IMAGES="docker.elastic.co/observability-ci/node-playwright:${NODEJS_VERSION}
+docker.elastic.co/observability-ci/node-puppeteer:${NODEJS_VERSION}"
+
+if [ -x "$(command -v docker)" ]; then
+  for di in ${DOCKER_IMAGES}
+  do
+  (retry 2 docker pull "${di}") || echo "Error pulling ${di} Docker image, we continue"
```
v1v, line 12: is docker build also required to cache the docker images?

.ci/packer_cache.sh
```diff
@@ -0,0 +1,14 @@
+# shellcheck disable=SC1091
+source /usr/local/bin/bash_standard_lib.sh
+
+readonly NODEJS_VERSION="12"
```
v1v, line 4: Is this version defined somewhere else that could be evaluated? Then we don't need to worry about this file if it changes in the future

package.json
vigneshshanmugam: We use `chore` or `test` for the test changes as per the conventional commit suggestion. 

```
types other than fix: and feat: are allowed, for example @commitlint/config-conventional (based on the the Angular convention) recommends build:, chore:, ci:, docs:, style:, refactor:, perf:, test:, and others.
```

vigneshshanmugam (approved):
Lgtm

package.json
mdelapenya: Is it acceptable if I rewrite commit history changing the three commit messages?

vigneshshanmugam: We squash and merge anyway, so feel free to do it when squash merging the PR in the UI itself.

mdelapenya: So let me add a commit on top reverting the addition of the new scope in Husky, thanks!

.ci/packer_cache.sh
mdelapenya: Yeah, we could add it to a static file and reuse it. I like the idea.

.ci/packer_cache.sh
mdelapenya: I haven't seen it in other places: https://github.com/elastic/apm-pipeline-library/blob/master/.ci/packer_cache.sh

package.json
mdelapenya: At the moment I reverted it and pushed, lint failed. I pushed force the branch with the commit messages updated

mdelapenya:
@v1v I added your suggestions for building the images at Packer cache time, so that it will reuse as many layers as possible at PR build time. Thank you!

.ci/packer_cache.sh
```diff
@@ -0,0 +1,16 @@
...
+readonly NODEJS_VERSION="$(cat ./dev-utils/.node-version)"
+
+FLAVOURS="playwright
+puppeteer"
+
+if [ -x "$(command -v docker)" ]; then
+  for flavour in ${FLAVOURS}
+  do
+  local di="docker.elastic.co/observability-ci/node-${flavour}:${NODEJS_VERSION}"
+  docker build --force-rm -t ${di} ./.ci/docker/node-${flavour}
```
v1v, line 13:
```suggestion
  docker build --build-arg NODEJS_VERSION=${NODEJS_VERSION} -t ${di} ./.ci/docker/node-${flavour}
```

.ci/packer_cache.sh
```diff
@@ -0,0 +1,16 @@
...
+  (retry 2 docker pull "${di}") || echo "Error pulling ${di} Docker image, we continue"
```
v1v, line 14:
```suggestion
  local di="docker.elastic.co/observability-ci/node-${flavour}:${NODEJS_VERSION}"
  (retry 2 docker pull "${di}") || echo "Error pulling ${di} Docker image, we continue"
  docker build --build-arg NODEJS_VERSION=${NODEJS_VERSION} -t ${di} ./.ci/docker/node-${flavour}
```
```

File: .ci/scripts/pull_and_build.sh [hunk 7]
```diff
@@ -1,6 +1,12 @@
 #!/usr/bin/env bash
+export NODEJS_VERSION=$(cat ./dev-utils/.node-version)
+
 docker-compose -f ./dev-utils/docker-compose.yml --log-level INFO pull --quiet --ignore-pull-failures
-docker-compose -f ./dev-utils/docker-compose.yml --log-level INFO build --force-rm >docker-compose.log 2>docker-compose.err
+
+# We are building the images here even though the Docker images are already cached in Packer.
+# This is because there could be changes in the PR affecting the files copied to the Docker image,
+# which we want to test in the current build. 
+docker-compose -f ./dev-utils/docker-compose.yml --log-level INFO build >docker-compose.log 2>docker-compose.err
```
v1v, line 9:
```suggestion
NODEJS_VERSION=${NODEJS_VERSION } docker-compose -f ./dev-utils/docker-compose.yml --log-level INFO build >docker-compose.log 2>docker-compose.err
```

v1v (approved):

mdelapenya:
Waiting for the CI to pass to merge. Thanks @vigneshshanmugam and @v1v for the review!

PR #940, opened 2020-12-01
fix: null is not an object (evaluating 'reason.message')

clyfish:
When reason is null, apm sdk crashes on evaluating `reason.message`.


clyfish:
Sorry for introducing the bug from commit 716d168a076334ab2262603312d43fb20f9e93f0.

vigneshshanmugam (commented):
No worries, good catch. Can you add a test case for the same as part of the same test?

clyfish:
done, and the travis ci passed.

vigneshshanmugam (approved):
Thanks @clyfish Will merge once our CI passes.

PR #941, opened 2020-12-01
chore(jenkins): fix running issues with the packer-cache

v1v:
### What

Fix an issue with the local prefix that's only supported for functions

### Why

Otherwise the packer cache automation won't cache those docker images

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -9,8 +9,8 @@ puppeteer"
 if [ -x "$(command -v docker)" ]; then
   for flavour in ${FLAVOURS}
   do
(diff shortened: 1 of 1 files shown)


mdelapenya:
Nice catch! Do you have a link with the error to take a look 👀 ?

mdelapenya (approved):

kuisathaverat (approved):

v1v:
jenkins run the tests please

v1v:
Unrelated errors since this is not executed in the build but as part of the CI Worker VM generation process that runs in the infra side

PR #948, opened 2020-12-18
feat: first draft for load testing APM Server

jahtalab:
This is the first draft for RUM benchmarking with sessions. The idea is to generate a diverse payloads to send to the APM server and records the servers' metrics at the end.


scripts/spike.js
```diff
@@ -0,0 +1,228 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+const faker = require('faker')
+const { merge } = require('lodash')
```
vigneshshanmugam, line 27: we can just use https://www.npmjs.com/package/lodash.merge instead of pulling in a big package.

scripts/spike.js
```diff
@@ -0,0 +1,228 @@
...
+    },
+    span_count: {
+      started: 82
+    },
+    sampled: true,
+    experience: {
+      tbt: 100,
+      cls: 0.5,
+      fid: 50
+    }
```
vigneshshanmugam, line 109: include breakdown metrics as well for transactions and spans?

scripts/spike.js
```diff
@@ -0,0 +1,228 @@
...
+  })
+  return response.text()
+}
+
+async function generatePayloads(trCount) {
+  let promises = []
+  for (let i = 0; i < trCount; i++) {
+    let payload = generateTransaction(50)
+    payload.unshift(merge({}, defaultMeta))
+    let p = postPayload('http://localhost:8200/intake/v2/rum/events', payload)
```
vigneshshanmugam, line 220: we can use our test config setup to get the server url https://github.com/elastic/apm-agent-rum-js/blob/e4e10d5b7c451a950808d605a2db99cd275d20eb/dev-utils/test-config.js#L45

scripts/spike.js
```diff
@@ -0,0 +1,228 @@
...
+    promises.push(p)
+  }
+  await Promise.all(promises)
+  let response = await fetch('http://localhost:8200/debug/vars')
```
vigneshshanmugam, line 224: wow, Is this a new endpoint?

vigneshshanmugam (commented):
@jahtalab PR looks good, but feels like we can do a lot more with a loading test framework like this [autocannon](https://github.com/mcollina/autocannon) where we can control the requests and burst frequency of them without writing our own. 

Couple of points
+ As we support compressed payloads,its worth gzipping the payload before sending it to the server. Required as server spends time decoding the same. 
+ Instead of local endpoint, can we create a dummy APM server in cloud to act as a load testing endpoint for us

scripts/spike.js
jahtalab: I don't think it makes a big different in our use-case here

scripts/spike.js
jahtalab: Added the transaction breakdown, the span break downs currently would add one more document since we currently have one type in this script

scripts/spike.js
jahtalab: It's not new but it's the same endpoint hey-apm uses.

jahtalab:
Thanks @vigneshshanmugam, I think using something like  autocannon is a good option but we still need to generate the payload which is the major part of the effort here. For now I prefer to keep the transaction generation logic simple, we can add more complex logic later if needed.

Re. the cloud instance for apm server, I don't think that's a good idea, since it adds unpredictability to our load test, so I'd prefer to use the local instance here.

vigneshshanmugam:
@jahtalab I would say the payload is the easy part and we need requests at different intervals and control the burst frequencies which would let us test the server capabilities in a better way. Lets add that later on. 

Regarding cloud server, May be its worth checking how `hey-apm` benchmarks are run and we can do the same approach when we set it up. 

What about the compression for payload, do you want to do it in follow up PR?

scripts/apm-server-load-test.js
```diff
@@ -0,0 +1,267 @@
...
+
+;(async function() {
+  const outputFile = process.argv[2]
+
+  const result = await generatePayloads(10)
+
+  if (outputFile) {
+    const outputPath = join(__dirname, '../', outputFile)
+    let ndJSONOutput =
+      '{"index": { "_index": "benchmarks-rum-load-test", "_type": "_doc"}}' +
```
vigneshshanmugam, line 260: need `timestamp` field as well for ES docs.

jahtalab:
@vigneshshanmugam, the main functionality in this PR is to add the fake payloads. A simple algorithm for the burst and frequency that can be implemented in a few lines of code is enough for this spike, no need to add more complexity at this point.

Re. compression of payload, let's do that separately.

vigneshshanmugam (approved):
Looks good

jahtalab:
The test failures are unrelated to the changes here.

PR #951, opened 2021-01-06
fix: click transactions supports window and document target

clyfish:
React 16 binds click events on `document`, but `shouldInstrumentEvent` only supports Element target.


packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
```diff
@@ -719,6 +719,46 @@ describe('PerformanceMonitoring', function() {
...
+      const listener = e => {
+        expect(e.type).toBe('click')
+      }
+
+      window.addEventListener('click', listener)
+      document.body.click()
+
+      let tr = transactionService.getCurrentTransaction()
+      expect(tr).toBeDefined()
+      expect(tr.name).toBe('Click - html')
```
vigneshshanmugam, line 738: shouldn't this be `BODY`?

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -186,14 +186,21 @@ export default class PerformanceMonitoring {
         task.eventType === 'click'
       ) {
         const target = task.target
-        const name = target.getAttribute('name')
+        const name = target.getAttribute ? target.getAttribute('name') : ''
```
vigneshshanmugam, line 189: nit: typeof target.getAttribute === "function"

vigneshshanmugam (commented):
Good catch and thanks for the PR @clyfish Added some comments.

packages/rum-core/test/performance-monitoring/performance-monitoring.spec.js
clyfish: @vigneshshanmugam `HTML` is parentNode of `BODY`, so I think `HTML` is better than `BODY` when target is `window` or `document`.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -186,14 +186,24 @@ export default class PerformanceMonitoring {
         task.eventType === 'click'
       ) {
         const target = task.target
-        const name = target.getAttribute('name')
+        const name =
+          typeof target.getAttribute === 'function'
+            ? target.getAttribute('name')
+            : ''
```
reecebenson, line 192: This logic here should also be applied to a line further down in this file, [click here (Line 224)](https://github.com/elastic/apm-agent-rum-js/blob/cd722fefc10dde2c29626144083e7b7a82477a97/packages/rum-core/src/performance-monitoring/performance-monitoring.js#L224). Otherwise we'll see a `TypeError: target.getAttribute is not a function` when the function does not exist.

reecebenson (changes requested):

packages/rum-core/src/performance-monitoring/performance-monitoring.js
clyfish: You are right, `document` and `window` have no `getAttribute`.

clyfish: @vigneshshanmugam Should we change `typeof xxx === 'function'` to `isFunction(xxx)` to save some bytes?
```bash
$ ag -c "'function'"
common/config-service.js:1
common/patching/event-target-patch.js:1
common/patching/history-patch.js:1
common/patching/patch-utils.js:1
common/logging-service.js:1
common/service-factory.js:1
common/event-handler.js:1
common/utils.js:10
common/compress.js:1
performance-monitoring/performance-monitoring.js:2
performance-monitoring/span-base.js:1
common/throttle.js:1
performance-monitoring/transaction-service.js:1
error-logging/error-logging.js:3
```

vigneshshanmugam: Hmmm might not give any advantage when its gzipped, did you verify?

clyfish: Before gzipped, saved 353 bytes: 57792-57439
After gzipped, saved only 5 bytes: 18808-18803

packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -42,7 +42,7 @@ for (let i = 0; i < eventTypes.length; i++) {
 
 function shouldInstrumentEvent(target, eventType, listenerFn) {
   return (
-    target instanceof Element &&
+    (target instanceof Element || target === window || target === document) &&
```
vigneshshanmugam, line 45: this would effectively mean we start capturing all click events everywhere on the page which are using event bubbling (Ex: react). As a side effect we would get everything as `Click - document` which is not meaningful. One way to fix this problem is to use the `target.activeElement` if available on the taret. https://developer.mozilla.org/en-US/docs/Web/API/Document/activeElement

This would work for all of the cases mentioned here - https://github.com/elastic/apm-agent-rum-js/issues/969#issue-812096856

clyfish: But React 17 Listen on `div#root`.
https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation

clyfish: My suggestion:

```javascript
if (target === window || target === document ||
    target && isFunction(target.contains) && target.contains(document.activeElement)) {
  // use document.activeElement instead of target
}
```

https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

clyfish: Or we can always use `document.activeElement` instead of `target`.

Anyway, it's a breaking change that using `document.activeElement` when target is not `document` or `window`.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -188,6 +188,14 @@ export default class PerformanceMonitoring {
         task.eventType === 'click'
       ) {
         const target = task.target
```
reecebenson, line 190: Getting the follow stacktrace:
```
TypeError: invalid assignment to const 'target'
    getEventTargetSub performance-monitoring.js:197
    sendOnly event-handler.js:35
    sendOnly event-handler.js:33
    send event-handler.js:45
    patchAll index.js:23
    ...
```

Preferably `let target = task.target` here so [line 197](https://github.com/elastic/apm-agent-rum-js/pull/951/files#diff-e3d6e2c6dedbb088511627f63de9cfdfed996ea94849ccf78f155b171cb33a5eR197) can update the var?

reecebenson (changes requested):
Looks good, just one bug that should be simple enough to address. I'm not seeing a consistent update here, and still not receiving `user-interaction` updates for most button clicks (i.e. from #969, I can't seem to get the `applyFiltersBtn` to be marked as a user-interaction, as it doesn't even seem to be getting classed as clicked). Will continue testing and feedback here.

packages/rum-core/src/common/patching/event-target-patch.js
vigneshshanmugam: Good point about the react 17 breaking change. So instead of touching the code here, Let's keep it as it is and check for target being on the `document`. Instead check when the event is actually called https://github.com/elastic/apm-agent-rum-js/blob/8dda5c57a55f815f3252f7e2e769191efeefd7fe/packages/rum-core/src/performance-monitoring/performance-monitoring.js#L190

Check if the target contains any `activeElement`. 

If you can do a test on that would be great. I will take your PR for a spin today and test it out.

packages/rum-core/src/performance-monitoring/performance-monitoring.js
clyfish: Fixed, and added the eslint rule.

https://eslint.org/docs/rules/no-const-assign

packages/rum-core/src/common/patching/event-target-patch.js
clyfish: I've tested on React 17, and it works.

```
[Elastic APM] startTransaction(43fbc8c4fe6af7f8, Click - button["button"], user-interaction)
```

clyfish:
If we can use `document.activeElement` to find a target, there is no need to patch `addEventListener`. We can directly listen on `window` for click events.

```javascript
const apm = initApm({
  serverUrl,
  serviceName,
  serviceVersion,
  environment,
  active,
  disableInstrumentations: ['eventtarget'],
})

window.addEventListener(
  'click',
  () => {
    const target = document.activeElement

    if (!target || target === document.body || target === document.documentElement) {
      return
    }

    const name = target.getAttribute('name')

    let additionalInfo = ''
    if (name) {
      additionalInfo = `["${name}"]`
    }

    const tagName = target.tagName.toLowerCase()

    const tr = apm.startTransaction(`Click - ${tagName}${additionalInfo}`, 'user-interaction', {
      managed: true,
      canReuse: true,
      reuseThreshold: 100,
    })

    if (tr) {
      tr.addContext({ custom: { selector: getSelectorByElement(target) } })
    }
  },
  true,
)
```

PR #958, opened 2021-02-02
chore: update @babel/cli and bundlesize

jahtalab:


package.json
```diff
@@ -73,7 +73,7 @@
     "@angular/platform-browser": "^8.2.2",
     "@angular/platform-browser-dynamic": "^8.2.2",
     "@angular/router": "^8.2.2",
-    "@babel/cli": "^7.7.5",
+    "@babel/cli": "^7.12.10",
```
vigneshshanmugam, line 76: this version would require babel/core to be also in same version. Aren't you getting warnings or errors during install?

jahtalab: No, I actually don't see any warnings while installing `@babel/cli`

vigneshshanmugam: tested locally, looks good now. Probably a version issue for me before.

vigneshshanmugam (approved):

PR #960, opened 2021-02-04
ci: remove travis references

v1v:
### What

Travis is not anymore supported in our end

### Actions

- [x] Remove webhook
- [x] Disable travis check in case it's mandatory


vigneshshanmugam (approved):
Looks good

PR #965, opened 2021-02-09
feat(rum-core): capture buffered longtask entries

### LINKED ISSUE #964: observe for buffered long tasks on the page
+ Starting from chrome 81, the buffered flag for long tasks has enabled by default - https://www.chromestatus.com/feature/6306761277440000

vigneshshanmugam:
+ fix #964

### DIFF
--- a/packages/rum-angular/test/e2e/with-router/app.e2e-spec.js
+++ b/packages/rum-angular/test/e2e/with-router/app.e2e-spec.js
@@ -60,9 +60,10 @@ describe('Angular router integration', function() {
     const routeTransaction = transactions[1]
     expect(routeTransaction.name).toBe('/contacts')
     expect(routeTransaction.type).toBe('route-change')
-    expect(routeTransaction.spans.length).toBe(1)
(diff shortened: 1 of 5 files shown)


vigneshshanmugam:
Interesting, the commits from base branch is not getting picked up. GH is down.

jahtalab (commented):
Thanks @vigneshshanmugam , just a few questions, I assume the issue with throwing an error when `buffered` flag is used, still exist for older browsers, right? Also what about any other browsers?

I think we might be missing the failures in browser tests because we catch and ignore the error. It might be worthwhile to see if we remove the catch part, how many of our browsers/tests break? (to be clear, this is only for testing and not for the final version)

vigneshshanmugam:
@jahtalab Not passing the buffered flag for long tasks was to avoid the warning sign in chrome as chrome was issuing warnings when long tasks are marked with `buffered: true`.

Oh that error handling was done for Safari mainly. We still don't know how many others would break and its a way of feature detecting which is recommended for Performance Observers.

Issue #962, opened 2021-02-08
Add support for Angular 9, 10 and 11


paulb-elastic:
Angular 8 is deprecated (up to 8.2.2 is supported by the current RUM Agent), in favour of:
* Angular 9 - released Feb 2020, LTS ends Aug 2021
* Angular 10 - released Jun 2020, LTS ends Dec 2021
* Angular 11 - released Nov 2020, LTS ends May 2022

ACs:
* Update the RUM Agent to work with Angular versions 9, 10 and 11

### Sub Tasks
+ [x] Add support for Angular 9
+ [x] Move the build system to `ng` to add support for Angular Package Format and AOT compiler. Issues - elastic/apm-agent-rum-js#757 elastic/apm-agent-rum-js#463 elastic/apm-agent-rum-js#693 elastic/apm-agent-rum-js#884
+ [x] Provide typings for the `elastic/angular` package. As part of the move to `ng`, this would be automatic. 
+ [x] Move the E2E test to the `ng` build system and use the new library once we land support for angular 9. Issues - elastic/apm-agent-rum-js#450 
+ [x] Update docs with the new angular package - #1000 
+ [ ] Release a new major version for Angular 9

Issue #884, opened 2020-09-02
Could not find a declaration file for module '@elastic/apm-rum-angular'


super-kamil:
1. Follow the steps from https://www.elastic.co/guide/en/apm/agent/rum-js/current/angular-integration.html
2. Run `ng serve`
3. Error: `Could not find a declaration file for module '@elastic/apm-rum-angular'`

mezzo1981:
Any update on this? We get also this error.

vitormsilva:
I added this to my project under the `./src/@types` folder and then reference on the typeRoots inside of the `tsconfig.json` 

```typescript
// elastic__apm-rum-angular.d.ts
declare module '@elastic/apm-rum-angular' {
    import { Router } from '@angular/router';
    import { ErrorHandler } from '@angular/core';

    export interface Config {
        serviceName: string;
        serverUrl: string;
    }

    export class ApmService {
        router: Router;
        static apm: any;
        constructor(router: Router);
        init(config: Config): any;
        observe(): void;
    }

    export class ApmErrorHandler extends ErrorHandler {
        static apm: any;
        handleError(error: any): void;
    }
}
```

blank-27:
can I work on this issue ?

victorpintot8:
Same error here, I installed the 1.1.9 version of the package.

mitschmidt:
Had the same issue, @victorpintot8's solution only worked for me after removing the import statements.

ivayloc:
Same here

JayeshPawar13:
Any update on this?

vigneshshanmugam:
We are working on the support for Angular > 9 version which would include fixing the typings as well. You can track it here https://github.com/elastic/apm-agent-rum-js/issues/962

vigneshshanmugam:
We published the RUM Angular integration v2 Alpha which includes support for Angular 9 and also includes type deps. Would you be able to test it and let us know if it fixes your problem. 

```
npm install @elastic/apm-rum-angular@alpha
```

You have to make small changes to the initialization code. Docs here - https://github.com/elastic/apm-agent-rum-js/blob/cc4aba03d7f843e40f7256ee5a73fb74d4c18c9a/docs/angular-integration.asciidoc

Once these issues are addressed, will go ahead do the proper release. Thanks for the help in advance!

vigneshshanmugam:
Version 2.x.x major has been published with type declaration files. Please do let us know if you face any issues.

PR #746, opened 2020-04-09
wip adding angular project using ng cli

chriscarpenter12:
This is a WIP updating the angular package to use the `ng` cli to build and test the project. There was a recent update adding `es2015` build, but it doesn't fully follow the [APF](https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs). Specifically it doesn't provide a `.metadata.json` in the build for use in AOT builds. By using the `ng` cli it provides everything an angular library needs. This also has the added benefit of setting up an actual angular app to test the project with.

Lerna has some added complexity wrapping an angular project which can also contain multiple apps/libraries :smiley: 

Additionally this has some changes to the Angular library code. The current static `apm` properties on the service and error classes cannot exist on an angular build. I refactored by creating an new InjectionToken used to store the apm object from the base library and then inject that into the service and error class. This also has the benefit of behaving similar to other libraries. Since this library provides the service with `providedIn: 'root'` the user should never have to create a [provider](https://www.elastic.co/guide/en/apm/agent/rum-js/master/angular-integration.html#_instrumenting_your_angular_application) in their module and supply the dependencies manually. Now they import the new `ApmModule` and can use Angular's dependency injection as normal.

### Things that still need done

- [ ]  Integrate shared Karma config
  The default Karma config from Angular works and all tests pass. I struggled trying to integrate the shared Karma config with the extra Webpack options.

- [ ]  E2E tests in an actual Angular app
  I bootstrapped a client app to consume the library and run full E2E tests. The initial issue off the bat is that by default Angular uses protractor out of the box, and a lot of the helper functions share use webdriver. There is benefit testing with protractor because it would be what users consuming the library would use.

- [ ]  Package dependency location
  Currently the angular dependencies are in the leaf package. Should they be hoisted?

- [ ]  CI/CD changes?

- [ ]  Publishing
  I modified `rum-angular` to publish the `dist/rum-angular` folder as that is common for Angular libraries instead of `files` in the package.json.

### Steps to reproduce
```
npx lerna bootstrap
cd packages/rum-angular
npm i
cd ../..
npx lerna link

npx lerna run build --scope=@elastic/apm-rum-angular
lerna notice cli v3.19.0
lerna info versioning independent
lerna notice filter including "@elastic/apm-rum-angular"
lerna info filter [ '@elastic/apm-rum-angular' ]
lerna info Executing command in 1 package: "npm run build"
[...]

Issue #693, opened 2020-03-09
RUM-Agent slows down startup since Angular 9


Flaxline:
Following situation:
There's an Angular Application with some Protractor e2e-Tests, whose execution is monitored by apm-rum (with a custom implementation of route-monitoring).
These tests and their monitoring worked fine with Angular 8.x and apm-rum agent 4.8.

Since a recent Update to Angular 9 (and later also to apm-agent 4.9), an active monitoring slows down the application to a point where the Protractor Tests fail due to script timeouts. (Does not happen when monitoring is inactive, and did not happen before the updates) - which is now worked around with increased timeouts. 

But the data on RUM server still show very similar durations for page-load transactions, and do not at all reflect any changes in a slower start of the application.

What could case the Application to have such a noticably slower startup when monitoring is enabled, than before?
Thanks for your help!

vigneshshanmugam:
@Flaxline Thanks for bringing that to the attention. We don't have a test suite at the moment to confirm it. Do you have a test application/repository where we can have a look?

Flaxline:
@vigneshshanmugam 
First, thank you for answering and sorry for the delay.
As I can't share the actual code where the issue came up, I tried to recreate the situation in a rather stripped setup: https://github.com/Flaxline/monitoring-slowdown-issue

During creation, I noticed that the issue didn't occur until I added the custom route-monitoring to the project. But as this was already used in the actual application with Angular 8, it still shouldn't be an issue?

So the example does things very similar to the application mentioned in the issue itself. 
It also throws the same timeout error:

> 1) workspace-project App should display welcome message
>  - Failed: script timeout
    (Session info: chrome=80.0.3987.132)
    (Driver info: chromedriver=79.0.3945.16 [...]

The error that may be thrown afterwards, and in the configuration with increased timeouts, is not a problem in the actual app as I can use an existing APM-setup there:

> 4)
>   - Expected [ Entry({ level: WARNING, message: 'http://localhost:4200/vendor.js 64987:23 "[Elastic APM] Failed sending transactions!" Error: http://localhost:8200/intake/v2/rum/events HTTP status: 0
>       at ApmServer._constructError (http://localhost:4200/vendor.js:64041:12)

As stated above, we now work around the issue with increased timeouts, but it'd be nice to know why there's a slowdown since Angular 9.

vigneshshanmugam:
[...]

PR #971, opened 2021-02-25
BREAKING CHANGE(rum-angular): add support for angular 9

### LINKED ISSUE #757: Anguar Package Format
I had opened a [PR which was closed](https://github.com/elastic/apm-agent-rum-js/pull/746) to address multiple issues with the Angular library namely using the `ng` cli to create a package following the [Angular Package Format](https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview). The `ng` cli creates an output that contains a `*.metadata.json` file which Angular uses for AOT compilation which is currently missing. It was a work in progress. It was a start at adding an actual Angular structure to the `rum-angular` lerna package. 

It was also meant to address multiple other issues. Like proper [e2e tests](https://github.com/elastic/apm-agent-rum-js/issues/450), and [protractor timeout issues](https://github.com/elastic/apm-agent-rum-js/issues/693) which you can't test because you don't have an Angular app defined. The few apps we've tested adding the APM Angular library cause all our Protractor tests to timeout.

As well as fixing improper service usage/declaration. The current documentation has users define the `ApmService` in the app module providers with a router dependency. This is not correct for an Angular library. While it works you should not have to define it like this since you're using `providedIn: 'root'` in the `Injectable`. Which lead to odd [issues like this](https://github.com/elastic/apm-agent-rum-js/issues/463).

This:
```
@NgModule({
  ...
  providers: [{ provide: ApmService, useClass: ApmService,deps: [Router]}]
  ...
})
export class AppModule {
  constructor(@Inject(ApmService) service: ApmService) {
    // API is exposed through this apm instance
    const apm = service.init({...})
  }
}
```
becomes this
```
@NgModule({
  ...
  imports: [ApmModule]
  ...
})
export class AppModule {
  constructor(service: ApmService) {
    // API is exposed through this apm instance
    const apm = service.init({...})
  }
}
```

I defined a task list of outstanding things that needed addressed, but it was a start to fix these things.

vigneshshanmugam:
+ fix #757, #884, #463, #693
+ part of #962 
+ Extracted some of the work done by @chriscarpenter12 on PR #746.. Huge thanks for the help in starting it. 
+ Supports Angular Packing format and AOT compilation via `metadata.json` as we get those for free using the ng toolschain. 
+ We now expose 1 module, 1 service and 1 Error handler from the `@elastic/apm-rum-angular` package. 
    + `ApmModule`
    + `ApmService`
    + `ApmErrorHandler`
+ We use `ng` ecosystem which is the recommended way to build, test and publish the angular libraries and application. 
    + `ng-packgr` is used for building the library - check `ng-package.json`
    + Each task build, test is specified using `angular.json` which has specific `tsconfig` respectively. 
  
**Note to Reviewer: Review commit by commit for better understanding**

Tasks to be done in separate PR
+ [ ] E2E migration / tests will be done in separate PR to reduce the number of changes in a single PR - #974 
+ [ ] Moving to angular 10, 11 etc. 
+ [ ] Updating documentation and breaking changes strategy

These tasks are already added to the main issue #962 to keep changes minimal and moving faster with iteration.


vigneshshanmugam:
@chriscarpenter12 If you have some time, can you have a look at the PR.

chriscarpenter12:
Is the plan to get support out for Angular 9 now? Current version is 11 with 12 coming out soon.

packages/rum-angular/src/apm.module.ts
```diff
@@ -23,12 +23,14 @@
...
+import { apm, ApmBase } from '@elastic/apm-rum'
 
-module.exports = function(api) {
-  /**
-   * Reads BABEL_ENV to decide between CJS and ESM formats
-   */
-  const env = api.env()
-  return getBabelConfig(env, PACKAGE_TYPES.ANGULAR)
-}
+export const APM = new InjectionToken<ApmBase>('APM Base Client')
```
chriscarpenter12, line 30: Later versions of angular the builds were complaining about this being in the `.module.ts` file. In our package I pulled this injection token out to it's own file `apm.model.ts`. YMMV, but shouldn't be an issue for 9 though.

packages/rum-angular/src/apm.service.ts
```diff
@@ -24,19 +24,19 @@
...
 })
 export class ApmService {
-  static apm: any
-
-  constructor(public router: Router) {}
+  constructor(@Inject(APM) public apm: ApmBase, public router: Router) {}
 
   init(config) {
-    const apmInstance = ApmService.apm.init(config)
+    const apmInstance = this.apm.init(config)
```
chriscarpenter12, line 39: Issues I had regarding Protractor timeouts I traced back to here. When the `init` method is called `zone.js` was triggering change detection on these long running requests. I move this call out of zone and Protractor timeouts went away. Like this:

```
let apmInstance;

this.zone.runOutsideAngular(() => {
  apmInstance = this.apm.init(config);
});
```

And add zone to your constructor dependencies: `private readonly zone: NgZone`

packages/rum-angular/src/apm.service.ts
```diff
@@ -24,19 +24,19 @@
...
+import { APM } from './apm.module'
 
 @Injectable({
   providedIn: 'root'
 })
 export class ApmService {
-  static apm: any
-
-  constructor(public router: Router) {}
+  constructor(@Inject(APM) public apm: ApmBase, public router: Router) {}
```
chriscarpenter12, line 36: I marked router private in mine since it's not needed from clients calling this service. `private readonly router: Router`

chriscarpenter12 (commented):
Otherwise looks good. Thanks @vigneshshanmugam !

chriscarpenter12:
What will be the process for upgrading apps? Because devs will need to remove the ApmService in their AppModule providers?

vigneshshanmugam:
> Is the plan to get support out for Angular 9 now? Current version is 11 with 12 coming out soon.

No, I am just doing it for 9 now to keep the changes minimal. We will most likely skip version 9 and go straight to 11 as its the current active version https://angular.io/guide/releases#support-policy-and-schedule. Will do a follow up PR

packages/rum-angular/src/apm.service.ts
vigneshshanmugam: Good point. It doesnt need to be public.

packages/rum-angular/src/apm.service.ts
vigneshshanmugam: Makes sense though on the timeout issue. Seems like the init is dependent on the UI being updated. We dont need the RUM agent to wait for any UI updates. I will move it out.

packages/rum-angular/src/apm.module.ts
vigneshshanmugam: Going to keep it as it is, will update if required.

packages/rum-angular/package.json
```diff
@@ -4,46 +4,35 @@
...
-    "test:unit": "npm run script runUnitTests packages/rum-angular",
-    "test:e2e": "npm run script runE2eTests packages/rum-angular/wdio.conf.js",
-    "test:sauce": "npm run script runSauceTests packages/rum-angular true test:unit build:e2e test:e2e",
-    "test": "npm run test:sauce",
-    "script": "node ../../dev-utils/run-script.js",
-    "karma:dev": "karma start --auto-watch --restartOnFileChange",
-    "karma:coverage": "karma start --coverage --singleRun"
+    "build": "ng build --prod",
+    "test:unit": "ng test",
+    "test": "MODE=none npm run test:unit",
```
jahtalab, line 22: What is currently the issue with running in saucelabs?

packages/rum-angular/src/apm.service.ts
```diff
@@ -24,19 +24,25 @@
...
+  constructor(
+    @Inject(APM) public apm: ApmBase,
+    private router: Router,
+    private readonly ngZone: NgZone
+  ) {}
 
   init(config) {
-    const apmInstance = ApmService.apm.init(config)
+    const apmInstance = this.ngZone.runOutsideAngular(() =>
+      this.apm.init(config)
```
jahtalab, line 44: Why does the init has to be out side the zone?

packages/rum-angular/test/polyfills.ts
```diff
@@ -28,30 +28,6 @@
...
-import 'core-js/features/regexp'
-import 'core-js/features/map'
-import 'core-js/features/weak-map'
-import 'core-js/features/set'
-import 'core-js/features/reflect'
-
-/*
- * Evergreen browsers require these
- */
-import 'core-js/proposals/reflect-metadata'
```
jahtalab, line 53: Is the `reflect-metadata` no longer needed? 
Is type metadata stored in a different way?

package.json
```diff
@@ -161,6 +168,8 @@
     "serve-index": "^1.9.1",
     "stats-lite": "^2.2.0",
     "terser-webpack-plugin": "^3.1.0",
+    "ts-node": "^8.3.0",
+    "tslib": "^1.10.0",
```
jahtalab, line 172: Is `tslib` needed here since it's also added to the angular package.json?

jahtalab (commented):
Thanks @vigneshshanmugam ,  A few points:
- It seems that this PR breaks the support for version 8, which is fine but we should release this in a breaking change (major) version.
- I think the saucelabs test would give us a better assessment of what might break, so I would prefer if we enable that before merging. We can migrate e2e tests in a separate PR.

vigneshshanmugam:
@jahtalab 

> It seems that this PR breaks the support for version 8, which is fine but we should release this in a breaking change (major) version.

Yes thats right. We will release a new major for the angular package alone. 

> I think the saucelabs test would give us a better assessment of what might break, so I would prefer if we enable that before merging. We can migrate e2e tests in a separate PR.

We have to change the whole build pipeline for that to even go to that step. We have to convert the `test/e2e/app` in to an angular application. Let's do it in a separate PR, it will make things easier for review.

packages/rum-angular/test/polyfills.ts
vigneshshanmugam: the angular `ng` build system adds their own polyfills based on `browserlist` in package.json. I will update this with correct list in the upcoming E2E PR.

packages/rum-angular/src/apm.service.ts
vigneshshanmugam: Check the discussion here -  https://github.com/elastic/apm-agent-rum-js/pull/971#discussion_r583328989

packages/rum-angular/package.json
vigneshshanmugam: basically the build system has to migrated from `babel` -> `ng`. 

The test framework needs to be updated to https://www.protractortest.org which is the recommended way of testing Angular projects. Will be done in separate PR.

packages/rum-angular/package.json
```diff
@@ -4,46 +4,35 @@
...
+    "directory": "dist"
   },
   "repository": {
     "type": "git",
     "url": "https://github.com/elastic/apm-agent-rum-js.git",
     "directory": "packages/rum-angular"
   },
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=12.0.0"
```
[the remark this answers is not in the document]
vigneshshanmugam: 8,10 is deprecated. As we are doing a major for angular package. its good time to update this one.

packages/rum-angular/package.json
```diff
@@ -4,46 +4,35 @@
...
   "homepage": "https://www.elastic.co/guide/en/apm/agent/rum-js/current/index.html",
   "license": "MIT",
-  "main": "dist/lib/index.js",
-  "module": "dist/es/index.js",
-  "es2015": "dist/es2015/index.js",
-  "source": "src/index.ts",
-  "files": [
-    "dist",
-    "src"
-  ],
```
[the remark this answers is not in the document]
vigneshshanmugam: check `publishConfig.directory` in `package.json`. we can still keep `files.dist` not the `src`

package.json
vigneshshanmugam: We don't need this for the library. but needed anyway for E2E app. lets keep it as it is since we eventually have to add it back.

vigneshshanmugam:
@jahtalab If there is no major concern on the this PR, I am going to go ahead and merge and address some of the concerns on the next E2E test PR which would make things easier to understand and answer some of your questions as well. Makes my iteration faster.

jahtalab:
@vigneshshanmugam , I would like to run this locally but without the tests it's a bit hard to verify the code actually does what it's supposed to do. 

Would it help you if we make an intermediary branch to merge this into?

jahtalab:
I've created `feat-angular` branch, feel free to merge into that one, I will also checkout that branch to run your changes locally. This way if there's a need for releasing from main branch we can still do it without worrying about untested features.

vigneshshanmugam:
Had a offline discussion, we will merge all Angular related changes in to the `feat-angular` branch and do a test publish run then do a final merge in master before releasing the major.

PR #973, opened 2021-03-04
feat(rum-core): support regex same origin

baruchiro:
close #943 

I didn't test locally because of #972 and other things related to initialize dev env.

PR #977, opened 2021-03-09
docs(rum-core): distributedTracingOrigins regex

baruchiro:
Related to #973

I'm not sure I wrote it right. And my English not always good.
Please suggest me changes.


baruchiro:
Closed, moved to #973

PR #989, opened 2021-03-16
fix(rum): discard longtask spans from prev transaction

### LINKED ISSUE #988: [bug] longtasks from previous transaction are added to next spans
+ With enabling the `buffered` flag for long tasks, the observer entries from last transaction (Eg: `page-load`) is added to the next subsequent transactions (`route-change`). 
+ This bug would have existed even before, but was able to find only after enabling the longtaks as other transactions does not have `experience` field which was only recorded for `page-load` transaction.

vigneshshanmugam:
+ fix #988 
+ We filter all of the longtasks for the page load transaction that happened even before it to make sure we calculate all of the buffered long tasks as well starting from 0.
+ For all subsequent route-change transactions, we only account for long tasks that happened after the transaction start time. This ensures that we are not including all previous buffered long task entries. 

**As a result all of the managed transactions gets adjusted to the start time of the first span, Now all of the subsequent transaction timings would be wrong and broken in the UI**

### DIFF
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -287,8 +287,7 @@ export class PerfEntryRecorder {
   constructor(callback) {
     this.po = {
       observe: noop,
-      disconnect: noop,
-      takeRecords: noop
+      disconnect: noop
     }
     if (window.PerformanceObserver) {
       this.po = new PerformanceObserver(callback)
@@ -315,11 +314,6 @@ export class PerfEntryRecorder {
   }
 
   stop() {
-    /**
-     * Empties out the performance entries from the
-     * current observer
-     */
-    this.po.takeRecords()
     this.po.disconnect()
   }
 }
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -67,9 +67,7 @@ export function createLongTaskSpans(longtasks, agg) {
     const span = new Span(`Longtask(${name})`, LONG_TASK, { startTime })
     agg.count++
     agg.duration += duration
-    if (duration > agg.max) {
-      agg.max = duration
-    }
+    agg.max = Math.max(duration, agg.max)
 
     /**
      * use attribution data to figure out the culprits of the longtask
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -314,6 +315,11 @@ export class PerfEntryRecorder {
   }
 
   stop() {
+    /**
+     * Empties out the performance entries from the
+     * current observer
+     */
+    this.po.takeRecords()
```
vigneshshanmugam, line 322: Just a defensive mechanism to clear out the entries if we don't have any listeners taking care of the records.

vigneshshanmugam: Not required, tested it with different scenarios and we handle all the records properly. Also safari does not have the `takeRecords` which needs extra unnecessary `typeof === function` check. Hence removed it.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -66,10 +66,19 @@ class TransactionService {
...
+         * relative start time of transaction is set to zero
+         * to include all the buffered events
+         */
         if (tr.type === PAGE_LOAD) {
           capturePaint = true
         }
-        const { spans, marks } = captureObserverEntries(list, { capturePaint })
+        const { spans, marks } = captureObserverEntries(list, {
+          capturePaint,
+          trStart: capturePaint ? 0 : tr._start
```
jahtalab, line 80: I think we should use `tr.type === PAGE_LOAD` , since capturePaint is not relevant here.

jahtalab (approved):
Thanks @vigneshshanmugam ,  I think we should do a refactoring of `captureObserverEntries` to avoid similar problems with other metrics. Also, as part of that I think some general correctness tests should be added. Let's do these separately.

Also, if I understand correctly, this bug only affects our latest version `5.7.0` since we didn't capture any buffered metrics for transactions other than `page-load`, is that correct?

vigneshshanmugam:
>  I think we should do a refactoring of captureObserverEntries to avoid similar problems with other metrics. Also, as part of that I think some general correctness tests should be added. Let's do these separately.

Let me know if you have any ideas on the testing part. If browsers started reporting other metrics which they don't at the moment, then we will see these bugs in future. Not a concern right now as web vitals are reported only for hard navigation. But worth refactoring that part.

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: I am opening a new PR to refactor this bit, we can use a different flag as capturePaint was added long before when we had only paint metrics.

PR #996, opened 2021-03-19
chore: update angular to latest 11.2.5

vigneshshanmugam:
+ Updates angular to the latest release 11.2.5 
+ Make sure our library works without any issues in latest version of angular without any code changes. Running unit and e2e tests ensures that it works.

As the alpha is rolled out, this would be final update before we do the major bump and release without `alpha` tags.


vigneshshanmugam:
Will clean this PR after
+ merging the `feat-angular` branch with master
+ Releasing the major v2.0.0 for `@elastic/apm-rum-angular` without alpha tags.

PR #1000, opened 2021-03-23
docs: update angular integration for >9 versions

vigneshshanmugam:
+ part of #962 
+ Overrides the previous integration document with the new changes.

### DIFF
--- a/docs/angular-integration.asciidoc
+++ b/docs/angular-integration.asciidoc
@@ -6,7 +6,7 @@ This document covers how to use Real User Monitoring JavaScript agent with Angul
 [[angular-supported-versions]]
 ==== Supported versions
 
-This integration supports Angular versions < 9.
+This integration supports Angular versions ≥ 9.0.
 
 [[installing-angular-integration]]
 ==== Installing Elastic APM Angular package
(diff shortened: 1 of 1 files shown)


vigneshshanmugam:
@bmorelli25 We have not published the new versions yet, I would like to get your review so we can prepare the alpha release and ask users to test the latest versions. Also do you think we need to have dedicated section for angular versions <9 or is there a way to link to the previous view of the docs in some way?

docs/angular-integration.asciidoc
```diff
@@ -15,43 +15,43 @@ Install the `@elastic/apm-rum-angular` package as a dependency to your applicati
 
 [source,bash]
 ----
-npm install @elastic/apm-rum-angular --save
+npm install @elastic/apm-rum-angular@alpha --save
 ----
 
+NOTE: If you are in Angular version < 9.0, please use @elastic/apm-rum-angular 1.x versions for
+instrumenting your application.
```
bmorelli25, line 22: Seeing as Angular <9 is EOL, I don't think it makes sense to maintain a separate section for the old docs. However, the old docs will continue to live on in version 4.x of the RUM agent reference, so we can link there as a convenience for now.

```suggestion
NOTE: If you are using an Angular version < 9.0, use @elastic/apm-rum-angular 1.x to instrument your application. Details are available in a https://www.elastic.co/guide/en/apm/agent/rum-js/4.x/angular-integration.html[prior release].
```

bmorelli25: If a new major is releasing for this change, we can edit the link above to point to version 5.x instead of 4.x

vigneshshanmugam: Not really the major is only going to be effective for Angular package not for other packages, so we can reference to 4.x versions. Good idea 👍

vigneshshanmugam:
@bmorelli25 Apologies in advance, I had to rebase all my commits as a result of publishing the alpha version and lost the coauthored commit. Hope you are okay with it.

bmorelli25:
No worries 😄

Issue #634, opened 2020-02-19
Introduce the user session concept


jahtalab:

paulb-elastic:
Drew has documented the use cases for sessions, [here](https://docs.google.com/document/d/1OvcdngoVi9jXgurV36OUXCWWIMJ6FrQwbMhgjEcofmA/edit#)

CBlaisure:
I was provided this issue from a question in Slack ([Link](https://elastic.slack.com/archives/C5L79P420/p1623771795045100)) after asking about monitoring user usage with RUM.  I have 2 basic questions.

1) Given that this was previously planned for 7.13 but is now in backlog, where does it stand on the delivery roadmap (roughly)?

2) If user sessions aren't delivered, what are the recommendations from the APM team in tracking this info in a custom way using existing APM/RUM tooling?

Our use case.  We are building a new knowledge base tool where Support Engineers will be creating, commenting, and publishing articles for customers.  My goal is to measure internal engagement with the new tool based on different user activities (Still TBD).  Happy to jump on a call and discuss this further if you are looking for use cases and real-world test environments.  

cc: @skh

PR #1005, opened 2021-03-31
feat(rum): add session to transactions

### LINKED ISSUE #998: Add session id and transaction number to transaction docs
Whilst working through the RUM Session implementation (https://github.com/elastic/apm-agent-rum-js/issues/634), we should at least add the session id to all transaction documents, so that we (and/or users) can manually report on the session data available (albeit likely with less than efficient queries).  Similarly, it should be possible to identify the order of the transactions from just the transaction documents themselves (whilst we don’t have anything for sessions).  This can also be used to compare the performance of first page hits to other pages.

ACs:
* Add the session id to transaction documents
* Add a transaction number identifier to transaction documents for that session
  * e.g. `transactionNumber=1` for the first transaction (index from `1` for the `first` as this is more human readable, for anyone running queries)
  * This will be related to the active session (e.g. if not browsing the site and the session has expired, the next page hit will generate a new session id, and the transaction number will be reset back to `1` again)
* The APM Server will need to be updated to ensure [these two new fields are indexed](https://github.com/elastic/apm-server/issues/5032)

jahtalab:
This PR adds the session information to transactions. Part of #634

Fixes #998

Todo:
- [x] Add more tests

### DIFF
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -90,7 +90,8 @@ class Config {
       centralConfig: false,
       monitorLongtasks: true,
       apiVersion: 2,
-      context: {}
+      context: {},
+      session: false
     }
 
     this.events = new EventHandler()
@@ -240,15 +241,19 @@ class Config {
   }
 
   getLocalConfig() {
-    let config = sessionStorage.getItem(LOCAL_CONFIG_KEY)
+    let config = localStorage.getItem(LOCAL_CONFIG_KEY)
     if (config) {
       return JSON.parse(config)
     }
   }
 
-  setLocalConfig(config) {
+  setLocalConfig(config, merge) {
     if (config) {
-      sessionStorage.setItem(LOCAL_CONFIG_KEY, JSON.stringify(config))
+      if (merge) {
+        const prevConfig = this.getLocalConfig()
+        config = { ...prevConfig, ...config }
+      }
+      localStorage.setItem(LOCAL_CONFIG_KEY, JSON.stringify(config))
     }
   }
 }
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -240,8 +240,18 @@ class Config {
     return errors
   }
 
+  /**
+   * The localStorage is needed for session feature
+   * but since this is a breaking change for central config,
+   * we're using the session flag until the next major version
+   * which should remove the use of sessionStorage.
+   */
   getLocalConfig() {
-    let config = localStorage.getItem(LOCAL_CONFIG_KEY)
+    let storage = sessionStorage
+    if (this.config.session) {
+      storage = localStorage
+    }
+    let config = storage.getItem(LOCAL_CONFIG_KEY)
     if (config) {
       return JSON.parse(config)
     }
@@ -253,7 +263,11 @@ class Config {
         const prevConfig = this.getLocalConfig()
         config = { ...prevConfig, ...config }
       }
-      localStorage.setItem(LOCAL_CONFIG_KEY, JSON.stringify(config))
+      let storage = sessionStorage
+      if (this.config.session) {
+        storage = localStorage
+      }
+      storage.setItem(LOCAL_CONFIG_KEY, JSON.stringify(config))
     }
   }
 }
--- a/packages/rum-core/src/common/constants.js
+++ b/packages/rum-core/src/common/constants.js
@@ -160,7 +160,7 @@ const TRUNCATED_TYPE = '.truncated'
 const KEYWORD_LIMIT = 1024
 
 /**
- * Session timeout is set to 30 minutes.
+ * Session timeout is set to 30 minutes (in milliseconds).
  */
 const SESSION_TIMEOUT = 30 * 60000
 
--- a/packages/rum-core/src/common/constants.js
+++ b/packages/rum-core/src/common/constants.js
@@ -159,6 +159,11 @@ const TRUNCATED_TYPE = '.truncated'
  */
 const KEYWORD_LIMIT = 1024
 
+/**
+ * Session timeout is set to 30 minutes.
+ */
+const SESSION_TIMEOUT = 30 * 60000
+
 export {
   SCHEDULE,
   INVOKE,
@@ -204,5 +209,6 @@ export {
   FIRST_INPUT,
   LAYOUT_SHIFT,
   OUTCOME_SUCCESS,
-  OUTCOME_FAILURE
+  OUTCOME_FAILURE,
+  SESSION_TIMEOUT
 }
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -50,11 +50,13 @@ import {
   PAINT,
   TRUNCATED_TYPE,
   FIRST_INPUT,
-  LAYOUT_SHIFT
+  LAYOUT_SHIFT,
+  SESSION_TIMEOUT
 } from '../common/constants'
 import { addTransactionContext } from '../common/context'
 import { __DEV__, state } from '../state'
 import { slugifyUrl } from '../common/url'
+import { generateRandomId } from '../../dist/lib/common/utils'
 
 class TransactionService {
   constructor(logger, config) {
@@ -308,6 +310,7 @@ class TransactionService {
               }
             }
           }
+          this.setSession(tr)
         }
         /**
          * Categorize the transaction based on the current location
@@ -350,6 +353,43 @@ class TransactionService {
         }
       }
     )
+  }
+
+  setSession(tr) {
+    let session = this._config.get('session')
+    if (session) {
+      if (typeof session == 'boolean') {
+        tr.session = {
+          id: generateRandomId(16),
+          sequence: 1
+        }
+      } else {
+        if (
+          session.timestamp &&
+          Date.now() - session.timestamp > SESSION_TIMEOUT
+        ) {
+          tr.session = {
+            id: generateRandomId(16),
+            sequence: 1
+          }
+        }
+
+        tr.session = {
+          id: session.id,
+          sequence: session.sequence ? session.sequence + 1 : 1
+        }
+      }
+
+      const sessionConfig = {
+        session: {
+          id: tr.session.id,
+          sequence: tr.session.sequence,
+          timestamp: Date.now()
+        }
+      }
+      this._config.setConfig(sessionConfig)
+      this._config.setLocalConfig(sessionConfig, true)
+    }
   }
 
   adjustTransactionTime(transaction) {
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -35,7 +35,8 @@ import {
   extend,
   getEarliestSpan,
   getLatestNonXHRSpan,
-  isPerfTypeSupported
+  isPerfTypeSupported,
+  generateRandomId
 } from '../common/utils'
 import { captureNavigation } from './capture-navigation'
 import {
@@ -56,7 +57,6 @@ import {
 import { addTransactionContext } from '../common/context'
 import { __DEV__, state } from '../state'
 import { slugifyUrl } from '../common/url'
-import { generateRandomId } from '../../dist/lib/common/utils'
 
 class TransactionService {
   constructor(logger, config) {
--- a/packages/rum-core/src/performance-monitoring/transaction-service.js
+++ b/packages/rum-core/src/performance-monitoring/transaction-service.js
@@ -357,6 +357,7 @@ class TransactionService {
 
   setSession(tr) {
     let session = this._config.get('session')
+
     if (session) {
       if (typeof session == 'boolean') {
         tr.session = {
@@ -372,11 +373,11 @@ class TransactionService {
             id: generateRandomId(16),
             sequence: 1
           }
-        }
-
-        tr.session = {
-          id: session.id,
-          sequence: session.sequence ? session.sequence + 1 : 1
+        } else {
+          tr.session = {
+            id: session.id,
+            sequence: session.sequence ? session.sequence + 1 : 1
+          }
         }
       }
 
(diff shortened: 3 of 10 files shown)


packages/rum-core/src/common/constants.js
```diff
@@ -159,6 +159,11 @@ const TRUNCATED_TYPE = '.truncated'
  */
 const KEYWORD_LIMIT = 1024
 
+/**
+ * Session timeout is set to 30 minutes.
+ */
+const SESSION_TIMEOUT = 30 * 60000
```
vigneshshanmugam, line 165:
```suggestion
const SESSION_TIMEOUT = 30 * 60
```

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,43 @@ class TransactionService {
     )
   }
 
+  setSession(tr) {
+    let session = this._config.get('session')
+    if (session) {
+      if (typeof session == 'boolean') {
+        tr.session = {
+          id: generateRandomId(16),
```
vigneshshanmugam, line 363: Any reason for keeping the id to 16 digits?

packages/rum-core/src/common/config-service.js
```diff
@@ -240,15 +241,19 @@ class Config {
   }
 
   getLocalConfig() {
-    let config = sessionStorage.getItem(LOCAL_CONFIG_KEY)
+    let config = localStorage.getItem(LOCAL_CONFIG_KEY)
```
vigneshshanmugam, line 244: This is going to affect centralConfig as well right? It's a breaking change as session is different from local storage and if a user configured transactionSampleRate, then thats not going to change ever unless the user explicits clears the storage.

packages/rum-core/src/common/config-service.js
```diff
@@ -240,15 +241,19 @@ class Config {
...
 
-  setLocalConfig(config) {
+  setLocalConfig(config, merge) {
     if (config) {
-      sessionStorage.setItem(LOCAL_CONFIG_KEY, JSON.stringify(config))
+      if (merge) {
+        const prevConfig = this.getLocalConfig()
+        config = { ...prevConfig, ...config }
+      }
+      localStorage.setItem(LOCAL_CONFIG_KEY, JSON.stringify(config))
```
vigneshshanmugam, line 256: Any reason for not using a cookie? They are easier to maintain and reason about. one issue i see here with local storage is, we never clean up the key from the user's machine which is bad and with cookie you set expires then its cleaned up by the browser.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -308,6 +310,7 @@ class TransactionService {
               }
             }
           }
+          this.setSession(tr)
```
vigneshshanmugam, line 313: The place where we are calling this is wrong. Session should start right when the navigation started not when our transaction is ended and we have to associate them for each transaction that ever happens within that particular session.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,43 @@ class TransactionService {
     )
   }
 
+  setSession(tr) {
+    let session = this._config.get('session')
+    if (session) {
+      if (typeof session == 'boolean') {
```
vigneshshanmugam, line 361: Better to keep the object shape consistent here, first its boolean and then its object.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,43 @@ class TransactionService {
...
+          id: session.id,
+          sequence: session.sequence ? session.sequence + 1 : 1
+        }
+      }
+
+      const sessionConfig = {
+        session: {
+          id: tr.session.id,
+          sequence: tr.session.sequence,
+          timestamp: Date.now()
```
vigneshshanmugam, line 387: `session.start` matches it better than timestamp

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,43 @@ class TransactionService {
...
+        ) {
+          tr.session = {
+            id: generateRandomId(16),
+            sequence: 1
+          }
+        }
+
+        tr.session = {
+          id: session.id,
+          sequence: session.sequence ? session.sequence + 1 : 1
```
vigneshshanmugam, line 379: This feels wrong, what does sequence refer to when it comes to measuring session length? 

Right now, this would be increasing for each and every `click` transactions on the page which feels incorrect. Is this what we want?

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,43 @@ class TransactionService {
     )
   }
 
+  setSession(tr) {
```
vigneshshanmugam, line 358: this function is super confusing. can you change this a bit?.

vigneshshanmugam (commented):
@jahtalab I have few other concerns as well, Lets catch up offline to go over the items.

packages/rum-core/src/common/constants.js
jahtalab: This value is in milliseconds, I'll change the comment to make it clear.

packages/rum-core/src/common/config-service.js
jahtalab: Yeah, I agree it should be a breaking change. But since the session feature is behind a feature flag (disabled by default), we can use that flag here to use the correct storage and only switch on our upcoming major version.

packages/rum-core/src/common/config-service.js
jahtalab: There are a few problems with using cookies here but the main issue is that, we don't want this data to be sent to the server on every request. Essentially we would be trading network bandwidth for a little bit of local storage, which I don't is a good idea.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: I actually started with creating the session on the transaction start, but considering that there are some cases in which we might not send (or even end) a transaction, I think this creates inconsistencies, e.g. we start a session for a transaction that we never send to the server and the session would have wrong data.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: The alternative here is to add another configuration option, I'm not against that idea either, but it seems unnecessary to me.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: @vigneshshanmugam , this is not a very useful comment, I'm happy to make this function more clear, but you need to provide more detail on what exactly do you mean here.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: Actually, we currently use sessions only for page loads (and maybe route-changes in the future), so click transactions are not involved. 

Please see the original issue for a bit more explanation on the sequence field, but happy to go into more detail if needed.

packages/rum-core/src/performance-monitoring/transaction-service.js
vigneshshanmugam: My bad, forgot to post the alternative snippet, was thinking to reduce the conditionals in this way or similar pattern. 

```js
setSession(tr) {
    let session = this._config.get("session");
    let value = {};
    if (
      typeof session == "boolean" ||
      (session.timestamp && Date.now() - session.timestamp > SESSION_TIMEOUT)
    ) {
      value = {
        id: generateRandomId(16),
        sequence: 1,
      };
    } else {
      value = {
        id: session.id,
        sequence: session.sequence ? session.sequence + 1 : 1,
      };
    }
    tr.session = value;

    const sessionConfig = {
      session: { timestamp: Date.now(), ...value },
    };
    this._config.setConfig(sessionConfig);
    this._config.setLocalConfig(sessionConfig, true);
  }

```

Lets not do this now, May be this is going to change after our discussion.

packages/rum-core/src/common/config-service.js
vigneshshanmugam: You mean sending session value on every APM request?  Thats a single header value on every request and headers are compressed with newer protocols like HTTP2 and QUIC which is too small to even reason about. The main issue here is we are leaving this data on the user's machine which is more concerning IMO. If we can find a way to delete this data, we are good with either one of them.

packages/rum-core/src/common/config-service.js
jahtalab: I've changed added the flag here to use the correct storage until we're ready to make the change in a major version.

packages/rum-core/src/common/config-service.js
jahtalab: Generally, I don't think it's a good idea to send data on the network, if that data only is needed locally.
In this case, we should investigate ways to delete the data, rather than sending these configs on each request.

jahtalab:
Meeting note:
- Discussed setting the session timestamp at the start of the transaction, but since we want to be consistent with sequence number we're doing it at transaction end, might reconsider this in the future.
- Double check localStorage issue across domains.
- Added a todo to only allow session.id on config.
- Change timestamp to `lastActive`

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,44 @@ class TransactionService {
...
+          session.timestamp &&
+          Date.now() - session.timestamp > SESSION_TIMEOUT
+        ) {
+          tr.session = {
+            id: generateRandomId(16),
+            sequence: 1
+          }
+        } else {
+          tr.session = {
+            id: session.id,
```
vigneshshanmugam, line 378: feels unnecessary to assign the same id again. maybe directly update the sequence alone.

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,44 @@ class TransactionService {
...
+            sequence: session.sequence ? session.sequence + 1 : 1
+          }
+        }
+      }
+
+      const sessionConfig = {
+        session: {
+          id: tr.session.id,
+          sequence: tr.session.sequence,
+          timestamp: Date.now()
```
vigneshshanmugam, line 388: we are updating the timestamp every time session is set which feels incorrect? this should be only happen when we start session for first time or session expired right?

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -635,6 +637,80 @@ describe('TransactionService', function () {
...
+    const clock = jasmine.clock()
+    clock.install()
+    clock.mockDate()
+    config.setConfig({ session: true })
+    const tr = new Transaction('test', 'test')
+    transactionService.setSession(tr)
+    expect(tr.session.id).toBeDefined()
+    const firstConfig = config.getLocalConfig()
+    const firstTime = firstConfig.session.timestamp
+    console.log(firstConfig)
```
vigneshshanmugam, line 664: nit: remove console

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -635,6 +637,80 @@ describe('TransactionService', function () {
...
+    expect(firstConfig.session).toEqual({
+      id: tr.session.id,
+      sequence: 1,
+      timestamp: jasmine.any(Number)
```
vigneshshanmugam, line 668: let's test with correct timestamp, `timestamp === firstTime`

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -635,6 +637,80 @@ describe('TransactionService', function () {
...
+    })
+
+    clock.tick(500)
+    const tr2 = new Transaction('tr2', 'test')
+    transactionService.setSession(tr2)
+
+    expect(tr2.session).toEqual({ id: tr.session.id, sequence: 2 })
+    const secondConfig = config.getLocalConfig()
+    console.log(secondConfig)
```
vigneshshanmugam, line 677: nit: remove `console`

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -352,6 +355,44 @@ class TransactionService {
     )
   }
 
+  setSession(tr) {
+    let session = this._config.get('session')
+
+    if (session) {
```
vigneshshanmugam, line 361: nit: too much nesting, it's good to return if session is false.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: We're setting the id from local storage on the transaction, this is needed here.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: As we discussed yesterday, this is supposed to be `lastActive`, in other words, we keep the session active until 30 mins of inactivity. This was what we discussed previously, but can we can change in the future if needed.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: Maybe this is a difference in style, but I think this is more clear compared to a return at the top. Let's keep it for now.

packages/rum-core/src/common/config-service.js
jahtalab: Discussed this in the meeting, we're going to use localStorage since we don't want to send unnecessary data over network on every request.

packages/rum-core/src/performance-monitoring/transaction-service.js
jahtalab: @vigneshshanmugam , I'm going to keep the name `timestamp` for this field for now, until we finalise the last activity vs start/end timestamps. But this is still useful for testing data transforms, so we can release as is.

PR #1012, opened 2021-04-19
chore: add e2e test for sessions

jahtalab:
This PR adds e2e test for sessions. it also adds session to APM server load testing

Part of #634

### DIFF
--- a/scripts/apm-server-load-test.js
+++ b/scripts/apm-server-load-test.js
@@ -32,6 +32,9 @@ const { version } = require('../packages/rum/package.json')
 const { writeFile } = require('fs').promises
 
 const serverUrl = process.env.APM_SERVER_URL || 'http://localhost:8200'
+const esUrl = process.env.ES_URL || 'http://localhost:9200'
+const esAuth = process.env.ES_AUTH
+const debug = process.env.DEBUG
 
 /**
  * To make the random id generation work.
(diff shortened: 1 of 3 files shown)


scripts/apm-server-load-test.js
```diff
@@ -167,9 +177,13 @@ function generateTransaction(spanCount) {
       sampled: spanCount != 0,
       breakdown,
       context: {
+        tags: {
```
vigneshshanmugam, line 180: any reason for not adding to the transaction.session field?

scripts/apm-server-load-test.js
```diff
@@ -215,52 +229,119 @@ async function postPayload(url, payload) {
...
-
-  const result = await generatePayloads(10)
+  const results = []
+  for (let iteration = 0; iteration < iterations; iteration++) {
+    if (debug) {
+      console.log(`Iteration: ${iteration}`)
+    }
+    const result = await generatePayloads(100)
+    results.push(result)
+    await asyncTimeout(1000)
```
vigneshshanmugam, line 336: is this for simulating load? may be give it a random number between `1-1000ms` as well for delay?

vigneshshanmugam (approved):
E2E test fails though waiting for the element. Looks good otherwise.

scripts/apm-server-load-test.js
jahtalab: This is actually to avoid hitting the rate limit, otherwise requests will be rejected

scripts/apm-server-load-test.js
jahtalab: Mainly because it's not recognised by APM server at the moment, I've added it for future support.

PR #1017, opened 2021-04-21
fix(rum): add number, boolean types to addLabels

jahtalab:


packages/rum/src/index.d.ts
```diff
@@ -179,7 +179,7 @@ type FilterFn = (payload: Payload) => Payload | boolean | void
 type Payload = { [key: string]: any }
 
 type TaskId = string | number
-type LabelValue = string
+type LabelValue = string | number | boolean
```
vigneshshanmugam, line 182: we could also add `any` to the list since we convert those to string anyways in our code. Also to match with docs
```
value - Any string, boolean, or number. All other data types will be converted to a string before being sent to the APM Server.
```

vigneshshanmugam (approved):

packages/rum/src/index.d.ts
jahtalab: I think we shouldn't support the any type here, e.g. `[object Object]` label value is not useful. But, also dropping support for it, is probably a breaking change.

PR #1019, opened 2021-04-28
feat(rum-core): add apmServerRequest config

### LINKED ISSUE #1018: Provide a configuration to modify the RUM agent request to APM server
In certain use-cases, it is needed to modify or stop requests that the RUM agent makes to the APM server. A possible solution can be adding a configuration that accepts a function which is called before a requests is send to the APM server, i.e:

```
elasticApm.init({
  apmServerRequest: function (xhr) { 
    xhr.setRequestHeader('custom', 'value')
    return true
  }
})
```

If this function returns a falsy value the request will be rejected, this will provide the possibility to skip some requests.

jahtalab:
This PR adds apmServerRequest config to intercept requests to APM server.

Fixes #1018

### DIFF
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -81,7 +81,7 @@ class ApmServer {
     const headers = {
       'Content-Type': 'application/x-ndjson'
     }
-    const apmServerRequest = this._configService.get('apmServerRequest')
+    const apmRequest = this._configService.get('apmRequest')
     return compressPayload(payload, headers)
       .catch(error => {
         if (__DEV__) {
@@ -90,7 +90,7 @@ class ApmServer {
             error.message
           )
         }
-        return { payload, headers, beforeSend: apmServerRequest }
+        return { payload, headers, beforeSend: apmRequest }
       })
       .then(result => this._makeHttpRequest('POST', endPoint, result))
       .then(({ responseText }) => responseText)
@@ -158,7 +158,7 @@ class ApmServer {
 
       let canSend = true
       if (typeof beforeSend === 'function') {
-        canSend = beforeSend(xhr, url, method, headers, payload)
+        canSend = beforeSend({ url, method, headers, payload, xhr })
       }
       if (canSend) {
         xhr.send(payload)
@@ -190,11 +190,11 @@ class ApmServer {
       configEndpoint += `&ifnonematch=${localConfig.etag}`
     }
 
-    const apmServerRequest = this._configService.get('apmServerRequest')
+    const apmRequest = this._configService.get('apmRequest')
 
     return this._makeHttpRequest('GET', configEndpoint, {
       timeout: 5000,
-      beforeSend: apmServerRequest
+      beforeSend: apmRequest
     })
       .then(xhr => {
         const { status, responseText } = xhr
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -81,6 +81,7 @@ class ApmServer {
     const headers = {
       'Content-Type': 'application/x-ndjson'
     }
+    const apmServerRequest = this._configService.get('apmServerRequest')
     return compressPayload(payload, headers)
       .catch(error => {
         if (__DEV__) {
@@ -89,7 +90,7 @@ class ApmServer {
             error.message
           )
         }
-        return { payload, headers }
+        return { payload, headers, beforeSend: apmServerRequest }
       })
       .then(result => this._makeHttpRequest('POST', endPoint, result))
       .then(({ responseText }) => responseText)
@@ -119,7 +120,11 @@ class ApmServer {
     return new Error(message)
   }
 
-  _makeHttpRequest(method, url, { timeout = 10000, payload, headers } = {}) {
+  _makeHttpRequest(
+    method,
+    url,
+    { timeout = 10000, payload, headers, beforeSend } = {}
+  ) {
     return new Promise(function (resolve, reject) {
       var xhr = new window.XMLHttpRequest()
       xhr[XHR_IGNORE] = true
@@ -150,7 +155,20 @@ class ApmServer {
         const { status, responseText } = xhr
         reject({ url, status, responseText })
       }
-      xhr.send(payload)
+
+      let canSend = true
+      if (typeof beforeSend === 'function') {
+        canSend = beforeSend(xhr, url, method, headers, payload)
+      }
+      if (canSend) {
+        xhr.send(payload)
+      } else {
+        reject({
+          url,
+          status: 0,
+          responseText: 'Request rejected by user configuration.'
+        })
+      }
     })
   }
 
@@ -172,7 +190,12 @@ class ApmServer {
       configEndpoint += `&ifnonematch=${localConfig.etag}`
     }
 
-    return this._makeHttpRequest('GET', configEndpoint, { timeout: 5000 })
+    const apmServerRequest = this._configService.get('apmServerRequest')
+
+    return this._makeHttpRequest('GET', configEndpoint, {
+      timeout: 5000,
+      beforeSend: apmServerRequest
+    })
       .then(xhr => {
         const { status, responseText } = xhr
         if (status === 304) {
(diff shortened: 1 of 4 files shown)


vigneshshanmugam (commented):
Thinking a bit loud, I see two issues with the current approach.

1. Tying in to XHR as the method of transport which wont work as we will be replacing the beaconing mechanism sooner with sendBeacon or Fetch keepalive. 

2. Not allowing any customization on controlling timeouts, responses from the APM server. 


We could improve the interface by exposing a `onReport` or similar function where user would control the whole payload and add whatever headers they need and also beacon it with their choice. RUM agent would only call that reporting function when the page is loaded or when the user has navigated away from the page (pagehide, visibility change, etc)
```js

function onReport(url, payload) {
  fetch(url, {
    method: 'POST',
    body: payload,
    keepalive: true,
  });
}

const config = {
    serviceName: 'test'
    onReport,
}
apm.init(config);

```

For now our default reporting would be to send it via XHR via `_makeHttpRequest`. I think this solution is better for the long term.

jahtalab:
@vigneshshanmugam , With the current approach controlling the whole request is also possible, e.g.:

```
apm.init({
  apmServerRequest(xhr, url, method, headers, payload) {
      fetch(url, {
      method: method,
      body: payload,
      keepalive: true,
    });
    return false
  }
})
```

But if users don't care about controlling the request and only want to add a header (as it is requested in that issue), this makes it simpler:
```
  apmServerRequest(xhr) {
    xhr.setRequestHeader('custom', 'value')
    return true
  }
```

vigneshshanmugam:
@jahtalab Yeah totally, the advantage is there when it comes to users just needing to add extra headers. But I am worried with the tying in on the XHR approach as we will be moving out of XHR anytime soon for transport mechanism. 

Going a step back, As we are trying to solve only the problem of extra headers for now, WDYT about introducing a config only to add extra headers for the APM server and dont expose the whole interface. This should solve the current problem and we can think about coming back to it once we establish the proper transport mechanism. 

```js
const config = {
    serviceName: 'test'
    serverHeaders: {
         "custom": "value"
    },
}
apm.init(config);
```

jahtalab:
@vigneshshanmugam , I think the change to the fetch with keepalive, should be done in a major version since that's a breaking change. This change also provides a way for offline use-case as described in #262, so I think we should use the function config since it cover more use-cases.

We can change the signature of `apmServerRequest` config in the major version that uses fetch, when are ready to make that change.

packages/rum-core/src/common/apm-server.js
```diff
@@ -150,7 +155,20 @@ class ApmServer {
         const { status, responseText } = xhr
         reject({ url, status, responseText })
       }
-      xhr.send(payload)
+
+      let canSend = true
+      if (typeof beforeSend === 'function') {
+        canSend = beforeSend(xhr, url, method, headers, payload)
```
vigneshshanmugam, line 161: Is the expectation here that the provided function must be `synchronous`?  If the users provide `async fn` here, our error handling logic would break and result in not capturing those errors.

vigneshshanmugam, line 161: Also let's pass those params as object if we are going with this approach. Too many args

vigneshshanmugam:
@jahtalab I am not proposing to change the implementation to keep alive or sendBeacon. If we go with this approach as you have mentioned we can do the underlying transport mechanism only on major as we are exposing the XHR as part of the function parameters and it would be breaking. 

Otherwise, we can rely on not tying in with any transport mechanism(XHR/Fetch/Beacon) and can change implementation beneath in minor version which will not have any impact and have the same API for providing headers.

If we are okay with doing the transport layer change in major version, lets go with this approach and address these
- Document the function signature might change in future and use it only for advanced use-cases. 
- Explain if we expect a sync/async function and add tests for the same in this PR.

packages/rum-core/src/common/config-service.js
```diff
@@ -91,7 +91,8 @@ class Config {
       monitorLongtasks: true,
       apiVersion: 2,
       context: {},
-      session: false
+      session: false,
+      apmServerRequest: null
```
vigneshshanmugam, line 95: we can just keep it as `onRequest` to keep it simpler? Thoughts?

packages/rum-core/test/common/apm-server.spec.js
```diff
@@ -404,6 +404,40 @@ describe('ApmServer', function () {
...
+        expect(headers).toBeUndefined()
+        return true
+      }
+    })
+  })
+
+  it('should reject the request if beforeSend returns falsy', async () => {
+    let promise = apmServer._makeHttpRequest('POST', '/test', {
+      payload: 'test',
+      beforeSend: (xhr, url, method, headers, payload) => {
```
vigneshshanmugam, line 425: let's add a test for async case.

jahtalab:
@vigneshshanmugam, I think a changing to keep alive should be a breaking change regardless of this change. 

The function callback is supposed to be synchronous, since we only need to know whether the user wants us to handle the request or would they do it separately. we can support the async scenario but at this point I don't think it's necessary for the use-cases.
I will change the function argument to an object so it becomes easier to change this function in the future.

packages/rum-core/src/common/apm-server.js
jahtalab: That's a good idea, will do that.

packages/rum-core/src/common/config-service.js
jahtalab: I thought about `onRequest` but I think that's too generic and might be confused for all http request (regards of the apm server)

packages/rum-core/src/common/apm-server.js
jahtalab: That's correct! see the main comment!

packages/rum-core/test/common/apm-server.spec.js
jahtalab: Let's decide about the async use-case first.

vigneshshanmugam: Yeah, we dont need to for now, we can skip and document it.

packages/rum-core/src/common/config-service.js
vigneshshanmugam: Hmmm. `apmRequest`? I was thinking about users building their own backend, but dont expect it to support our format though. Anyways I am fine with as it is or `apmRequest` up to you.

vigneshshanmugam:
Sounds good, lets get this in once the test passed, Do you want to do the docs in follow up PR or here? I am fine with both.

jahtalab:
@vigneshshanmugam you can see the docs here: #1025

packages/rum-core/src/common/apm-server.js
```diff
@@ -150,7 +155,20 @@ class ApmServer {
...
+      if (typeof beforeSend === 'function') {
+        canSend = beforeSend({ url, method, headers, payload, xhr })
+      }
+      if (canSend) {
+        xhr.send(payload)
+      } else {
+        reject({
+          url,
+          status: 0,
+          responseText: 'Request rejected by user configuration.'
```
vigneshshanmugam, line 169:
```suggestion
          responseText: 'Request rejected by configuration(apmRequest) .'
```

packages/rum-core/src/common/apm-server.js
```diff
@@ -150,7 +155,20 @@ class ApmServer {
...
-      xhr.send(payload)
+
+      let canSend = true
+      if (typeof beforeSend === 'function') {
+        canSend = beforeSend({ url, method, headers, payload, xhr })
+      }
+      if (canSend) {
+        xhr.send(payload)
+      } else {
+        reject({
```
vigneshshanmugam, line 166: forgot to catch this before, do we want to fail here explicitly which will show up as warnings in console `Failed sending events` or silently fail by resolve it with empty response text?

vigneshshanmugam (approved):

packages/rum-core/src/common/apm-server.js
jahtalab: Yeah, the warning is useful if users have made a mistake in their callback function.

PR #1046, opened 2021-06-20
feat(rum-react): support for React 17 and React Router 5

### LINKED ISSUE #1031: Support for React 17
Hi,

I was going to try out integrating this library with our react app, but it turns out it only supports React 16, while we are using 17. Are there plans of upgrading to this version?
Same with React Router, we use v5.

FedeOmoto:
Closes #1031


vigneshshanmugam:
Closing in favor of #1047, turned out to be a bit more complicated than i intended, Apologies for the new PR.

PR #1047, opened 2021-06-21
feat(rum-react): add official support for react 17

### LINKED ISSUE #1031: Support for React 17

vigneshshanmugam:
+ fix #1031 
+ RUM agent worked properly before with react 17 but with warnings during the installation phase as we relied on react 16 and older router versions for peer deps, This PR updates all of that deps and also fixes the test suites with new Enzyme adapters for 17.


shahzad31 (approved):
LGTM, Do we also have tests for Functional react components?

Didn't do smoke, only code !!

vigneshshanmugam:
> LGTM, Do we also have tests for Functional react components?

We have both kind in here - https://github.com/elastic/apm-agent-rum-js/tree/master/packages/rum-react/test/e2e/components and they are tested the same way.

PR #1048, opened 2021-06-22
fix: set stage result

kuisathaverat:
The catchError is hidden and stage error so we need to force the stage result to know which stage is failing.

```
Benchmarks: there was a response with an error. Review response {"took":8,"errors":true,"items":[{"index":{"_index":"benchmarks-rum-load-test","_type":"_doc","_id":"3tGPJXoBfbEDP-36W9O7","status":400,"error":{"type":"mapper_parsing_exception","reason":"failed to parse","caused_by":{"type":"not_x_content_exception","reason":"not_x_content_exception: Compressor detection can only be called on some xcontent bytes or compressed xcontent bytes"}}}}]}
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -242,7 +242,7 @@ pipeline {
               post {
                 always {
                   archiveArtifacts(allowEmptyArchive: true, artifacts: "${BASE_DIR}/${env.REPORT_FILE}", onlyIfSuccessful: false)
-                  catchError(message: 'sendBenchmarks failed', buildResult: 'FAILURE', stageResult: 'FAILURE') {
+                  catchError(message: 'sendBenchmarks failed', buildResult: 'SUCCESS', stageResult: 'UNSTABLE') {
                     log(level: 'INFO', text: "sendBenchmarks is ${env.CHANGE_ID?.trim() ? 'not enabled for PRs' : 'enabled for branches'}")
                     whenTrue(env.CHANGE_ID == null){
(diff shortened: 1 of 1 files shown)


v1v:
I'm not sure if they wanted to fail the whole build, worth to double check with the @elastic/apm-agent-rum

cachedout (approved):

vigneshshanmugam (commented):
I don't think we need this to be the default behaviour. Both benchmarks and load testing are good to have, we don't require CI to fail for these two steps. 

Let's not do this change.

kuisathaverat:
> I don't think we need this to be the default behaviour. Both benchmarks and load testing are good to have, we don't require CI to fail for these two steps.
> 
> Let's not do this change.

It is falling now, but you do not know in which stage because the stage is reported as a success, even though the build is marked as failure.

Let's see an example, the following build was failure, but all stages are green

image

If you check the load test that stage has a step that failed

image

If you do not report that stage as failed, you have to click on all stages to see what's going on.

vigneshshanmugam:
Hmmm yeah thats true, In that case can we mark that particular stage as failure, but still keep the full test as a PASS? 

We dont want to fail the build for load test and benchmarks. Thats my only concern as they are more prone to errors.

kuisathaverat:
ok, I have changed it to work as the bundle size, it marks the build as success and the stage as unstable (!)

vigneshshanmugam (approved):

Issue #1004, opened 2021-03-30
pageLoadSampled documented as string


thenewguy:
Please verify https://www.elastic.co/guide/en/apm/agent/rum-js/current/configuration.html#page-load-sampled

Should it be a boolean? Docs say data should be a string.  Can you provide an example.

vigneshshanmugam:
Good catch, it should be a boolean not a string. Would you like to open a PR?

ikalkov:
I was confused by the documentation as well. Would be great to have it fixed!

PR #1068, opened 2021-08-12
docs: Change `pageLoadSampled` to Boolean

bmorelli25:
Closes https://github.com/elastic/apm-agent-rum-js/issues/1004.


vigneshshanmugam (commented):
Thanks @bmorelli25 While you are at it, could you also include that this option is applicable only to `page-load` transactions. 

For the lint failure, you can fix the commit with `docs: update config` and it should pass,.

vigneshshanmugam (approved):

Issue #1045, opened 2021-06-18
export Transaction, Span, Options rum typings


vigneshshanmugam:
From Kibana core team, 

> I think the agent could export all the public types: AgentConfigOptions, TransactionOptions, Transaction, Span, SpanOptions. that user interact with.

+ Currently the agent only exposes the `APM`, `AgentConfigOptions` interface, We could also expose the Transaction, Span, TransactionOptions and SpanOptions typings in addition.

PR #1100, opened 2021-11-22
feat(rum): export rum types (#1045)

devcorpio:
# Summary

With this change the following types are going to be exposed:

- Transaction
- Span
- TransactionOptions
- SpanOptions

### DIFF
--- a/packages/rum/src/index.d.ts
+++ b/packages/rum/src/index.d.ts
@@ -64,14 +64,6 @@ declare module '@elastic/apm-rum' {
      */
     canReuse?: boolean
   }
-
-  export interface SpanOptions {
-    /**
-     * undocumented, might be removed in future versions
-     */
-    startTime?: number
-    sync: boolean
-  }
   export interface AgentConfigOptions {
     apiVersion?: 2 | 3
     serviceName?: string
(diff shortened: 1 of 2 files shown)


devcorpio:
In the file ```packages/rum/src/index.d.ts``` we still have `declare class Transaction` and `declare class Span`. Since now we have the types we don't need those statements anymore.

Although if we remove them we could be forcing users that were already relying on them given typescript make them available anyway! (without import)

Should we remove them? Any thoughts?

devcorpio:
Could we have any unexpected issue for removing` declare class SpanOptions` and `declare class Transaction`?

packages/rum/src/index.d.ts
```diff
@@ -24,6 +24,51 @@
  */
 
 declare module '@elastic/apm-rum' {
+  export class Span extends BaseSpan {
+    sync: boolean
```
vigneshshanmugam, line 28: nit: this is optional

packages/rum/src/index.d.ts
```diff
@@ -24,6 +24,51 @@
...
+     * undocumented, might be removed in future versions
+     */
+    isFinished(): boolean
+  }
+  export interface SpanOptions {
+    /**
+     * undocumented, might be removed in future versions
+     */
+    startTime?: number
+    sync: boolean
```
vigneshshanmugam, line 59: would make it as optional as well.

vigneshshanmugam (approved):
LGTM

Issue #1104, opened 2021-12-01
Cannot read properties of undefined (reading 'fetchStart')


kanadaj:
After deploying RUM, I don't seem to get data from some users. Got their data in Sentry though, so here it is:

### Browser
name | Chrome
-- | --
version | 96.0.4664

### OS
Name | Windows
-- | --
Version | 10

### SDK
Name | @elastic/apm-rum
-- | --
Version | 5.9.1

### Stacktrace
```ts
TypeError: Cannot read properties of undefined (reading 'fetchStart')
  at ? (./node_modules/@elastic/apm-rum-core/dist/es/common/compress.js:115:27)
  at Array.forEach(<anonymous>)
  at compressMarks(./node_modules/@elastic/apm-rum-core/dist/es/common/compress.js:113:31)
  at compressTransaction(./node_modules/@elastic/apm-rum-core/dist/es/common/compress.js:195:8)
  at ? (./node_modules/@elastic/apm-rum-core/dist/es/common/apm-server.js:313:94)
  at Array.map(<anonymous>)
  at ndjsonTransactions(./node_modules/@elastic/apm-rum-core/dist/es/common/apm-server.js:291:25)
  at sendEvents(./node_modules/@elastic/apm-rum-core/dist/es/common/apm-server.js:359:87)
  at onFlush(./node_modules/@elastic/apm-rum-core/dist/es/common/apm-server.js:33:27)
  at flush(./node_modules/@elastic/apm-rum-core/dist/es/common/queue.js:34:10)
  at apply(./node_modules/@elastic/apm-rum-core/dist/es/common/queue.js:20:20)
  at r(./node_modules/@sentry/browser/esm/helpers.js:72:23)
```

It seems to work for most users, and not for some others.

kanadaj:
To add a bit, I think it may be an issue caused by ads loading stuff in in the background, causing an issue with working out the end of the span? Not really sure though.

vigneshshanmugam:
Its strange how this could happen unless the transaction object is modified and the transaction marks are removed somewhere else in the code. Do you have any logic to change the transaction object through these API? https://www.elastic.co/guide/en/apm/agent/rum-js/current/agent-api.html#apm-add-filter.

We can always add a null check for navigationTiming here - https://github.com/elastic/apm-agent-rum-js/blob/07544e8fd1585430b192bd127c9c70cc1237ee52/packages/rum-core/src/common/compress.js#L123.

Can you help debug this? 

[...]

PR #1107, opened 2021-12-03
fix(rum-core): handle empty marks when compressing (#1104)

devcorpio:
# Summary

In order to prevent the agent to throw an error when compressing transactions we are now checking the existence of navigation timing and agent marks.

Following the scout principle news tests added which check that marks are being compressed properly 👍

### DIFF
--- a/packages/rum-core/src/common/compress.js
+++ b/packages/rum-core/src/common/compress.js
@@ -168,10 +168,6 @@ function compressAgentMarks(compressedNtMarks, agentMarks) {
     }
   }
 
-  if (Object.keys(compressed).length === 0) {
-    return null
-  }
-
   return compressed
 }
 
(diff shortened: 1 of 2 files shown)


packages/rum-core/src/common/compress.js
```diff
@@ -120,26 +120,56 @@ function compressMarks(marks) {
...
+      compressed.fp = fp
+    }
+    if (lp) {
+      compressed.lp = lp
+    }
   }
-  if (lp) {
-    compressed.a.lp = lp
+
+  if (Object.keys(compressed).length === 0) {
```
vigneshshanmugam, line 171: nit: we dont need this check, truncate model should take care of properties without keys.

vigneshshanmugam (approved):
LGTM

packages/rum-core/src/common/compress.js
devcorpio: I removed the check since the intake api is accepting an empty object as well.

vigneshshanmugam: Ah good point, null is better as as fields with null values will be removed when serialized,

PR #1113, opened 2021-12-09
chore: always use the same sauce labs config

graphaelli:


devcorpio:
Hi @graphaelli ,

Do you know the status of this PR? Is this change something that we want still to do?

graphaelli:
@vigneshshanmugam can you weigh in here?  My understanding is that this isn't used at all.

vigneshshanmugam:
Yes, we dont use the `SAUCELABS_SECRET_CORE` token for running the tests. We can remove them. 

Need to change the PR message for the lint to pass through. We can fix that.

vigneshshanmugam:
@devcorpio Need to lint the commit message as well to make it pass the lint check.

devcorpio (approved):

PR #1128, opened 2021-12-28
feat(rum-core): update cls definition

### LINKED ISSUE #1033: update the implementation of CLS to use new definition
+ More details on this thread - https://web.dev/evolving-cls/

+ This wont be breaking change as we only capture CLS events that are reported during the lifetime of the page load and with the new definition, the score would likely go down than before as `maximum session window with 1 second gap, capped at 5 seconds.`

devcorpio:
# Summary

+ fix #1033 

As per https://web.dev/evolving-cls/ this PR updates the way the RUM agent calculates the Cumulative Layout Shift metric.

The new technique used to calculate the metric is known as **max-session-gap1s-limit5s**

### DIFF
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -174,18 +174,35 @@ export function calculateTotalBlockingTime(longtaskEntries) {
 }
 
 /**
- * Calculates Cumulative Layout Shift using
+ * Calculates Cumulative Layout Shift through the max-session-gap1s-limit5s technique using
  * 'layout-shift' entries captured through PerformanceObserver.
  * https://developer.mozilla.org/en-US/docs/Web/API/LayoutShift
- * https://web.dev/cls/
+ * https://web.dev/evolving-cls/
  */
 export function calculateCumulativeLayoutShift(clsEntries) {
+  let maxScore = 0
+  let currentSessionScore = 0
+  let firstEventTime = Number.NEGATIVE_INFINITY
+  let previousEventTime = Number.NEGATIVE_INFINITY
+
   clsEntries.forEach(entry => {
     // Only add the layout shift events without recent user input
     if (!entry.hadRecentInput && entry.value) {
-      metrics.cls += entry.value
+      const shouldCreateNewSession =
+        entry.startTime - firstEventTime > 5000 ||
+        entry.startTime - previousEventTime > 1000
+      if (shouldCreateNewSession) {
+        firstEventTime = entry.startTime
+        currentSessionScore = 0
+      }
+
+      previousEventTime = entry.startTime
+      currentSessionScore += entry.value
+      maxScore = Math.max(maxScore, currentSessionScore)
     }
   })
+
+  metrics.cls = maxScore
 }
 
 /**
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -32,7 +32,6 @@ import {
 } from '../common/constants'
 import { noop, PERF } from '../common/utils'
 import Span from './span'
-import { state } from '../state'
 
 export const metrics = {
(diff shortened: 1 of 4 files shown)


awahab07 (approved):
LGTM!

The new CLS calculation method gives the same result as the approach described in https://web.dev/cls/#measure-cls-in-javascript. The test cases in the PR yield the same score for both algos.

packages/rum-core/src/state.js
```diff
@@ -28,7 +28,14 @@ const state = {
   //  Time when agent is bootstrapped and patching of modules happens
   bootstrapTime: null,
   // Time when the document is last backgrounded
-  lastHiddenStart: Number.MIN_SAFE_INTEGER
+  lastHiddenStart: Number.MIN_SAFE_INTEGER,
+  // Tracking info of cumulative layout shift calculation
+  cls: {
```
devcorpio, line 33: Since the max-session-gap1s-limit5s technique requires to have into account all the layout shift events over time a state has been created to ensure the proper calculation.

This state is only accessed/modified in the CLS calculation.

I'm open to proposals in case you prefer other ways of handling this. 😊

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -174,18 +175,36 @@ export function calculateTotalBlockingTime(longtaskEntries) {
...
  * 'layout-shift' entries captured through PerformanceObserver.
  * https://developer.mozilla.org/en-US/docs/Web/API/LayoutShift
- * https://web.dev/cls/
+ * https://web.dev/evolving-cls/
+ *
+ * We use clsState to keep track of previous layout shift events since the used technique requires
+ * to have them into account to perform the proper calculation
  */
-export function calculateCumulativeLayoutShift(clsEntries) {
+export function calculateCumulativeLayoutShift(clsState, clsEntries) {
```
vigneshshanmugam, line 186: Would actually put this state inside the metrics.cls to keep it simpler, we do something similar to TBT calc. Would reduce no of fields getting added to global state.

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -174,18 +175,36 @@ export function calculateTotalBlockingTime(longtaskEntries) {
...
   clsEntries.forEach(entry => {
     // Only add the layout shift events without recent user input
     if (!entry.hadRecentInput && entry.value) {
-      metrics.cls += entry.value
+      const shouldCreateNewSession =
+        entry.startTime - clsState.firstEventTime > 5000 ||
```
vigneshshanmugam, line 191: nit: firstEntryTime and prevEntryTime. Makes it easier to reason about than events.

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -174,18 +175,36 @@ export function calculateTotalBlockingTime(longtaskEntries) {
...
+        entry.startTime - clsState.previousEventTime > 1000
+      if (shouldCreateNewSession) {
+        clsState.firstEventTime = entry.startTime
+        clsState.currentSessionScore = 0
+      }
+
+      clsState.previousEventTime = entry.startTime
+      clsState.currentSessionScore += entry.value
+      clsState.maxScore = Math.max(
```
vigneshshanmugam, line 200: you could also get rid of maxScore, instead do metrics.cls.score to reduce the no of fields. Just for reduce bytes, nothing wrong with the current logic :)

vigneshshanmugam (commented):
Except for the state logic, Everything else looks good.

packages/rum-core/src/performance-monitoring/metrics.js
devcorpio: I agree, it is better to take advantage of the state that we already have in `metrics.js` file 👍

vigneshshanmugam (approved):
LGTM 🎉

PR #1131, opened 2022-01-05
chore: update dependencies pointed out by our automations

lucasfcosta:
# Summary

This fixes dependencies that were pointed out by our automations as needing updates.

Please notice that even though all the dependencies pointed out were dev dependencies, we still update them as they could impact the build, and thus they also need checking. Also, we had to force some downstream dependencies to resolve to particular versions, such as `lodash` and `semver-regex`. For that, we used [`npm-force-resolutions`](https://github.com/rogeriochaves/npm-force-resolutions).

@devcorpio has helped me update these dependencies (thank you 🙌 ) and will help me out with testing this on the RUM agent itself as he's more experienced with this codebase.

We also want to ensure CI is passing, which is why we created this PR early.

### DIFF
--- a/dev-utils/build.js
+++ b/dev-utils/build.js
@@ -56,6 +56,11 @@ const PACKAGE_TYPES = {
   VUE: 'VUE'
 }
 
+// We must exclude `parse5` from the packages we ignore
+// because otherwise it would not be transpiled. Therefore,
+// it would break IE11.
+const babelExclusionRule = /node_modules\/(?!parse5).+/
+
 function getBabelPresetEnv(bundleType) {
   const isBrowser = [
     BROWSER_DEV,
@@ -152,10 +157,7 @@ function getBabelConfig(bundleType, packageType) {
   let options = {
     comments: false,
     babelrc: false,
-    // We must exclude `parse5` from the packages we ignore
-    // because otherwise it would not be transplied. Therefore,
-    // it would break IE11.
-    exclude: /node_modules\/(?!parse5).+/,
+    exclude: babelExclusionRule,
     presets: getBabelPresetEnv(bundleType),
     plugins: []
   }
@@ -205,10 +207,7 @@ function getCommonWebpackConfig(bundleType, packageType) {
       rules: [
         {
           test: /\.(js|jsx|ts)$/,
-          // We must exclude `parse5` from the packages we ignore
-          // because otherwise it would not be transplied. Therefore,
-          // it would break IE11.
-          exclude: /node_modules\/(?!parse5).+/,
+          exclude: babelExclusionRule,
           loader: 'babel-loader',
           options: getBabelConfig(bundleType, packageType)
         }
--- a/dev-utils/build.js
+++ b/dev-utils/build.js
@@ -152,7 +152,10 @@ function getBabelConfig(bundleType, packageType) {
   let options = {
     comments: false,
     babelrc: false,
-    exclude: '/**/node_modules/**',
+    // We must exclude `parse5` from the packages we ignore
+    // because otherwise it would not be transplied. Therefore,
+    // it would break IE11.
+    exclude: /node_modules\/(?!parse5).+/,
     presets: getBabelPresetEnv(bundleType),
     plugins: []
   }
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -29,8 +30,7 @@
   },
   "lint-staged": {
     "*.{js,jsx,ts}": [
-      "eslint . --fix",
-      "git add"
```
devcorpio, line 33: `git add` is not required anymore with the new version of lint-staged we are using, that's why it has been removed 👍

package.json
```diff
@@ -67,11 +67,11 @@
   "license": "MIT",
   "homepage": "https://www.elastic.co/guide/en/apm/agent/rum-js/current/index.html",
   "devDependencies": {
-    "@angular-devkit/build-angular": "^12.0.4",
-    "@angular/cli": "^12.0.4",
+    "@angular-devkit/build-angular": "^13.1.2",
```
vigneshshanmugam, line 70: we have to test angular 13 as we have separate angular package for RUM. Lets not upgrade frameworks(react, angular) as part of this deps upgrade PR.

lucasfcosta:
Just as a piece of context, I've had to add the `json-chema` package to the list of packages for which we ignore licenses.

Previously, they already mentioned which license they used in their README and those seem to be the same licenses in the new LICENSE file.

From the old `README`:

```
Code is licensed under the AFL or BSD license as part of the Persevere 
project which is administered under the Dojo foundation,
and all contributions require a Dojo CLA.
```

I just wanted to confirm that it's fine to proceed given the new license seems to match what was previously in the README (in other words, we seem to have already been adhering to it anyway).

devcorpio:
LGTM! 🎉 thanks for moving this @lucasfcosta!

devcorpio (approved):

dev-utils/build.js
```diff
@@ -152,7 +152,10 @@ function getBabelConfig(bundleType, packageType) {
   let options = {
     comments: false,
     babelrc: false,
-    exclude: '/**/node_modules/**',
+    // We must exclude `parse5` from the packages we ignore
```
devcorpio, line 155: nit: 

What do you think about adding the exclusion rule in a variable? Then you could the comment only once, something like:


```
   // We must exclude `parse5` from the packages we ignore
   // because otherwise it would not be transpiled. Therefore,
   // it would break IE11.
   const babelExclusionRule = /node_modules\/(?!parse5).+/,
```

lucasfcosta: Great idea!

lucasfcosta:
As a note, @devcorpio (who was super helpful with pushing forward this PR 💖 ), also did local tests and sent me his results over Slack:

> and this a manual test with an agent built from your branch, sending events correctly in a website and no throwing any error
> Screenshot 2022-01-19 at 15 41 39

package.json
```diff
@@ -195,5 +196,9 @@
   "engines": {
     "node": ">=12 <=16",
     "npm": "6"
+  },
+  "resolutions": {
+    "semver-regex": "4.0.2",
+    "lodash": "4.17.21"
```
vigneshshanmugam, line 202: Lodash package version is not getting enforced for me locally, Even nuking the modules and reinstalling didnt help. 
```
      - "lodash": "4.17.21",
      + "lodash": "^4.16.6",
```

lucasfcosta: @vigneshshanmugam did you try `npm install` or `npm ci`? Can you send me the Node version you're using?

With `npm ci` the dependencies should indeed be frozen. With `install`, on the other hand, the dependencies down the tree will still be updated, as that's NPM's default behaviour.

Having said that, it just came to mind that then running `npm-force-resolutions` as a `preinstall` doesn't really make much sense as you should really be locking dependencies for `npm ci`, not for `install`. Therefore, I've added it as just another NPM script.

I was thinking about making it a pre-commit hook, what do you think?

vigneshshanmugam: I tried npm install. 

>Having said that, it just came to mind that then running npm-force-resolutions as a preinstall doesn't really make much sense as you should really be locking dependencies for npm ci, not for install

++ I think this is a better approach in general. 

> I was thinking about making it a pre-commit hook, what do you think?

Yeah that should be good.

vigneshshanmugam (approved):

PR #1132, opened 2022-01-10
chore: expose node and npm supported versions

devcorpio:
# Summary

We are making visible the node and npm versions currently supported in the RUM agent.

Note: The version for the .nvmrc file has been extracted from the Jenkins pipeline.


package.json
```diff
@@ -191,5 +191,9 @@
     "@elastic/apm-rum-core": "file:packages/rum-core",
     "@elastic/apm-rum-react": "file:packages/rum-react",
     "@elastic/apm-rum-vue": "file:packages/rum-vue"
+  },
+  "engines": {
+    "node": ">=12 <=16",
```
vigneshshanmugam, line 196: Can we also change the CI to use node 12 images? 

https://github.com/elastic/apm-agent-rum-js/blob/07bc85f7ea9fb036063faea9d03a4535ee18df7f/Jenkinsfile#L89

vigneshshanmugam (approved):

.nvmrc
```diff
@@ -0,0 +1 @@
+v12.22.8
```
vigneshshanmugam, line 1: Did we enforce this version somewhere?

package.json
devcorpio: good point!

.nvmrc
devcorpio: Currently, neither the documentation nor the code have any reference guiding/indicating which version of Node.js to use when working with the agent. Nevertheless, since this [PR](https://github.com/elastic/apm-agent-rum-js/pull/935) we are establishing version 12 for several steps of the pipeline. 

I extracted that version added there from logs of Jenkins. I added the exact version to have sync between development and pipeline. But maybe I'm being too picky, at the end of the day adding "12" might be enough since a tool such as nvm will install the latest version of 12 anyway... (there is a [12.22.9 version available](https://user-images.githubusercontent.com/15065076/148920046-9b70d28d-ebf6-4393-bf20-0c6d7e602ca9.png) since January 1st)

What do you think?

vigneshshanmugam: Gotcha, Yeah just 12 works for me.

PR #1138, opened 2022-01-13
chore: add spy on support for isolated functions

devcorpio:
# Summary

With the changes added in this PR, we will able to spy on functions exported from modules:

**before the change:**
``` 
// module.js

  export function hello() { }


  // test file
  import * as module from './module.js'
  
  const helloSpy = spyOn(module, 'hello') ---> throws "hello not declared writable or has no setter"
  
```

**after the change:**

``` 
  // test file
  import * as module from './module.js'
  import { spyOnFunction } from './utils/spy-function'

  const helloSpy = spyOnFunction(module, 'hello') ---> "works like a charm!"
  
```


packages/rum-core/test/bootstrap.spec.js
```diff
@@ -39,4 +41,12 @@ describe('bootstrap', function () {
     )
     window.performance.now = nowFn
   })
+
+  it('should call patchAll', () => {
+    const patchAllSpy = spyOnFunction(patcher, 'patchAll')
```
devcorpio, line 46: This test has been added in order to give an example of its usage. 👍

vigneshshanmugam (approved):
LGTM

PR #1141, opened 2022-01-18
chore: neutral naming for apm-agent-rum-js

v1v:
## What does this pull request do?

Neutral naming

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,7 +39,7 @@ Once your changes are ready to submit for review:
 3. Rebase your changes
 
    Update your local repository with the most recent code from the main repo,
-   and rebase your branch on top of the latest master branch.
+   and rebase your branch on top of the latest main branch.
    We prefer your initial changes to be squashed into a single commit.
    Later,
    if we ask you to make changes,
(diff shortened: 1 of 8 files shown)


vigneshshanmugam (approved):

devcorpio:
Hi @v1v,

I would like to share with you a behaviour that I noticed in the pipeline:

If a pipeline fails in the step "Test Puppeteer" because of a container being unhealthy, from that moment the subsequent reattempts triggered by the `test` command are going to fail too, with no exception.

The way I found to overcome that issue so far is by adding a new commit. For some reason doing that fixes the situation and there are not more failures caused by the unhealthy containers.

One example is this PR, as you can see the pipeline **[3](https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/PR-1141/3/consoleFull)** and **[4](https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/PR-1141/4/consoleFull)** failed with the unhealthy issue, but after adding an extra commit now it works

Do you know what could it be happening?

v1v:
@devcorpio , I just copied your comment in one of the internal discussions we have to track any issues regarding the CI. We can then to further analysis there

PR #1150, opened 2022-01-25
chore: update jenkins file

devcorpio:
# Summary

The jenkins file configuration for 4.x is outdated causing a pipeline error all the time

Btw, with these change if we update the docs again, the pipeline will only execute the commands related to docs rather than executing the whole flow (tests, etc)


devcorpio:
closing it, we don't need it

Issue #969, opened 2021-02-19
React: User Interaction only returns "Click - div"


reecebenson:
This issue is possibly related to the intended fix from PR elastic/apm-agent-rum-js#951, however applying the changes there do not resolve the below issue for me.

Whenever a target is clicked, the USER_INTERACTION event is fired. As an example button:

```html
<button name="applyFiltersBtn" class="ui icon primary left labeled button">
  <i aria-hidden="true" class="check icon"></i>
  Apply
</button>
```

This then fires the following log:

```
[Elastic APM] startTransaction(86115b370489ca7c, Click - div, user-interaction)
```

As you can see, the name of this transaction is titled `Click - div`, even though the (what looks like) intended behaviour of [`performance-monitoring.js:189`](https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/src/performance-monitoring/performance-monitoring.js#L191) is to grab the `name` attribute from the element, in this case `applyFiltersBtn`. I'd expect this transaction to be named `Click - button["applyFiltersBtn"]` and for this to be recorded in the APM under a "user-interaction", whilst capturing network traffic to include spans such as API requests, etc.

When doing some debugging, I logged out what was being passed to [`shouldInstrumentEvent`](https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum-core/src/common/patching/event-target-patch.js#L43) (or from elastic/apm-agent-rum-js#951, I have [this applied change](https://github.com/elastic/apm-agent-rum-js/blob/cd722fefc10dde2c29626144083e7b7a82477a97/packages/rum-core/src/common/patching/event-target-patch.js#L45), along with [the other changes from the PR](https://github.com/elastic/apm-agent-rum-js/pull/951/files)). When logging the `target` variable from the `shouldInstrumentEvent` function and also logging the `target` from the [getEventTargetSub](https://github.com/elastic/apm-agent-rum-js/blob/cd722fefc10dde2c29626144083e7b7a82477a97/packages/rum-core/src/performance-monitoring/performance-monitoring.js#L188), in all cases both parts of the aforementioned give back one of the following:

- `<div id="root">`
- `Window`
- `<react><react/>`.

It never gives back the _actual_ target of (in the example above) a Button. Is this intentional, if so, how would one implement a transaction to follow up on user interactions with buttons and log them correctly in APM, with React?

StefanoSega:
same issue ...
[...]

PR #1155, opened 2022-02-02
fix: identify on which element occurred a click event

devcorpio:
# Context

Web applications which rely on event delegation tend to declare click events through these ways:

```
window.addEventListener('click', //...)
document.addEventListener('click', //...)
document.querySelector("div.root").addEventListener('click', //...)

// more strategies ...

```

The agent is not handling  properly these cases. In our instrumentation we are not using the [target](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) to build the user interaction transaction. Instead what we are doing is to use the equivalent to [currentTarget](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget)
since we define the target as the element where the listener is being attached

# Summary 

The fix consists in capture the event generated by the browser and pick the target property from it, which from now on it will be the one that we will use to build the user-interaction transaction.

### DIFF
--- a/packages/rum-core/src/common/patching/event-target-patch.js
+++ b/packages/rum-core/src/common/patching/event-target-patch.js
@@ -133,11 +133,11 @@ export function patchEventTarget(callback) {
     existingTasks.push(task)
 
     function wrappingFn() {
-      // To identify properly the element on which the event occurred we need to make sure we distinguish between currentTarget and target
+      // To identify properly the element on which the event occurred we need to make sure
+      // we distinguish between currentTarget and target
       // currentTarget always refers to the element to which the event handler has been attached
       // target identifies the element on which the element occurred
       const [event] = arguments
-      task.currentTarget = event.currentTarget
       task.target = event.target
 
       callback(SCHEDULE, task)
--- a/packages/rum-core/src/common/patching/event-target-patch.js
+++ b/packages/rum-core/src/common/patching/event-target-patch.js
@@ -41,7 +41,7 @@ for (let i = 0; i < eventTypes.length; i++) {
 }
 
 function shouldInstrumentEvent(target, eventType, listenerFn) {
-  // EventTarget has many children. E.g. Element, Document and Window
+  // EventTarget has multiple children. E.g. Element, Document and Window
   // and since we are only handling clicks is not necessary to exclude others
   // such as XMLHttpRequest or AudioContext. Hence, we can rely on EventTarget being enough
   return (
--- a/packages/rum-core/src/common/patching/event-target-patch.js
+++ b/packages/rum-core/src/common/patching/event-target-patch.js
@@ -136,7 +136,7 @@ export function patchEventTarget(callback) {
       // To identify properly the element on which the event occurred we need to make sure
       // we distinguish between currentTarget and target
       // currentTarget always refers to the element to which the event handler has been attached
-      // target identifies the element on which the element occurred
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/common/patching/event-target-patch.js
```diff
@@ -131,6 +133,13 @@ export function patchEventTarget(callback) {
     existingTasks.push(task)
 
     function wrappingFn() {
+      // To identify properly the element on which the event occurred we need to make sure we distinguish between currentTarget and target
+      // currentTarget always refers to the element to which the event handler has been attached
+      // target identifies the element on which the element occurred
+      const [event] = arguments
+      task.currentTarget = event.currentTarget
+      task.target = event.target
```
devcorpio, line 141: In order to make the change more visible I added the currentTarget too as a property of our task. But I understand that we can remove it if we are not going to use it

What do you think?

packages/rum-core/test/common/event-target-patch.spec.js
```diff
@@ -151,26 +151,6 @@ describe('EventTargetPatch', function () {
         expect(count).toBe(4)
         expect(events.length).toBe(8)
       })
-
-      it('should not instrument non-Element targets', () => {
```
devcorpio, line 155: This is not required anymore since from now on the **click** will be also handled in document and window

packages/rum-core/src/common/patching/event-target-patch.js
vigneshshanmugam: We can remove the currentTarget, the comment conveys it clearly. Thanks

vigneshshanmugam (approved):
Great work @devcorpio, Much better than the other solution which we discussed on the thread. 

Could you double check the example posted on this issue once https://github.com/elastic/apm-agent-rum-js/issues/969 before we mark this as closed? Thanks.

devcorpio:
@vigneshshanmugam I also tested app using react, this is one of the apps tested


```
<!DOCTYPE html>
<html>
    <head>
        <div id="root"></div>
    </head>
    <body>
        <script crossorigin src="https://unpkg.com/react@17/umd/react.production.min.js"></script>
        <script crossorigin src="https://unpkg.com/react-dom@17/umd/react-dom.production.min.js"></script>
        <script crossorigin src="https://unpkg.com/babel-standalone@6/babel.min.js"></script>
        <script src="../elastic-apm-rum.umd.js" crossorigin></script>
        <script>
            elasticApm.init({
                serviceName: 'react-test-for-pr',
                serverUrl: 'http://apm-server:8200',
                logLevel: "debug",
            })
        </script>
        <script type="text/babel">
            ReactDOM.render(
                <div>
                    <h1>Hello! I'm testing the PR with a react app too</h1>
                    <div>
                        I'm creating a button with a custom name&nbsp;
                        <button name="pr-button-test">I'm a button meant to be clicked!</button>
                    </div>
                </div>,
                document.getElementById('root')
            );
        </script>
    </body>
</html>
```

Before the fix, the transaction was:

**in log**:
Screenshot 2022-02-03 at 14 11 08

**in ui**:
Screenshot 2022-02-03 at 14 12 37



After the fix, the transaction is:

**in log**:
Screenshot 2022-02-03 at 14 19 15

**in ui**:
Screenshot 2022-02-03 at 14 20 52


🎉

vigneshshanmugam:
@devcorpio Awesome 🎉 , Let's get this in.

PR #1187, opened 2022-03-16
feat: add pageLoadParentId configuration

devcorpio:
# Context

One of the requirements for [Crosslinking Synthetics with APM](https://github.com/elastic/synthetics/issues/265) issue was to allow the RUM agent the creation of the `page-load` transaction as child of the one created in Synthetics.

The main goal of that is to provide visibility into how the synthetic journeys are executed and what actions are happening inside every step. This [image](https://user-images.githubusercontent.com/3902525/116486823-ccd0f200-a88e-11eb-95da-527940a74081.png) extracted from the issue linked above it's a great example.

# Summary

We expose a new [agent configuration](https://www.elastic.co/guide/en/apm/agent/rum-js/current/configuration.html) option named `pageLoadParentId` whose value will be the id of the transaction that we want to establish as **parent** of the `page-load` transaction.  Before this PR `page-load` transaction was always treated as the **root** transaction, now it will be possible to set a parent if needed thanks to this new config.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -183,6 +183,15 @@ NOTE: After each flush of the queue, the next flush isn't scheduled until an ite
 
 This option overrides the page load transactions trace ID.
 
+[float]
+[[page-load-parent-id]]
+=== `pageLoadParentId`
+
+* *Type:* String
+
+This option allows the creation of the page load transaction as child of an existing one. By default,
+the agent treats it as the root transaction.
+
 
 [float]
 [[page-load-sampled]]
--- a/packages/rum-core/src/common/compress.js
+++ b/packages/rum-core/src/common/compress.js
@@ -231,6 +231,7 @@ export function compressTransaction(transaction) {
 
   const tr = {
     id: transaction.id,
+    pid: transaction.parent_id,
     tid: transaction.trace_id,
     n: transaction.name,
     t: transaction.type,
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -447,6 +447,7 @@ export default class PerformanceMonitoring {
 
     const transactionData = {
       id: transaction.id,
+      parent_id: transaction.parentId,
       trace_id: transaction.traceId,
       session: transaction.session,
       name: transaction.name,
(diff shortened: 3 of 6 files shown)


vigneshshanmugam (commented):
Thanks @devcorpio Changes LGTM, But we need to do few more things
1. Add the limit check to truncation logic that should truncate if the field is past 1024 bytes- https://github.com/elastic/apm-agent-rum-js/blob/434d6213a46c035680821b7843a15f628217ec64/packages/rum-core/src/common/truncate.js#L95
2. Document the pageloadParentId in the config option along with other trace options- https://github.com/elastic/apm-agent-rum-js/blob/main/docs/configuration.asciidoc#pageloadtraceid

I would also like to test with the Synthetics APM PR before we close this down. Thanks!

devcorpio:
> Thanks @devcorpio Changes LGTM, But we need to do few more things
> 
> 1. Add the limit check to truncation logic that should truncate if the field is past 1024 bytes- https://github.com/elastic/apm-agent-rum-js/blob/434d6213a46c035680821b7843a15f628217ec64/packages/rum-core/src/common/truncate.js#L95
> 2. Document the pageloadParentId in the config option along with other trace options- https://github.com/elastic/apm-agent-rum-js/blob/main/docs/configuration.asciidoc#pageloadtraceid
> 
> I would also like to test with the Synthetics APM PR before we close this down. Thanks!

The truncation logic it's already working. The function ```truncate``` default limit is 1024 as we can see [here](https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/common/truncate.js#L122). 

Something that caught my attention is that if I replace ```parent_id: true``` with ```parent_id: [KEYWORD_LIMIT, true]``` and the value for `parent_id` is empty the `truncate logic` will set the 'N/A' value which in my opinion feels a bit weird value for a `parent_id`. 

One way to avoid that is setting ```parent_id: [KEYWORD_LIMIT, false]``` but seeing that the value was already being truncated, I would advocate for letting the config as it is. What do you think?

vigneshshanmugam:
> One way to avoid that is setting parent_id: [KEYWORD_LIMIT, false] but seeing that the value was already being truncated, I would advocate for letting the config as it is. What do you think?

No, we dont need to set N/A and that flag exists only when we cannot send a empty string to the APM server which are marked as `required: true` in the server spec.

vigneshshanmugam (approved):
LGTM

I will try to test with Synthetics package before we do a release of this one.

PR #1188, opened 2022-03-22
chore: test for partial page-load transaction

devcorpio:
# Summary

Test that demonstrates that if a page becomes backgrounded all the data gathered until that moment will be sent even when the load event from the browser has not been triggered.

### DIFF
--- a/packages/rum/test/specs/apm-base.spec.js
+++ b/packages/rum/test/specs/apm-base.spec.js
@@ -36,6 +36,10 @@ import {
 import { TRANSACTION_END } from '@elastic/apm-rum-core/src/common/constants'
 import { getGlobalConfig } from '../../../../dev-utils/test-config'
 import Promise from 'promise-polyfill'
+import {
+  hidePageSynthetically,
+  setDocumentVisibilityState
+} from '@elastic/apm-rum-core/test'
 
 var enabled = bootstrap()
 const { serviceName, serverUrl } = getGlobalConfig('rum').agentConfig
@@ -43,10 +47,30 @@ const { serviceName, serverUrl } = getGlobalConfig('rum').agentConfig
 describe('ApmBase', function () {
   let serviceFactory
   let apmBase
+  let originalAddEventListener
+  let unobserveVisibilitychange = function noop() {}
 
   beforeEach(function () {
     serviceFactory = createServiceFactory()
     apmBase = new ApmBase(serviceFactory, !enabled)
+
+    // Each time apmBase is initialized the visibilitychange event is attached to window
+    // To avoid side effects when testing we should detach it after each execution
+    originalAddEventListener = window.addEventListener
+    window.addEventListener = (type, listener, options) => {
+      if (type === 'visibilitychange') {
+        unobserveVisibilitychange = () => {
+          window.removeEventListener(type, listener, options)
+        }
+      }
+
+      originalAddEventListener(type, listener, options)
+    }
+  })
+
+  afterEach(() => {
+    window.addEventListener = originalAddEventListener
+    unobserveVisibilitychange()
   })
 
   it('should report whether agent is active or inactive', () => {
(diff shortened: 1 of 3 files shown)


packages/rum/test/specs/apm-base.spec.js
```diff
@@ -36,17 +36,41 @@ import {
...
+  let unobserveVisibilitychange = function noop() {}
 
   beforeEach(function () {
     serviceFactory = createServiceFactory()
     apmBase = new ApmBase(serviceFactory, !enabled)
+
+    // Each time apmBase is initialized the visibilitychange event is attached to window
+    // To avoid side effects when testing we should detach it after each execution
+    originalAddEventListener = window.addEventListener
+    window.addEventListener = (type, listener, options) => {
```
devcorpio, line 60: offtopic curiosity:

In the future if needed this logic might be extended to intercept all the events occurred while testing and detach them after each execution. It could be a global behaviour for all our tests :)

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -83,6 +107,43 @@ describe('ApmBase', function () {
...
+    setDocumentReadyState('loading')
+
+    apmBase.init({ serviceName, serverUrl })
+
+    var tr = apmBase.getCurrentTransaction()
+
+    expect(tr.name).toBe('Unknown')
+    expect(tr.type).toBe(PAGE_LOAD)
+
+    hidePageSynthetically('visibilitychange')
```
devcorpio, line 132: Despite `visibilitychange` being the default value of `hidePgeSynthetically` I added it explicitly because my intent is to help developers reading this to understand that is entwined with the beforeEach logic

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -36,17 +36,41 @@ import {
 import { TRANSACTION_END } from '@elastic/apm-rum-core/src/common/constants'
 import { getGlobalConfig } from '../../../../dev-utils/test-config'
 import Promise from 'promise-polyfill'
+import {
+  hidePageSynthetically,
+  setDocumentVisibilityState
+} from '@elastic/apm-rum-core/test'
```
vigneshshanmugam, line 42: does this work? now the test of packages/rum should build packages/rum-core each time for tests, I dont think this would work with current build system. We only build on packages/rum/src changes.

vigneshshanmugam (commented):
Code changes looks good to me except for the test build script.

packages/rum/test/specs/apm-base.spec.js
devcorpio: Thanks for asking this! When I was doing this that was precisely my doubt, it felt a bit weird to import something from another package, but at that moment it's when I saw this line in the very same file:

```import { TRANSACTION_END } from '@elastic/apm-rum-core/src/common/constants'```. It turns out that is something that we are already doing. And when I started looking at other specs I saw that it's our way of doing across all the project, examples:

- https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum/test/specs/index.spec.js
- https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-vue/test/specs/index.spec.js
- https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-react/test/specs/get-with-transaction.spec.js
- and many more

And yes, it works. The current build system relies on `lerna` for this. For instance, when you execute the `npm run build` corresponding to the main package.json, the order of builds is going to be the following one:

- apm-rum-core
- apm-rum
- apm specific frameworks

This is happening because in the package.json of each package we are saying which packages depend on each other:

- **apm-rum** depends on **apm-rum-core**
- **apm-rum-react** depends on **@elastic/apm-rum**
- **apm-rum-vue** and **apm-rum-angular**  depends on **@elastic/apm-rum** and **@elastic/apm-rum-core**

For instance, **lerna** guarantees that the build for **apm-rum** will not start until the one from **apm-rum-core** is finished, since it's a dependency.

Hope this clarify your question.

--
Having said that, in future projects where we could use lerna with multiple packages, we should think if this is the way to proceed since now we the packages are not really independent. We will trade-offs on those discussions :)

vigneshshanmugam: > import { TRANSACTION_END } from '@elastic/apm-rum-core/src/common/constants'

Was pretty sure this would work as we build `src` all the time and As you mentioned lerna does guarantee the order based on the dependency tree. But did not knew `test` directory was also taken in to account which feels a bit wrong. But all good 👍🏽

Issue #1162, opened 2022-02-08
the User-centric Metrics "Largest contentful paint (LCP)" is not correct


rainydayDY:
I have a question that page-load trigger too early，if my website render too lower，the 'largest-contentful-paint' doesn't trigger, or only trigger once, but the page-load trigger, the 'largest-contentful-paint' time will not correct
`const sendPageLoadMetrics = () => { // to make sure PerformanceTiming.loadEventEnd has a value setTimeout(() => tr.removeTask(PAGE_LOAD)) }`
There's something wrong with that logic

devcorpio:
Hi @rainydayDY,

Thanks for contacting us!

I have a couple questions about **"Largest contentful paint (LCP)" is not correct**

1. Do you mean that the LCP value is not calculated? (so no LCP entry)
2. Or do you mean that the calculated time should be different than the one being calculated?

I have a couple questions related to agent and environment:

1. What version of the RUM agent are you using?
2. Could you share with us the webpage you are debugging?


--

On the other hand, I have reproduced a case where the LCP is not being reported by the RUM agent:

If the largest content of a page is an image there are times where we are **not** reporting LCP. This particular case happens regardless of how fast the page-load triggers. And as you are mentioning seems related to the logic that you are pointing out. 
It seems we stop the performance observer from receiving any performance entry before getting the LCP. It requires further investigation.

Is this particular case related to the one you are bringing to us?

rainydayDY:
Hi, devcorpio,
Thanks for your reply！
**（_Note: github can't upload pictures, so I posted the picture link_）**
First of all, I don't think the case you mentioned is the same as what I said, but I'm not too sure，Because, when I watched discovery on kibana, 0.06% of pageload requests were reported to LCP；

【https://p3-juejin.byteimg.com/tos-cn-i-k3u1fbpfcp/90611a9caf0d4f8e976b108968c238e3~tplv-k3u1fbpfcp-watermark.image?】
【https://p1-juejin.byteimg.com/tos-cn-i-k3u1fbpfcp/1c82f471469942cebfc8493bd5b28221~tplv-k3u1fbpfcp-watermark.image?】

The uncertain point is that our page looks like the following, there are pictures in it, and the background image can be said to be the largest part of the current page；

【https://p3-juejin.byteimg.com/tos-cn-i-k3u1fbpfcp/1997f0f66f3745f88ae47e7220f54b6b~tplv-k3u1fbpfcp-watermark.image?】

[...]

PR #1190, opened 2022-03-28
fix: report LCP score properly

devcorpio:
# Context

Two different [issues](https://github.com/elastic/apm-agent-rum-js/issues/1162) related to LCP have been spotted:

- agent not reporting LCP at all when it should have.
- agent reporting a wrong LCP score value.

A little summary of how the flow related to page-load transaction is:

1. browser triggers **load** event.
2. agent ends the page **load** transaction and stops observing performance entries (such as LCP) for that transaction.
3. a beacon with all the information is sent

The point 2 is where the **key** of the issue is.

# Summary

Added a delay of **1 second** before sending the page load metrics to make sure we gather the proper data. 

# Why this approach

The delay is of the utmost importance:

When it comes to the page load event we need to take into account he following statement from the [spec]((https://www.w3.org/TR/uievents/#event-type-load)):

"_The browser dispatch the load event when the DOM implementation finishes loading the resource (such as the document) and any dependent resources (such as images, style sheets, or scripts)._"

After the page load happens, the browser might still need more time to render particular elements (e.g. large block of text, images, etc). It's at that moment when a new LCP [spec](https://w3c.github.io/largest-contentful-paint/#sec-add-lcp-entry) entry might be reported by the browser.

Unfortunately, without the added delay that might happen after the page load transaction has finished that's why we would be reporting wrong data.

## Example

The following website have been used for testing this

Screenshot 2022-03-28 at 13 35 38

**Before the change**:

Screenshot 2022-03-28 at 13 54 00

At that image we can see how the page load time is **89.5** and the last LCP entry captured by the performance observer **outside** of the agent shows renderTime **371.9**.  The performance observer of the agent is not capturing that because the transaction was ended just after the page load as I explained above in the PR.

**After the change**:

Screenshot 2022-03-28 at 13 59 32

At that image we can see how the page load time is **117.70000000111759** and the last LCP entry captured by the performance observer **outside** AND **within** of the agent shows renderTime **410**.  This time and thanks to the delay added, that LCP entry will observed by the agent and then will be included in the **page-load** transaction.

# Additional notes:

The "send events when user leave the page" https://github.com/elastic/apm-agent-rum-js/pull/1146 issue makes sure that the info gathered until that moment will be sent too.

@vigneshshanmugam. Since this issue is related to LCP maybe we could do this logic in browsers which support it. And in other browsers keep doing what we were doing. Something that could support that statement: not all the browsers support the fetch keep alive logic, in the worst case scenario we could be losing some info (since the pageload now will have an extra 1 second). What do you think?

Tasks:

- Create test somehow for this scenario to avoid regressions []


vigneshshanmugam (commented):
I gave it more thought on this approach, Some points that comes to my mind

- Users are going to see increased page load transaction duration as a result of this change. 
- User Experience page suddenly going to show large numbers as a result for `Page load duration` and `Front end time`. 

I propose we do the following
- Monitor for interesting events (XHR, Network requests to fetch static assets or LCP events)
- When interesting resources are detected within the 1 second timeout, we wait until they complete and that would also be included as a result in the page load transaction.
- If nothing interesting happens during the additional timeout, we fire the
transaction end event after the timeout. The event end time will be adjusted to the time of the last tracked resource.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -498,12 +498,13 @@ describe('TransactionService', function () {
     const lastSpan = transaction.startSpan('last-span-name', 'last-span')
     lastSpan.end()
     longSpan.end()
-    longSpan.end += 500
+    longSpan._end += 500
```
devcorpio, line 501: This test was giving a false positive. We were adding the +500 to the wrong property and we were not taking into account that `adjustTransactionTime` also truncates spans. 

The truncate point is important because it mutates the spans being used in the test. 
After adjustTransactionTime execution **transaction._end** and **longSpan_end** were equal.  That's why the assertion was still passing properly.

Removing the code that was intended to be tested was not affecting the results of the test at all.

Besides setting the +500 to the correct property I also assigned the longSpan end to a new variable (which is the one now used in the assertion) so we make sure the span truncation does not affect the results of the test anymore.

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -101,6 +103,10 @@ describe('ApmBase', function () {
       expect(endedTr).toEqual(tr)
       expect(document.readyState).toBe('complete')
       expect(tr.detectFinish).toHaveBeenCalled()
+      expect(window.setTimeout).toHaveBeenCalledWith(
+        jasmine.any(Function),
+        PAGE_LOAD_DELAY
```
devcorpio, line 108: btw, since this test is not faking setTimeout nor using system.clock the duration of this test will be affected by the real setTimeout. Hence, this test will last an extra second. I think it's not a big deal if the delay remains at that number, but we should be careful with it if in the future we decide to increase the delay for whatever reason.

What do you think? Do you think that we should use the system clock or let this as it is?

devcorpio:
@vigneshshanmugam the only point missing in this PR is the adaptation of the e2e mentioned yesterday in the call. All the changes in the agent with the new behaviour are done in case you want to review again 👍

packages/rum-core/test/utils/apm-server-mock.js
```diff
@@ -82,6 +82,16 @@ class ApmServerMock {
       useMocks ? () => Promise.resolve() : undefined
     )
 
+    this.resetMock = function () {
```
devcorpio, line 85: This change and the one done in waitForApmServerCalls permit us to use waitForApmServerCalls more than once in the same test if required. Like in https://github.com/elastic/apm-agent-rum-js/pull/1190/files#diff-6c17ea89599249cc59d7b2195a384042aa5c7a12ed67601e5ca04119b8b798c3

packages/rum/package.json
```diff
@@ -32,7 +32,7 @@
     "test:unit": "npm run script runUnitTests packages/rum",
     "test:e2e:supported": "npm run script runE2eTests packages/rum/wdio.conf.js",
     "test:e2e:failsafe": "npm run script runE2eTests packages/rum/wdio-failsafe.conf.js",
-    "test:sauce": "npm run script runSauceTests packages/rum true test:unit test:e2e:supported test:e2e:failsafe",
+    "test:sauce": "npm run script runSauceTests packages/rum true build:e2e test:unit test:e2e:supported test:e2e:failsafe",
```
devcorpio, line 35: Make sure we build the e2e before helps the developer experience when making changes in the test. (We were already doing the build in the other packages containing e2e btw)

packages/rum-react/test/e2e/with-router/general.e2e-spec.js
```diff
@@ -79,14 +79,10 @@ describe('General usecase with react-router', function () {
       'expected manual component to be rendered'
     )
 
-    const { sendEvents } = waitForApmServerCalls(0, 2)
+    const { sendEvents } = waitForApmServerCalls(0, 1)
     const { transactions } = sendEvents
 
-    const pageLoadTransaction = transactions[0]
```
devcorpio, line 85: This is not necessary this step triggers a soft navigation. (without resetting the mock as we are doing right now we would not be able to doing that, so another benefit)

packages/rum-vue/test/specs/index.spec.js
```diff
@@ -196,7 +196,10 @@ describe('APM route hooks', () => {
     await router.isReady()
     router.afterEach(() => {
       const tr = apm.getCurrentTransaction()
-      expect(tr.spans[0].name).toBe('before-route-enter')
```
devcorpio, line 199: This test was failing in the pipeline but not in the macbook. Debugging this in another laptop (M1 it's too much powerful 😄 )  helped me to understand where the problem was:

Since we are adding the 1 second delay, we cannot assure anymore that the first span will be the one expected. 

Karma during its lifecycle may trigger long tasks,  because of that the `spans[0]` was not the expected and it was the longtask instead

 Screenshot with longtask logged during test execution:

longtask-screenshot

`context.html` is the value of the src of Karma's iframe

vigneshshanmugam (commented):
Looking good overall, Great work @devcorpio.

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -498,17 +499,99 @@ describe('TransactionService', function () {
...
+      span.end()
+      span._end += 20
+
+      const externalSpan = transaction.startSpan('span-name', 'external')
+      externalSpan.end()
+
+      const lcp = 10
+      transaction.addMarks({
+        agent: {
+          largestContentfulPaint: lcp
```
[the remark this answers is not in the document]
devcorpio: Ups!, this was a leftover!, the chosen one was `metrics.lcp`, thanks for spotting it 🎉

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -404,20 +405,33 @@ class TransactionService {
...
+
+    // Before ending the page-load transaction we are adding a delay to monitor events such as LCP and network requests.
+    // We need to make sure that we are not adding that extra time to the transaction end time
+    // if nothing has been monitored or if the last monitored event end time is less than the delay.
+    if (transaction.type === PAGE_LOAD) {
+      const lcp = metrics.lcp || 0
+      const latestXHRSpan = getLatestXHRSpan(spans) || {}
+      const latestXHRSpanEnd = latestXHRSpan._end || 0
+
+      transaction._end = Math.max(latestSpanEnd, latestXHRSpanEnd, lcp)
```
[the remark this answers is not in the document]
devcorpio: Really good point. 

It's true that we have 'Fire "load" event' as a span and the value it is equal to `domComplete`. But, it seems that **sometimes** the span 'Fire "load" event' is not generated (although `domComplete` has always value and we could pick this then).

Were you aware of this behaviour?

devcorpio:
@vigneshshanmugam the domComplete logic and changes related to your last comments are done. If you approve this I will release a new version of the agent

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -413,10 +413,17 @@ class TransactionService {
     // if nothing has been monitored or if the last monitored event end time is less than the delay.
     if (transaction.type === PAGE_LOAD) {
       const lcp = metrics.lcp || 0
+      const agentMarks = transaction.getAgentMarks() || {}
+      const domComplete = agentMarks.domComplete || 0
```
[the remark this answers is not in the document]
devcorpio: done 🕺

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -404,20 +406,39 @@ class TransactionService {
...
+    const latestSpan = getLatestNonXHRSpan(spans) || {}
+    const latestSpanEnd = latestSpan._end || 0
+
+    // Before ending the page-load transaction we are adding a delay to monitor events such as LCP and network requests.
+    // We need to make sure that we are not adding that extra time to the transaction end time
+    // if nothing has been monitored or if the last monitored event end time is less than the delay.
+    if (transaction.type === PAGE_LOAD) {
+      const transactionEndWithoutDelay = transaction._end - PAGE_LOAD_DELAY
+      const lcp = metrics.lcp || 0
+      const latestXHRSpan = getLatestXHRSpan(spans) || {}
```
vigneshshanmugam, line 418: Do we need this still? I forgot why we did added this check
1. Since the transaction always waits for the XHR tasks which are blocking, we are covered in this scenario which means transaction.end will always be greater than XHR span end. 

Let me know if I got it wrong.

vigneshshanmugam (approved):
LGTM, Thanks @devcorpio for the extensive tests.

packages/rum-core/src/performance-monitoring/transaction-service.js
devcorpio: This check is making sure that the XHR tasks with the following characteristics are handled:

1. triggered during the page load delay
2. finished **before** the delay

Let's imagine a scenario with this timeline

1. Non XHR span               90ms
2. LCP entry                      120 ms
3. page fires load event   135ms
4. 1 second delay is set

During that 1-second window where we observe for activity, there are two possible scenarios related to XHR spans that need to be handled:

<ins>**Scenario A**</ins> (the one that you are commenting, where the extra check it's not required)

A request that will last 1200ms have been observed. The timeline then it would be:

5. delay finishes.            1135ms
6. request finishes.        1335ms (and here the transaction ends, with the proper duration as you mentioned)

Screenshot 2022-04-20 at 11 14 40



<ins>**Scenario B**</ins> (the one that explains the extra check)

A request that will last 300ms have been observed. The timeline then it would be:

5. request finishes.        435ms
6. delay finishes.            1135ms (and here the transaction ends, with the **wrong** duration)

In such scenario we have there a gap of 700ms that we need to adjust, that's why we have that check too, you can see it in the image below

Screenshot 2022-04-20 at 11 15 56

I will add a comment in the codebase explaining the why of this in a concise way

vigneshshanmugam: Gotcha, that makes sense. Thanks for the details @devcorpio 👍🏽 Lets get this in.

PR #1201, opened 2022-04-18
ci: use withNode and use retryWithSleep

v1v:
### What

- ~~Bump versions to `8.1` and `7.17`~~ requires https://github.com/elastic/apm-agent-rum-js/pull/1193
- Fix version bump (since `7.x` is not anymore a pattern to be searched)
- Use [`retryWithSleep`](https://github.com/elastic/apm-pipeline-library/blob/main/vars/retryWithSleep.groovy) to simplify the pipeline
- Use [`withNode`](https://github.com/elastic/apm-pipeline-library/blob/main/vars/withNode.groovy) to assign unique agents.

### Why

Trying to solve the issue with the `Test Puppeteer` when the subsequent reattempts triggered by the test command fail. I'm still confused since `Test Puppeteer` does not cache anything in the CI workers, and the `Gobld Jenkins plugin` ensures the workers are not reused.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -120,23 +120,23 @@ pipeline {
           matrix {
             agent { label 'linux && immutable' }
             axes {
-                axis {
-                  name 'ELASTIC_STACK_VERSION'
-                  // The below line is part of the bump release automation
-                  // if you change anything please modifies the file
-                  // .ci/bump-stack-release-version.sh
-                  values '8.1.0-SNAPSHOT', '7.16.0'
-                }
-                axis {
-                  name 'SCOPE'
-                  values (
-                    '@elastic/apm-rum',
-                    '@elastic/apm-rum-core',
-                    '@elastic/apm-rum-react',
-                    '@elastic/apm-rum-angular',
-                    '@elastic/apm-rum-vue',
-                  )
-                }
+              axis {
+                name 'ELASTIC_STACK_VERSION'
+                // The below line is part of the bump release automation
+                // if you change anything please modifies the file
+                // .ci/bump-stack-release-version.sh
+                values '8.1.0-SNAPSHOT', '7.16.0'
+              }
+              axis {
+                name 'SCOPE'
+                values (
+                  '@elastic/apm-rum',
+                  '@elastic/apm-rum-core',
+                  '@elastic/apm-rum-react',
+                  '@elastic/apm-rum-angular',
+                  '@elastic/apm-rum-vue',
+                )
+              }
             }
             stages {
               stage('Scope Test') {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -485,14 +485,12 @@ def runScript(Map args = [:]){
       "APM_SERVER_URL=http://apm-server:8200",
       "APM_SERVER_PORT=8200",
       "GOAL=${goal}"]) {
-      retry(2) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -585,15 +580,18 @@ def runAllScopes(Map args = [:]) {
...
+
   withGithubNotify(context: "Test ${scope} - ${stack} - ${mode}", tab: 'tests') {
-    runScript(
-      label: "${scope}",
-      stack: "${stackVersion}",
-      scope: "${scope}",
-      goal: 'test'
-    )
+    if (allocateNode) {
+      withNode(labels: 'linux && immutable', sleepMax: 5, forceWorspace: true, forceWorker: true){
```
v1v, line 589: It requests the ephemeral worker with an unique ID (that's something the `Gobld` provisioner does under the hood)

image

Jenkinsfile
```diff
@@ -501,8 +498,6 @@ def runScript(Map args = [:]){
             sh(label: "Run tests: Elastic Stack ${stack} - ${scope} - ${env.MODE}", script: '.ci/scripts/test.sh')
           }
         }
-      } catch(e) {
-        throw e
```
vigneshshanmugam, line 505: question: do we swallow errors now or is it handled somewhere else?

vigneshshanmugam (approved):
LGTM

Jenkinsfile
v1v: `try/finally` works fine as long as the body in the `try` closure can handle the exception. In this case, I decided to simplify the try/catch/finally since the `catch` does nothing but throwing the error, so it's a bit more idiomatic, therefore this should not shallow any errors

devcorpio:
@v1v just in case it helps you to test this change somehow: 
the [build]((https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/PR-1202/1/consoleFull)) of this [PR](https://github.com/elastic/apm-agent-rum-js/pull/1202) is failing because of unhealthy containers. If we execute the command to re-run the tests we will have a lot of chances of having the issue reproduced

kuisathaverat (approved):

v1v:
> @v1v just in case it helps you to test this change somehow:
the build of this https://github.com/elastic/apm-agent-rum-js/pull/1202 is failing because of unhealthy containers. If we execute the command to re-run the tests we will have a lot of chances of having the issue reproduced

I'm running the build again, though, it uses ephemeral workers and the previous build failed yesterday, so there is no cache or state between builds.

As far as I see the [healthcheck failed](https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/PR-1202/1/artifact/docker-info/@elastic/apm-rum-8_0_0-SNAPSHOT/9a9bf24dfd3a-inspect.json):

image

I wonder if the issue is always happening with the docker image for `ElasticSearch:8.0.0-SNAPSHOT`, or whether it happened with some other and older versions. I'm somehow inclined to thing there is a bug with `ElasticSearch:8.0.0-SNAPSHOT`. `8.0.0-SNAPSHOT` is not built anymore since Feb 2022:

```
$ docker inspect docker.elastic.co/elasticsearch/elasticsearch:8.0.0-SNAPSHOT | grep org.label-schema.build-date
                "org.label-schema.build-date": "2022-02-10T15:45:39.792423829Z",
```

Issue #1196, opened 2022-04-05
e2e react flaky test related to user-interaction threshold


devcorpio:
# Context

A while ago it was noticed that one of the [e2e test](https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-react/test/e2e/with-router/general.e2e-spec.js#L68) of the react package was behaving flakily.

The degree of flakiness varied depending on where its execution was triggered, that is: locally or in the pipeline

Executing it locally tends to fail more than in the pipeline, although the two links below are an example of failures that took place in the pipeline: (there are more instances)

- https://apm-ci.elastic.co/job/apm-agent-rum/job/apm-agent-rum-mbp/job/main/28/execution/node/1513/log/
- https://apm-ci.elastic.co/blue/rest/organizations/jenkins/pipelines/apm-agent-rum/pipelines/apm-agent-rum-mbp/pipelines/PR-1146/runs/19/nodes/1254/steps/1546/log/?start=0

To locate the error search for **'ManualComponent'** on the logs

## Why we didn't create the issue back then:

Back then there were two reasons that caused not to focus on this:

- the pipeline retries until 3 times, and **most of the time** the pipeline does not get stuck for this.
- we just came of updating all the [e2e dependencies](https://github.com/elastic/apm-agent-rum-js/pull/1118) and we decided that to spend more time on debugging something that was not blocking the pipeline was not the right investment of time

## Why we are creating the issue now:

- Because while working of the LCP [issue](https://github.com/elastic/apm-agent-rum-js/pull/1190) that test required changes too (adapt it to new logic).  Unfortunately, the was failing often and it was being a bit painful to figure out if the failure was coming for the change of logic or for the known flakiness.
- The reason behind the flakiness has been found 🎉 

## Why is the test flaky:

Context: the test is checking that if the react router loads a component lazily (dynamically) we are getting the proper spans

**The problem lies on timings:** 

The click transactions have a 300ms threshold, during that period among other things they can:

1. (this is the one to "blame") capture **resource** and user timing spans (non blocking spans,: e.g. the transaction will be finished even if there is a resource loading in progress).
2. capture interesting things such as xhr/fetch requests (blocking spans, meaning: transaction will not be finished until request is done)
3. Have their name and type redefined. E.g. a route-change happens after the click

[...]

PR #1204, opened 2022-04-19
feat(rum): override threshold when redefining transaction

devcorpio:
Relates to: https://github.com/elastic/apm-agent-rum-js/issues/1196

With this change we make sure that the reuseThreshold is overridden every time a transaction is redefined. This change will benefit users browsing with slow networks in cases like the one explained [here](https://github.com/elastic/apm-agent-rum-js/issues/1196#issuecomment-1097908486)

https://user-images.githubusercontent.com/15065076/164041735-b5299857-3959-4c04-aaf7-a495b3eca118.mov

@vigneshshanmugam I added feat(rum) in the commit. If you think that this should be a **"chore:"** instead let me know

### DIFF
--- a/packages/rum-core/src/performance-monitoring/transaction.js
+++ b/packages/rum-core/src/performance-monitoring/transaction.js
@@ -84,6 +84,10 @@ class Transaction extends SpanBase {
     }
 
     if (options) {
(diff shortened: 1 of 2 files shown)


vigneshshanmugam (approved):
LGTM

Issue #1210, opened 2022-04-27
aborted xhr/fetch requests should set transaction outcome to unknown


devcorpio:
# Context
At present, the **xhr** and **fetch** instrumentations create http transactions with a successful outcome even when the instrumented requests have been aborted.

Aborted requests falls under two categories:
1.  programmatic abort
2. user aborts(e.g. tab closing)

# Summary
The goal of this issue is to start setting the outcome of the aforementioned transactions to **unknown** as per agreed in a previous discussion.
The idea of setting the outcome to **aborted** rather than **unknown** was also mentioned. But given that it would require changes in APM server we have decided to stay with **unknown**. Apart from that, it will be easily recognisable in the UI, since unknown will be only visible with aborted requests (at least in RUM context)

# Notes
At this moment, the APM UI doesn't show the badge with the value "unknown". An additional issue will be created to tackle this.

PR #1211, opened 2022-04-27
feat(rum): set unknown outcome for aborted http requests

devcorpio:
# Summary

Relates to: https://github.com/elastic/apm-agent-rum-js/issues/1210

The goal of this change is to set the outcome of transactions/spans with aborted requests to **unknown**

@vigneshshanmugam:

There is some **drawback** that I would like to highlight related to Fetch API: 

At present, it's not possible to tell if the request has aborted with a timeout or not. It means, that we would be treating as "unknown" (rather than a failure) the ones that are caused by a timeout. Although, the agent currently consider them as a **successful** ones, so we could consider this as an improvement? (thinking out loud  🤔 )

There are upcoming browser APIs that will address that, such as: [AbortSignal.timeout](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout). For instance, it will be available in [Chrome 103](https://chromestatus.com/feature/5768400507764736)

WDYT about this drawback?

**Note**: As explained in the issue's description, we will need to make a change in Kibana UI, since the unknown badge is not created. But at least the information will be visible in the metadata

**Note 2**: I ended up creating the ticket as an enhancement and not as a bug. Let me know if you disagree with that. 
I created this as an enhancement, because I feel that this is something that was discussed but not developed rather than something that is happening "accidentally"

### DIFF
--- a/packages/rum-core/src/common/patching/fetch-patch.js
+++ b/packages/rum-core/src/common/patching/fetch-patch.js
@@ -97,6 +97,7 @@ export function patchFetch(callback) {
         error => {
           reject(error)
           scheduleMicroTask(() => {
+            task.data.aborted = isAbortError(error)
             task.data.error = error
             invokeTask(task)
           })
@@ -105,4 +106,8 @@ export function patchFetch(callback) {
       globalState.fetchInProgress = false
     })
   }
+}
+
+function isAbortError(error) {
+  return error && error.name === 'AbortError'
 }
--- a/packages/rum-core/src/performance-monitoring/performance-monitoring.js
+++ b/packages/rum-core/src/performance-monitoring/performance-monitoring.js
@@ -48,6 +48,7 @@ import {
   USER_INTERACTION,
   OUTCOME_FAILURE,
   OUTCOME_SUCCESS,
+  OUTCOME_UNKNOWN,
   QUEUE_ADD_TRANSACTION
 } from '../common/constants'
 import {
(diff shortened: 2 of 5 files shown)


packages/rum-core/src/common/patching/fetch-patch.js
```diff
@@ -106,3 +107,7 @@ export function patchFetch(callback) {
     })
   }
 }
+
+function isAbortError(error) {
+  return error && error.name === 'AbortError'
```
devcorpio, line 112: All the browsers that we support (the ones from our E2E test) and are compatible with fetch handle this properly. I'm highlighting this, because in MDN 'AbortError' is described as [experimental](https://developer.mozilla.org/en-US/docs/Web/API/DOMException). The legacy ways were: **ABORT_ERR** or code **20**

packages/rum-core/src/performance-monitoring/performance-monitoring.js
```diff
@@ -337,12 +338,14 @@ export default class PerformanceMonitoring {
...
         let outcome
-        if (data.status != 'abort') {
+        if (data.status != 'abort' && !data.aborted) {
           if (status >= 400 || status == 0) {
             outcome = OUTCOME_FAILURE
           } else {
             outcome = OUTCOME_SUCCESS
           }
+        } else {
+          outcome = OUTCOME_UNKNOWN
```
vigneshshanmugam, line 348: We should not need to set this, APM server treats the outcome as `Unknown` if its not set. 

Could you double check this one?

vigneshshanmugam (approved):
LGTM

packages/rum-core/src/performance-monitoring/performance-monitoring.js
devcorpio: You are right, transaction.outcome is set to `Unknown` if it's not set. Nevertheless,  with the span.outcome it's not occurring the same.

Screenshot of aborted span request **without** the change:

Screenshot 2022-04-28 at 12 52 58

Screenshot of aborted span request **with** the change:

Screenshot 2022-04-28 at 12 53 41

vigneshshanmugam: Wow TIL, not sure if it's the intended behavior but all good from our side 👍🏽

PR #1213, opened 2022-05-04
fix(rum): end fetch span when response stream ends

devcorpio:
# Summary

From now on, fetch spans will be ended once the response stream is completely consumed.
- Browser not supporting [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) will keep working as they were already doing.
- If the server triggers an error while emitting chunks of data, the logic in charge of reading the stream will notice and then the transaction/span outcome will be set as a **failure**
- If the request is aborted while reading the stream, the transaction/span outcome will be set as a **unknown** (following the strategy started in a [previous PR](https://github.com/elastic/apm-agent-rum-js/pull/1211))

The memory analysis have not reported anything grave. A local website (always in incognito, without extensions) have been tested in the following situations. (btw, the test consisted in generate a **fetch** request every time I clicked in the text)

**1.  before applying the changes**

<ins>**Allocation timeline analysis**</ins>

https://user-images.githubusercontent.com/15065076/166743933-8b168437-d264-495a-9ff1-244fe9339190.mov

<ins>**Heap snapshots**</ins>

https://user-images.githubusercontent.com/15065076/166745774-41784d0d-7325-4eb6-9e5a-ea5aed57ca22.mov

**2.  after applying the changes**

<ins>**Allocation timeline analysis**</ins>

https://user-images.githubusercontent.com/15065076/166744224-91c134e3-aeca-4bec-832c-d5d5b8daa778.mov

<ins>**Heap snapshots**</ins>

https://user-images.githubusercontent.com/15065076/166744283-95c30b80-af41-47a2-abe1-36da06a9b82c.mov

**My high level thoughts about the memory analysis:**

The changes in memory after adding the change are negligible. As you can see in the comparison of the snapshots and when seeing the statistics chart.

What I have seen is that we might have **smalls** leaks of memory in the agent. (before and after the change). It might be interesting to check thoroughly in a separated task/issue.

### DIFF
--- a/packages/rum-core/test/common/fetch-patch.spec.js
+++ b/packages/rum-core/test/common/fetch-patch.spec.js
@@ -49,22 +49,58 @@ describe('fetchPatch', function () {
   })
 
   if (window.fetch) {
-    it('should fetch correctly when response cannot be read from a stream', function (done) {
-      let originalCloneFn = window.Response.prototype.clone
-      window.Response.prototype.clone = function () {
-        return {}
-      }
+    describe('when response cannot be read from a stream', () => {
+      let originalCloneFn
 
-      var promise = window.fetch('/')
-      expect(promise).toBeDefined()
-      expect(typeof promise.then).toBe('function')
-      expect(events.map(e => e.event)).toEqual(['schedule'])
-      promise.then(function (resp) {
-        expect(resp).toBeDefined()
-        Promise.resolve().then(function () {
-          expect(events.map(e => e.event)).toEqual(['schedule', 'invoke'])
-          window.Response.prototype.clone = originalCloneFn
-          done()
+      beforeEach(() => {
+        originalCloneFn = window.Response.prototype.clone
+        window.Response.prototype.clone = function () {
+          return {}
+        }
+      })
+
+      afterEach(() => {
+        window.Response.prototype.clone = originalCloneFn
+      })
+
+      it('should fetch correctly', function (done) {
+        var promise = window.fetch('/')
+        expect(promise).toBeDefined()
+        expect(typeof promise.then).toBe('function')
+        expect(events.map(e => e.event)).toEqual(['schedule'])
+        promise.then(function (resp) {
+          expect(resp).toBeDefined()
+          Promise.resolve().then(function () {
+            expect(events.map(e => e.event)).toEqual(['schedule', 'invoke'])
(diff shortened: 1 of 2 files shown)


devcorpio:
Ummm, for some reason, not all the videos are loading properly...

packages/rum-core/test/common/fetch-patch.spec.js
```diff
@@ -49,20 +49,146 @@ describe('fetchPatch', function () {
...
+
+          return streamFinished
+        }
+
+        it('should produce task events when all data have been read', function (done) {
+          const streamFinished = mockStreamReader()
+
+          window.fetch('/')
+          streamFinished.then(() => {
+            Promise.resolve().then(function () {
```
vigneshshanmugam, line 152: Do we need this microtask `Promise.resolve` here? I guess the stream already resolves after the microtask, which is what made me curious.

vigneshshanmugam (approved):
LGTM 🎉 Thanks for taking care of this one. 

> What I have seen is that we might have smalls leaks of memory in the agent. (before and after the change). It might be interesting to check thoroughly in a separated task/issue.

Not sure if it could be a leak, We already have benchmarks with memory allocation profile - https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum/test/benchmarks/BENCHMARKS.md. I can go over those benchmarks in our next sync so we can test those with these changes.

packages/rum-core/test/common/fetch-patch.spec.js
devcorpio: We need that **microtask** there, because in the `getReader` fake we are resolving the dummy promise before the getReader one:

```
    if (readCounter > 0) {
      const response = Promise.resolve({
        value: undefined,
        done: true
      })

      resolveStream()

      return response
    }
```

The microtask makes sure the expect will find the proper data.

Another way of doing this is to resolve the dummy at the end:

```
    if (readCounter > 0) {
      const response = Promise.resolve({
        value: undefined,
        done: true
      })

      // a macrotask here
      setTimeout(resolveStream, 0) 

      return response
    }
```

With this we wouldn't need the `Promise.resolve` in the test. I'm not have a strong opinion about this. Both ways are okay to me (since other tests were already doing something similar as well). WDYT?

PR #1214, opened 2022-05-05
feat(rum): disable beacon compression when flag is set

devcorpio:
# Context

In the past it has been common to find the situation where we needed a HAR file to investigate issues of a support case.

Most of the time the HAR files provided were not showing the payload of the events sent by the RUM agent. That was happening because they were generated in **Chrome**. That browser supports **Compression Streams API** which it is being used by the agent to compress events payload. Although it would be possible to generate them with a browser such as Firefox, we have found that in particular cases that wasn't a viable option for support.

Because of that, in a conversation with @vigneshshanmugam we agreed to provide a way of disabling payload compression for such scenarios.

# What the PR is doing

Disables the payload compression:

1.  if an item with the key `_elastic_inspect_beacon_` exists in `sessionStorage`.
2.  If the page is loaded including the url param `_elastic_inspect_beacon_`. E.g. `https://elastic.co?_elastic_inspect_beacon_` (this creates the item in sessionStorage automatically)

The scenario **2** covers the case where support is not directly creating the HAR and instead explaining to a customer how to do it. The way I see it, to add `?_elastic_inspect_beacon_` to a URL is much simpler than creating a sessionStorage item manually.

# demo 

- creating the sessionStorage item manually

https://user-images.githubusercontent.com/15065076/166950345-c52b426f-3ad1-470e-bb56-e4ec91eff595.mov

- accessing a website with the query param included

https://user-images.githubusercontent.com/15065076/166953487-f82b7473-5679-417f-9800-6f1b3e86bbde.mov

Pending task:

- [x] Update the troubleshooting section so as to add this

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -35,16 +35,16 @@ In order to debug the agent consider performing the following steps:
 [[disable-events-payload-compression]]
 === Disable events payload compression
 
-In browsers such as Chrome the payload of the events sent by the RUM JS agent is being compressed with gzip.
-Because of that, you will not see a readable content when inspecting the event with the Network panel of your browser developer tools.
+In browsers such as Chrome, the RUM agent event payload is compressed with gzip.
+Because of this compression, you will not see readable content when inspecting the event with the Network panel of your browser developer tools.
 
-There are situations where that can be an issue. For instance, HAR files will not show a readable information. Therefore, inspecting events
+There are situations where that can be an issue. For instance, HAR files will not show readable information. Therefore, inspecting events
 for debugging purposes will not be possible.
 
 There are two ways to disable the payload compression:
 
-1. Creating an item named `_elastic_inspect_beacon_` using the sessionStorage browser API.
-2. Loading the webpage including the query param `_elastic_inspect_beacon_` in the URL. E.g. `https://elastic.co?_elastic_inspect_beacon_`
+1. Create an item named `_elastic_inspect_beacon_` using the sessionStorage browser API.
+2. Load the webpage with the query param `_elastic_inspect_beacon_` in the URL. For example, `https://elastic.co?_elastic_inspect_beacon_=true`
 
 The effect of this will remain until the tab or browser is closed.
 
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -32,6 +32,23 @@ In order to debug the agent consider performing the following steps:
 4. Monitor the Console and Network panel in your browsers developer tools
 
 [float]
+[[disable-events-payload-compression]]
+=== Disable events payload compression
+
+In browsers such as Chrome the payload of the events sent by the RUM JS agent is being compressed with gzip.
+Because of that, you will not see a readable content when inspecting the event with the Network panel of your browser developer tools.
+
+There are situations where that can be an issue. For instance, HAR files will not show a readable information. Therefore, inspecting events
+for debugging purposes will not be possible.
+
+There are two ways to disable the payload compression:
+
+1. Creating an item named `_elastic_inspect_beacon_` using the sessionStorage browser API.
+2. Loading the webpage including the query param `_elastic_inspect_beacon_` in the URL. E.g. `https://elastic.co?_elastic_inspect_beacon_`
+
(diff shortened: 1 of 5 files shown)


packages/rum-core/src/common/utils.js
```diff
@@ -401,6 +401,36 @@ function isPerfTypeSupported(type) {
...
 }
 
+/**
+ * The goal of this is to make sure that HAR files
+ * can be created containing beacons with the payload readable
+ */
+function isBeaconInspectionEnabled() {
+  const flagName = '_elastic_inspect_beacon_'
+
+  if (sessionStorage.getItem(flagName) !== null) {
```
vigneshshanmugam, line 411: Why not look for values other than null or undefined?, I would change this to != null to cover for those two cases.

packages/rum-core/src/common/utils.js
```diff
@@ -401,6 +401,36 @@ function isPerfTypeSupported(type) {
...
+
+  if (!window.URL || !window.URLSearchParams) {
+    return false
+  }
+
+  try {
+    const parsedUrl = new URL(window.location.href)
+    const isFlagSet = parsedUrl.searchParams.has(flagName)
+    if (isFlagSet) {
+      sessionStorage.setItem(flagName, '')
```
vigneshshanmugam, line 423: maybe set to something to boolean?

vigneshshanmugam (commented):
LGTM with pending comments.

packages/rum-core/src/common/utils.js
devcorpio: Thanks for asking this. That is done in that why because storage (session and local) returns [null](https://developer.mozilla.org/en-US/docs/Web/API/Storage/getItem#return_value) when the key doesn't exist:

> A string containing the value of the key. If the key does not exist, null is returned.

Apart from that, the value of an item is always stored as a string, for instance:

```
sessionStorage.setItem('key', null) ---> stored as "null"
sessionStorage.setItem('key', undefined) ---> stored as "undefined"
// etc
```

packages/rum-core/src/common/utils.js
devcorpio: Yes I'm with you, even if the boolean is stored as a string, I believe it looks better than an empty string from the code standpoint.

docs/troubleshooting.asciidoc
```diff
@@ -31,6 +31,23 @@ In order to debug the agent consider performing the following steps:
...
+In browsers such as Chrome the payload of the events sent by the RUM JS agent is being compressed with gzip.
+Because of that, you will not see a readable content when inspecting the event with the Network panel of your browser developer tools.
+
+There are situations where that can be an issue. For instance, HAR files will not show a readable information. Therefore, inspecting events
+for debugging purposes will not be possible.
+
+There are two ways to disable the payload compression:
+
+1. Creating an item named `_elastic_inspect_beacon_` using the sessionStorage browser API.
+2. Loading the webpage including the query param `_elastic_inspect_beacon_` in the URL. E.g. `https://elastic.co?_elastic_inspect_beacon_`
```
vigneshshanmugam, line 47: https://elastic.co?_elastic_inspect_beacon_=true

docs/troubleshooting.asciidoc
```diff
@@ -31,6 +31,23 @@ In order to debug the agent consider performing the following steps:
 3. Wait at least 5 seconds after the page has loaded
 4. Monitor the Console and Network panel in your browsers developer tools
 
+[float]
+[[disable-events-payload-compression]]
+=== Disable events payload compression
+
+In browsers such as Chrome the payload of the events sent by the RUM JS agent is being compressed with gzip.
+Because of that, you will not see a readable content when inspecting the event with the Network panel of your browser developer tools.
```
bmorelli25, line 39:
```suggestion
In browsers such as Chrome, the RUM agent event payload is compressed with gzip.
Because of this compression, you will not see readable content when inspecting the event with the Network panel of your browser developer tools.
```

devcorpio:
Thanks for all the suggestions @bmorelli25. Changes applied! :elasticheart:

bmorelli25 (approved):
Docs LGTM

vigneshshanmugam (approved):

Issue #1174, opened 2022-02-28
patchEventTarget doubles events (under special conditions)


christianhaller3000:
When we deploy `apm-agent-rum-js` to our (pretty large) e-commerce website, we are losing some click events.

overwriting the native prototype `window.EventTarget.prototype` or wrapping the native events seems to have side effects
https://github.com/elastic/apm-agent-rum-js/blob/f58bc1b836f1e3caf07b668cbed6df619bdfdc92/packages/rum-core/src/common/patching/event-target-patch.js

Could you offer an config property to bypass the  event listener wrapper?

devcorpio:
hI @christianhaller3000 ,

Thanks for bringing this to our notice!

It is possible to disable the event target wrapping making use of [disableInstrumentations](https://www.elastic.co/guide/en/apm/agent/rum-js/4.x/configuration.html#disable-instrumentations) config. In this case the instrumentation to disable would be ```eventtarget```

Let me know if that helps you!

--

Apart from that, I would like to understand better the situation and I would like to ask a couple questions:

- What do you mean with exactly```we are losing some click events```? Is there any interaction failing?
- Could we reproduce somehow that scenario from our side? Is the website accessible?
- If so, could you tell us how to reproduce the issue with the clicks?


Thanks,
Alberto

christianhaller3000:
Hi @devcorpio,
disabling the instrumentation does not help.

But I can provide a simple example:

```html
<!DOCTYPE html>
<html>

<head>
    <meta charset="UTF-8">
    <title></title>
</head>
<body>
    <!-- this is before we add the click event listeners, move this this between -->
    <script src="https://cdn.jsdelivr.net/npm/@elastic/apm-rum@5.10.2/dist/bundles/elastic-apm-rum.umd.js"></script>
    <button>click me and and watch the console.log output</button>
    <script>
        const fn = function () {
            console.log("should be only one time logged");
        };
        document.querySelector('button').addEventListener('click', fn);
    </script>
    <!-- place elastic-apm-rum here and fn will be called twice 😳 -->
    <script>
        document.querySelectorAll('button')[0].addEventListener('click', fn);
    </script>
</body>
</html>
```

- If click the button, you will see console.log **1** time.
- If you move the `elastic-apm-rum.umd.js` *between* the addEventListener statements and click the button, you will see console.log **2** times. ❌
- If you drop `elastic-apm-rum.umd.js` and click the button, you will see console.log **1** time. ✅

[...]

Issue #772, opened 2020-04-30
Investigate breaking changes needed for api version 3


jahtalab:
Since we have started working on api version 3, planned to be used in agent version 6.0, we should identify if there are any other changes we want to make before we release our next major version.

So far we have:
- Compression of payload on v3 api

cc @vigneshshanmugam

seravifer:
If you are going to remove support for Internet Explorer 11, remove `promise-polyfill`.

PR #1219, opened 2022-05-11
fix(rum): listen to all page clicks

devcorpio:
Relates to https://github.com/elastic/apm-agent-rum-js/issues/1174

# Summary

From now on, the RUM agent will start listening to all page clicks. The instrumentation of the [add/remove]EventListener has been removed mainly because of the bug that takes places in [special conditions](https://github.com/elastic/apm-agent-rum-js/issues/1174#issuecomment-1055492388)

The details of the approach followed in this PR are available [here](https://github.com/elastic/apm-agent-rum-js/issues/1174#issuecomment-1082077607). (Proposal 4)


packages/rum-core/src/common/observers/page-clicks.js
```diff
@@ -0,0 +1,94 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ */
+
+import { USER_INTERACTION } from '../constants'
+
+export function observePageClicks(transactionService) {
+  const clickHandler = event => {
+    if (event.target instanceof Element) {
```
devcorpio, line 30: The tests related to that condition are available here: https://github.com/elastic/apm-agent-rum-js/blob/9658d373223862964c12453a434d6cdc3443edb0/packages/rum-core/test/common/observers/page-clicks.spec.js#L81

packages/rum-core/test/common/observers/page-clicks.spec.js
```diff
@@ -0,0 +1,142 @@
...
+  let serviceFactory
+  let transactionService
+  let performanceMonitoring
+  let unobservePageClicks
+  let containerElement
+
+  beforeEach(() => {
+    serviceFactory = createServiceFactory()
+    transactionService = serviceFactory.getService(TRANSACTION_SERVICE)
+    performanceMonitoring = serviceFactory.getService(PERFORMANCE_MONITORING)
```
devcorpio, line 45: Before creating this PR, PERFORMANCE_MONITORING was "PerformanceMonitoring". I considered that it was a good time to create the constant for that and for 'ErrorLogging'.

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -149,9 +151,31 @@ describe('ApmBase', function () {
...
+
+    const tr = apmBase.getCurrentTransaction()
+    expect(tr.name).toBe('Click - body')
+  })
+
+  it('should not observe click event if eventtarget instrumentation is disabled', () => {
+    apmBase.init({
+      serviceName,
+      serverUrl,
+      disableInstrumentations: [EVENT_TARGET]
```
vigneshshanmugam, line 167: we might have to change the naming to click. Feels weird to have it `eventtarget`. 

The disableInstrumentations can be monkeypatched to support both names and we can depreacte `eventtarget` later when doing `6.x`.

packages/rum-core/src/common/observers/page-clicks.js
vigneshshanmugam, line 30: Any reason to pick Element and not `EventTarget` ? Can you add a comment?

vigneshshanmugam (approved):
Love the new changes, Look at those size differences 🎉 Great work @devcorpio 

Added few comments, No blockers though.

packages/rum-core/src/common/observers/page-clicks.js
devcorpio: If we check against `EventTarget` we will need to handle also events that are not corresponding to a DOM element.

Two examples:

1. `window.dispatchEvent(new Event('click'))` will cause the event.target to be an instance of [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window). 
2. `document.dispatchEvent(new Event('click'))` will cause the event.target to be an instance of [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document).

It's important to mention that while I was working on this feature I had a "prototype" where I was handling them. In case of Window I was creating a transaction with the name "click - window" and in case of Document the name was "click - document"

But our goal is to capture **real** user click interactions (the target will be always `Element`). I decided not to handle them.

I'm going to add to the comment explaining the why ✍️ 😄

packages/rum/test/specs/apm-base.spec.js
devcorpio: Good point!. I was thinking that maybe what we can do is:

- Create the new name.
- Keep supporting the old one (like you are saying), but in the release notes and in the [documentation](https://www.elastic.co/guide/en/apm/agent/rum-js/current/configuration.html#disable-instrumentations) that it's considered **deprecated** and that in a future version will be removed

Will update the [6.x issue](https://github.com/elastic/apm-agent-rum-js/issues/772) too

docs/configuration.asciidoc
```diff
@@ -96,14 +96,16 @@ The valid options are:
...
 * `xmlhttprequest`
 * `fetch`
 * `error`
 
 NOTE: To disable all `http-request` transactions, add both `fetch` and `xmlhttprequest`.
 to this config.
 
-NOTE: To disable `user-interaction` transactions,  add `eventtarget` to this config.
+NOTE: To disable `user-interaction` transactions,  add `eventtarget` or `click` to this config.
+The option `eventtarget` is deprecated and it might be removed in a future release.
```
devcorpio, line 108: I have updated the documentation informing that `eventtarget` is deprecated and that might be removed in the future.

The screenshot below shows how it will look like:
Screenshot 2022-05-12 at 15 18 54


What do you think?

packages/rum-core/src/common/observers/page-clicks.js
vigneshshanmugam: Gotcha, that makes sense. Thanks for the details :)

docs/configuration.asciidoc
vigneshshanmugam, line 108:
```suggestion
The option `eventtarget` is deprecated and will be removed in the future releases.
```

vigneshshanmugam: Added a small suggestion, looks good 👍🏽

PR #1230, opened 2022-05-23
feat: warn users when invalid config options provided when init

### LINKED ISSUE #1228: Warn users when initialising agent with invalid config options
# Context

On different interactions with support and users of the community it has been detected that sometimes the agent is initialised with invalid config options:

- Trying to set options that are documented in other agents. **secretToken** is a case in point. 
- Because of small typos. For example, **enviroment** rather than **environment**

Since the agent doesn't warn about it, the users believe that everything related to those options is going to work as expected.

# What we want to achieve

We want the agent to warn users if they initialise the agent with non-existent config options. When it comes to the validation, it's important to bear in mind that at this moment there are two required options: `serverUrl` and `serviceName` and if they are invalid the agent **is not activated.**

The strategy to follow in this issue should be a bit different, the idea here is **to warn but keep permitting the agent to be activated**. We should realise that at this point likely there are websites having an agent with invalid config options set (for the reasons commented above) gathering info properly, if our strategy included to avoid the initialisation then those agents would stop working.

kyungeunni:
closes #1228 

Check the config options and print warnings when unknown config options are specified. 
As mentioned in the ticket, we won't affect any initiation logic but keep permitting the agent to be activated.

### DIFF
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -211,7 +211,7 @@ class Config {
         errors.missing.push(key)
       }
 
-      if (allKeys.indexOf(key) < 0) {
+      if (allKeys.indexOf(key) === -1) {
         errors.unknown.push(key)
       }
     })
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -95,7 +95,6 @@ class Config {
       session: false,
       apmRequest: null
     }
-
     this.events = new EventHandler()
     this.filters = []
     /**
@@ -198,16 +197,22 @@ class Config {
    */
   validate(properties = {}) {
     const requiredKeys = ['serviceName', 'serverUrl']
+    const allKeys = Object.keys(this.config)
     const errors = {
       missing: [],
-      invalid: []
+      invalid: [],
+      unknown: []
     }
     /**
-     * Check when required keys are missing
+     * Check when required keys are missing or unknown keys found
      */
     Object.keys(properties).forEach(key => {
       if (requiredKeys.indexOf(key) !== -1 && !properties[key]) {
         errors.missing.push(key)
+      }
+
+      if (allKeys.indexOf(key) < 0) {
+        errors.unknown.push(key)
       }
     })
     /**
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -95,6 +95,7 @@ class Config {
       session: false,
       apmRequest: null
     }
+
     this.events = new EventHandler()
     this.filters = []
     /**
--- a/packages/rum-core/test/common/config-service.spec.js
+++ b/packages/rum-core/test/common/config-service.spec.js
@@ -138,14 +138,16 @@ describe('ConfigService', function () {
     const errors1 = configService.validate({ serviceName: 'name' })
     expect(errors1).toEqual({
       missing: [],
-      invalid: []
+      invalid: [],
+      unknown: []
     })
 
     // missing required key serviceName
     const errors2 = configService.validate({ serviceName: undefined })
     expect(errors2).toEqual({
       missing: ['serviceName'],
-      invalid: []
+      invalid: [],
+      unknown: []
     })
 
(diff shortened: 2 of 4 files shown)


packages/rum-core/src/common/config-service.js
```diff
@@ -198,17 +197,23 @@ class Config {
...
     /**
-     * Check when required keys are missing
+     * Check when required keys are missing or unknown keys found
      */
     Object.keys(properties).forEach(key => {
       if (requiredKeys.indexOf(key) !== -1 && !properties[key]) {
         errors.missing.push(key)
       }
+
+      if (allKeys.indexOf(key) < 0) {
```
devcorpio, line 214: nit:

What do you think about checking against `!== -1` rather than `< 0`? 

(Proposing this to use the same style as others indexOf in the codebase)

packages/rum-core/src/common/config-service.js
```diff
@@ -198,17 +197,23 @@ class Config {
    */
   validate(properties = {}) {
     const requiredKeys = ['serviceName', 'serverUrl']
+    const allKeys = Object.keys(this.config)
```
devcorpio, line 200: nit:

What about configKeys?

packages/rum-core/test/common/config-service.spec.js
```diff
@@ -209,6 +213,18 @@ describe('ConfigService', function () {
     ])
   })
 
+  it('should check unknown config options', () => {
+    const errors1 = configService.validate({
```
devcorpio, line 217: nit:

What about `errors` instead?

devcorpio:
LGTM @kyungeunni, congratulations for the first of plenty of contributions to come 🎉  

I just added small suggestions, nothing that I deem blocking!

devcorpio (approved):

packages/rum-core/src/common/config-service.js
kyungeunni: yes, I'm all about consistency so I like your suggestion!! will update 👍

packages/rum-core/test/common/config-service.spec.js
kyungeunni: works for me 👍 (you can see I copied and pasted :D)

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -416,6 +416,16 @@ describe('ApmBase', function () {
     expect(loggingService.error).toHaveBeenCalledWith(
       `RUM agent isn't correctly configured. serviceName "abc.def" contains invalid characters! (allowed: a-z, A-Z, 0-9, _, -, <space>)`
     )
+
+    logErrorSpy.calls.reset()
```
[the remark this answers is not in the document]
kyungeunni: That is a great suggestion, I will apply the changes in both places 👍

vigneshshanmugam (approved):
LGTM 🎉 Congrats on the first PR to the RUM agent :dance

kyungeunni:
@devcorpio I've addressed your feedback!! could you take another look, please? 🙏

packages/rum/test/specs/apm-base.spec.js
```diff
@@ -386,36 +386,56 @@ describe('ApmBase', function () {
     expect(loggingService.warn).toHaveBeenCalledWith('RUM agent is inactive')
   })
 
-  it('should log errors when config is invalid', () => {
-    const loggingService = serviceFactory.getService(LOGGING_SERVICE)
-    spyOn(loggingService, 'warn')
+  describe('invalid config', () => {
     const logErrorSpy = spyOn(loggingService, 'error')
```
devcorpio, line 390: There loggingService reference doesn't exist there yet, and the build fails. A possible way to approach this could be to isolate the loggingService related stuff in one place: (just an example!)

```
 let loggingService

    beforeEach(() => {
      loggingService = serviceFactory.getService(LOGGING_SERVICE)
      spyOn(loggingService, 'warn')
      spyOn(loggingService, 'error')
    })
```

devcorpio:
LGTM! 🥳  I really like the changes you made in the unit tests!

PR #1238, opened 2022-06-06
feat: allow requests to be sent with credentials

### LINKED ISSUE #1237: Allow RUM agent to send credentials on CORS requests
We have our APM server behind a reverse proxy which enforces that all requests must be sent with a session cookie. Requests missing the session cookie receive a 401 response from the reverse proxy. When running apps on a subdomain, APM requests are rejected with a 401 because [CORS requests are not sent with cookies by default](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials). RUM requests from the same domain work without issue.

It would be nice to have an option in that enables RUM requests to be sent with credentials.

stevenhair:
fix #1237 

This allows RUM clients to send cookies to APM servers that are behind reverse proxies that require requests to be authenticated.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -396,3 +396,13 @@ apm.init({
   }
 })
 ----
+
+
+[float]
+[[send-credentials]]
+==== `sendCredentials`
+
+* *Type:* Boolean
+* *Default:* `false`
+
+Instructs the agent to send credentials when making requests to the APM server.
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -405,4 +405,8 @@ apm.init({
 * *Type:* Boolean
 * *Default:* `false`
 
-Instructs the agent to send credentials when making requests to the APM server.
+This allows the agent to send cookies when making requests to the APM server.
+This is useful on scenarios where the APM server is behind a reverse proxy that requires requests to be authenticated.
+
+NOTE: If APM Server is deployed in an origin different than the page’s origin, you will need to
+<<configuring-cors, configure Cross-Origin Resource Sharing (CORS)>>.
--- a/packages/rum-core/src/common/apm-server.js
+++ b/packages/rum-core/src/common/apm-server.js
@@ -136,19 +136,28 @@ class ApmServer {
     url,
     { timeout = HTTP_REQUEST_TIMEOUT, payload, headers, beforeSend } = {}
   ) {
+    const sendCredentials = this._configService.get('sendCredentials')
+
     // This bring the possibility of sending requests that outlive the page.
     if (!beforeSend && shouldUseFetchWithKeepAlive(method, payload)) {
       return sendFetchRequest(method, url, {
         keepalive: true,
         timeout,
         payload,
-        headers
+        headers,
+        sendCredentials
       }).catch(reason => {
         // Chrome, before the version 81 had a bug where a preflight request with keepalive specified was not supported
         // xhr will be used as a fallback to cover fetch network errors, more info: https://fetch.spec.whatwg.org/#concept-network-error
         // Bug info: https://bugs.chromium.org/p/chromium/issues/detail?id=835821
         if (reason instanceof TypeError) {
-          return sendXHR(method, url, { timeout, payload, headers, beforeSend })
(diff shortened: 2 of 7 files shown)


stevenhair:
`/test`

devcorpio:
Hi @stevenhair,

Thank you for your contribution to the RUM agent! 🎉  We will review it soon!

First, I would to discuss about the [issue](https://github.com/elastic/apm-agent-rum-js/issues/1237). I have a few comment/questions related to it:

It would be helpful to know the high level architecture you have before proceeding with this. With architecture I mean:

- the domain of your apm server
- the domain which creates the session cookie 
- the subdomain where the requests fail. 

**Note**: Please, use fake domains if necessary, what I want to know is their relationship. For instance, to know if the domain associated to your apm-server is the same one as the one creating the session cookie, etc.

> because CORS requests are not sent with cookies by default

Although this is true, adding the "sendCredentials" is not always enough, and since we are adding a new configuration to the agent there are very important nuances about [sites and origins](https://web.dev/same-site-same-origin/) and server configuration that we need to keep in mind when updating the documentation.

- Only `same-site` cookies will be included with the request even with the flag enabled. If for whatever reason someone wants to make sure that a cross-site cookie (third-party one) is added to the request, the cookie should be created with `SameSite` set to `None` and `Secure` to `true`.
 
- The server should set `Access-Control-Allow-Credentials` header to `true` and also to make sure that `Access-Control-Allow-Origin` header contains the specific origin rather than the wildcard. Otherwise it will fail.


By the way, have you been able to test somehow the changes that you have done in the agent against your website? (To confirm that those nuances are not affecting to your particular scenario)

stevenhair:
@devcorpio sorry for the delay - your response made me realize that I hadn't actually tested this, which led me to a small issue in our server-side configuration that I had to resolve before I could test.

Here's our setup:
- APM server URL: https://subdomain1.company.com/apm/...
- Cookie domain: `.company.com`
- Instrumented app URL: https://appname.subdomain2.company.com

The cookie domain is valid for both the app and APM URLs, but the browser doesn't send it by default because it's still a cross-origin request.

I can confirm that these changes work when requests are made using `XMLHttpRequest` (in Firefox), but they don't seem to work when using `fetch()` (in Chrome). I'll dig into that more.

stevenhair:
Ok, the issue with `fetch()` was indeed still server-side config - a combination of some strange Nginx behaviors and a few headers missing from the `Access-Control-Allow-Headers` header (unfortunately, [you can't set it to a wildcard when allowing credentials](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#directives)). I can confirm that this implementation works for my use case.

docs/configuration.asciidoc
```diff
@@ -396,3 +396,13 @@ apm.init({
...
+
+
+[float]
+[[send-credentials]]
+==== `sendCredentials`
+
+* *Type:* Boolean
+* *Default:* `false`
+
+Instructs the agent to send credentials when making requests to the APM server.
```
devcorpio, line 408:
```suggestion


[float]
[[send-credentials]]
==== `sendCredentials`

* *Type:* Boolean
* *Default:* `false`

This allows the agent to send cookies when making requests to the APM server.
This is useful on scenarios where the APM server is behind reverse proxies that require requests to be authenticated.

NOTE: This feature requires changes on your server configuration. The server has to set `Access-Control-Allow-Credentials` header to `true` and `Access-Control-Allow-Origin` header to the origin doing the request rather than a wildcard.

```

devcorpio: I recommend to update the documentation as suggested above (or something similar, in case you want to amend something)

On the screenshot below we can see how it would look like once the docs are updated:

Screenshot 2022-06-13 at 13 58 01

devcorpio:
@stevenhair LGTM!, just added a comment related to the docs!

Thanks again for your contribution 🥳

devcorpio:
cla/check

docs/configuration.asciidoc
stevenhair: I took your comment and ran with it - I added a link to the CORS configuration section.

image

devcorpio (approved):

PR #1239, opened 2022-06-07
fix(rum-core): capture stack trace from syntax errors properly

devcorpio:
# Context

The RUM agent was capturing errors thrown in [inline scripts](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#examples) with an **empty stack trace**.  There is a nuance to highlight, though.

This **only** affects to errors whose type is [SyntaxError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) and have been triggered because of a malformed JavaScript code. This tend to occur when the browser is parsing the document.

The following two examples will help to understand that nuance:

Example 1 (**KO**): SyntaxError because of malformed JavaScript while parsing the document:

```
<html>
  <head>Example 1: SyntaxError because of malformed JavaScript while parsing the document </head>
  <body>
     Hello, I am Example 1 page.
    <script>
         var a:      // This throws a Syntax Error
    </script>
  </body>
</html>
```

As we can see on the screenshot below the **stacktrace** was empty:

Screenshot 2022-06-07 at 16 20 22

Example 2 (**OK**): SyntaxError because of bad usage of JSON.parse

```
<html>
  <head>Example 2: SyntaxError because of bad usage of JSON.parse </head>
  <body>
     Hello, I am Example 2 page.
    <script>
         JSON.parse("bad json") // this will throw a SyntaxError too
    </script>
  </body>
</html>
```

As we can see on the screenshot below the stack trace was **not** empty, since this kind of SyntaxError was being handled properly:

Screenshot 2022-06-07 at 16 26 53

## How are we generating stacktraces?

One of the steps of the error capturing flow is the one that generates the stacktrace from an [error event](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent). Given the complexity that some cases have (cross browsing support for instance) we rely on a third-party library named [error-stack-parser](https://github.com/stacktracejs/error-stack-parser). Unfortunately, that case is not being handled as expected.

## Fix

The ErrorEvent that the browser dispatches when an uncaught error takes place contains all the information we need to generate the **stacktrace**.  So, that's what we do now, we create a stack trace of a single element which contains the missing information.

After the fix, the **stacktrace** of **example 1** looks this way: 🎉 

Screenshot 2022-06-07 at 16 48 43

## Next

We are going to create an issue in the third-party library to report the unexpected behaviour.

### DIFF
--- a/packages/rum-core/src/error-logging/stack-trace.js
+++ b/packages/rum-core/src/error-logging/stack-trace.js
@@ -23,8 +23,6 @@
  *
  */
 
-import stackParser from 'error-stack-parser'
-
 function filePathToFileName(fileUrl) {
   const origin =
     window.location.origin ||
@@ -91,7 +89,21 @@ function normalizeFunctionName(fnName) {
   return fnName
 }
 
-export function createStackTraces(errorEvent) {
+function isValidStackTrace(stackTraces) {
+  if (stackTraces.length === 0) {
+    return false
+  }
+
+  if (stackTraces.length === 1) {
+    const stackTrace = stackTraces[0]
+
+    return 'lineNumber' in stackTrace
+  }
+
+  return true
+}
+
+export function createStackTraces(stackParser, errorEvent) {
   const { error, filename, lineno, colno } = errorEvent
 
   let stackTraces = []
@@ -106,7 +118,11 @@ export function createStackTraces(errorEvent) {
     }
   }
 
-  if (stackTraces.length === 0) {
+  // error-stack-parser library doesn't generate proper stack traces from Syntax Errors
+  // thrown during the browser document parsing process. The outcome of the library depends on the browser:
+  // 1. on non-chromium browsers it throws an exception.
+  // 2. on chromium browsers it returns a malformed single stack trace element not containing the lineNumber property
+  if (!isValidStackTrace(stackTraces)) {
     stackTraces = [
       {
         fileName: filename,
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/error-logging/stack-trace.js
```diff
@@ -91,7 +89,21 @@ function normalizeFunctionName(fnName) {
...
+  if (stackTraces.length === 1) {
+    const stackTrace = stackTraces[0]
+
+    return 'lineNumber' in stackTrace
+  }
+
+  return true
+}
+
+export function createStackTraces(stackParser, errorEvent) {
```
kyungeunni, line 106: Is it correct that we are now passing stackParser here in order to mock this library when testing? :)

kyungeunni (approved):
LGTM!

packages/rum-core/src/error-logging/stack-trace.js
devcorpio: yes, that's right! 🎉 . In this particular scenario the option of injecting the dependency was the simplest strategy for making the testing experience smooth.

vigneshshanmugam (approved):
LGTM 🎉 Good catch Alberto.

packages/rum-core/src/error-logging/stack-trace.js
vigneshshanmugam: Any reason why you didnt go with mocking `error-stack-parser` using spy instead?

devcorpio: Unfortunately, with jasmine we cannot spy properly a third-party (node_modules one) dependency 😢

PR #1246, opened 2022-06-16
fix: check supportedEntryTypes when starting observer

### LINKED ISSUE #1245: agent logs warnings on not supported performance entry types
+ User has reported that the RUM agent logs warnings in FF pretty often - https://elasticstack.slack.com/archives/CNKF2LEMP/p1655314395008469?thread_ts=1655307617.469979&cid=CNKF2LEMP
```
Ignoring unsupported entryTypes: longtask
```

This happens because we try to observe for the performance entry types that is not supported by the FF browser - https://github.com/elastic/apm-agent-rum-js/blob/369800e17db59e39dbbfc3c9ab0e1483b45b2c9e/packages/rum-core/src/performance-monitoring/metrics.js#L334

+ The easy fix would be check if the entry type is supported in the browser before observing for the specific entry - https://w3c.github.io/performance-timeline/#dom-performanceobserver-supportedentrytypes

kyungeunni:
fixes #1245 

This PR checks `PerformanceObserver.supportedEntryTypes` before starting an observer for a certain entryType so that it doesn't print a warning on Firefox browser and doesn't trigger errors on Safari when it starts observer for unsupported types.

Before the change on Firefox (latest version):
[Image from Gyazo](https://gyazo.com/86731cfbb03863cbb454a57400d1d8db)

After the change on Firefox (local build of this PR):
<a href="https://gyazo.com/8324a6d181070ed09d4933651597428a">Image from Gyazo</a>

### Note
List of supported entry types:
- Chrome (102.0): 
`['element', 'event', 'first-input', 'largest-contentful-paint', 'layout-shift', 'longtask', 'mark', 'measure', 'navigation', 'paint', 'resource']`
- Firefox(101.0.1):  `[ "event", "first-input", "mark", "measure", "navigation", "paint", "resource"]`
- Safari(15.5): 
`["mark", "measure", "navigation", "paint", "resource"]`

### DIFF
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -30,7 +30,7 @@ import {
   FIRST_INPUT,
   LAYOUT_SHIFT
 } from '../common/constants'
-import { noop, PERF } from '../common/utils'
+import { noop, PERF, isPerfTypeSupported } from '../common/utils'
 import Span from './span'
 
 export const metrics = {
@@ -318,9 +318,7 @@ export class PerfEntryRecorder {
 
   start(type) {
     try {
-      const supportedEntryTypes =
-        window.PerformanceObserver?.supportedEntryTypes || []
-      if (supportedEntryTypes.indexOf(type) === -1) {
+      if (!isPerfTypeSupported(type)) {
         return
       }
       /**
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -317,12 +317,9 @@ export class PerfEntryRecorder {
   }
 
   start(type) {
-    /**
-     * Safari throws an error when PerformanceObserver is
-     * observed for unknown entry types as longtasks and lcp is
-     * not supported at the moment
-     */
-    try {
+    const supportedEntryTypes =
+      window.PerformanceObserver?.supportedEntryTypes || []
+    if (supportedEntryTypes.indexOf(type) !== -1) {
       /**
        * Start observing for different entry types depending on the transaction type
        * - `buffered`: true means we would be able to retrive all the events that happened
@@ -332,7 +329,7 @@ export class PerfEntryRecorder {
        *   buffered flag (https://w3c.github.io/performance-timeline/#observe-method)
        */
       this.po.observe({ type, buffered: true })
-    } catch (_) {}
+    }
   }
 
   stop() {
--- a/packages/rum-core/src/performance-monitoring/metrics.js
+++ b/packages/rum-core/src/performance-monitoring/metrics.js
@@ -317,9 +317,12 @@ export class PerfEntryRecorder {
   }
 
   start(type) {
-    const supportedEntryTypes =
-      window.PerformanceObserver?.supportedEntryTypes || []
-    if (supportedEntryTypes.indexOf(type) !== -1) {
+    try {
+      const supportedEntryTypes =
+        window.PerformanceObserver?.supportedEntryTypes || []
+      if (supportedEntryTypes.indexOf(type) === -1) {
+        return
+      }
       /**
        * Start observing for different entry types depending on the transaction type
        * - `buffered`: true means we would be able to retrive all the events that happened
@@ -329,6 +332,12 @@ export class PerfEntryRecorder {
        *   buffered flag (https://w3c.github.io/performance-timeline/#observe-method)
        */
       this.po.observe({ type, buffered: true })
+    } catch (_) {
+      /**
+       * Even though we check supportedEntryTypes before starting the observer,
+       * there are risks of exceptions according to the [spec](https://www.w3.org/TR/performance-timeline/#observe-method)
+       * But we ignore error since this is not intended to be seen by users
+       */
     }
   }
 
--- a/packages/rum-core/test/performance-monitoring/metrics.spec.js
+++ b/packages/rum-core/test/performance-monitoring/metrics.spec.js
@@ -33,6 +33,7 @@ import {
   createLongTaskSpans
 } from '../../src/performance-monitoring/metrics'
 import { LARGEST_CONTENTFUL_PAINT, LONG_TASK } from '../../src/common/constants'
+import { isPerfTypeSupported } from '../../src/common/utils'
 import {
   mockObserverEntryTypes,
   mockObserverEntryNames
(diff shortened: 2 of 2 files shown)


packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -317,12 +317,9 @@ export class PerfEntryRecorder {
   }
 
   start(type) {
-    /**
-     * Safari throws an error when PerformanceObserver is
-     * observed for unknown entry types as longtasks and lcp is
-     * not supported at the moment
-     */
-    try {
```
kyungeunni, line 325: I've removed try-catch based on the description in the comment, if folks think it's too risky, I will put it back.

devcorpio: I would remove the comment but I wouldn't remove the try-catch.

- **Main reason**: It's a browser API, I would treat it as an external dependency,  then I would follow a defensive approach.
- The [spec]((https://www.w3.org/TR/performance-timeline/#observe-method)) mention "error" and "exception" multiple times. **Theoretically** the only that can affect us is the one about the invalid entry type, theoretically... 😄 
- There are a ton of browsers (and versions) out there, and it's not completely out of the box to see bugs like the one commented on this issue: https://github.com/elastic/sdh-kibana/issues/2547#issuecomment-1029054684

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -332,7 +329,7 @@ export class PerfEntryRecorder {
        *   buffered flag (https://w3c.github.io/performance-timeline/#observe-method)
        */
       this.po.observe({ type, buffered: true })
-    } catch (_) {}
+    }
```
vigneshshanmugam, line 332: We added try/catch to avoid errors in unsupported browsers. Unsupported browsers like safari 10 would throw error. It's safer to have this.

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -317,12 +317,9 @@ export class PerfEntryRecorder {
...
 
   start(type) {
-    /**
-     * Safari throws an error when PerformanceObserver is
-     * observed for unknown entry types as longtasks and lcp is
-     * not supported at the moment
-     */
-    try {
+    const supportedEntryTypes =
+      window.PerformanceObserver?.supportedEntryTypes || []
```
devcorpio: We have the [isPerfTypeSupported](https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/common/utils.js#L396) util

packages/rum-core/src/performance-monitoring/metrics.js
kyungeunni: Thanks Alberto, all make 💯 sense!! I've put back try-catch and added some comment inside catch 🙏

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -317,12 +317,12 @@ export class PerfEntryRecorder {
...
   start(type) {
-    /**
-     * Safari throws an error when PerformanceObserver is
-     * observed for unknown entry types as longtasks and lcp is
-     * not supported at the moment
-     */
     try {
+      const supportedEntryTypes =
+        window.PerformanceObserver?.supportedEntryTypes || []
+      if (supportedEntryTypes.indexOf(type) === -1) {
```
devcorpio: Just thinking out loud:

[supportedEntryTypes](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/supportedEntryTypes#browser_compatibility) became available later than [observe](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe)

For instance:

                                            Chrome    Safari   Firefox
observe                                   52          11            57
supportedEntryTypes            73           12.1        68

If we stop reporting based on supportedEntryTypes we might start missing data. Okay, **maybe** in the case of Chrome and Firefox it's not a big deal since version 73 and 68 are old. Besides, they are [evergreen](https://learn-the-web.algonquindesign.ca/topics/browser-testing/#evergreen-browsers) browsers.

The case of Safari might be different. We support older versions such the 12. What is more, Safari is not an evergreen browser. 😭  As you can see [here](https://css-tricks.com/evergreen-does-not-mean-immediately-available/#aa-the-browsers-themselves) and [here](https://thingsthemselves.com/reminder-safari-is-not-an-evergreen-browser/)

It seems that we **might** be adding a breaking change discreetly.

Questions:

- Does the firefox warning has more importance than the versioning thingy?
- Is firefox the only browser reporting the warnings?

If firefox is the only one complaining and you both believe that what I commented above might be a problem, a possible proposal might be:

**Step 1:**

We are in 5.x version, we don't want new releases to cause unexpected regressions. What about checking if the user agent is firefox for the following entry types: **largest-contentul-paint,** **layout-shift**, and **long-task**. I know that checking user-agents is tricky. But I wanted to share this at least.

**Step 2:**

Once we go to 6.x version,  there is where we can do breaking changes. For instance, "We support the following browsers, etc, etc". And then we can remove the logic added in **Step 1**


--

I considered that at least was important to bring out this concern

What are your thoughts, @kyungeunni @vigneshshanmugam?

kyungeunni: This is a great point, thanks for bringing this up.
image

As you mentioned, most of the browsers shipped `supportedEntryTypes` later than `observe` so I can see this change will introduce regression as fewer metrics will be observed, even though the entryType is supported for the browsers.

I appreciate your suggestion on Step1, although would it be fair to assume the possibility each version of FF could have a different set of supported entry types? (e.g. let's say version 57 supports 3 entry types, version 60 supports 5 so not only **largest-contentul-paint**, **layout-shift,** and **long-task** would log the warning messages but other unsupported ones too). Also, there are uninvestigated browsers too that could possibly log the warning messages as same as FF does. That being said, it seems like we can't be sure to cover all the cases with this approach, It'd look like we didn't really fix this bug as we didn't cover all the cases.

I'd love to chat about plans for supported browsers for `6.x` and plan this change accordingly. Meanwhile, could we communicate with users that this is how Firefox-specific behavior and if we try to mute the log, it would introduce regression as we will capture fewer metrics? If it's not a critical bug from the user's perspective, that is.

devcorpio: I agree with you here. Likely, the most sensible thing we can do is not to mute the log since it would cause a regression and then wait until 6.x to retackle this. I would like to know @vigneshshanmugam thoughts before closing the PR though.

The issue related to this PR probably will end up having the breaking-change label. And also we will need inform about our decision to the support engineer that brought out this issue.

vigneshshanmugam: I would like us to think this in a different way. The main types we use for Observer are `largest-contentful-paint`, `long-tasks`, `paint`, `input-delay` - https://github.com/elastic/apm-agent-rum-js/blob/96c7e692c79528e9074408be9d94212f0a7cb375/packages/rum-core/src/performance-monitoring/transaction-service.js#L166-L169 .


 I believe none of those metrics would be supported by the older browsers such as `Safari 11`. If we can confirm that, then we should be good to do just doing a check with `supportedEntryTypes` even in 5.x versions. 

Let me know your thoughts.

devcorpio: @vigneshshanmugam thanks for your comment.

It's true, PAINT started to be compatible on Safari 14.1. I wouldn't worried about Chrome either, even if PAINT was available on 60 and supportedEntryTypes on 73. I understand that rely on Chrome being an evergreen browser allows us to be a bit more flexible with the browser support here.

🎉

devcorpio (approved):
LGTM 🎉

PR #1248, opened 2022-06-17
chore: remove script to lock dependency version

devcorpio:
# Summary

**Agent maintenance:**

A few months ago, in this [PR](https://github.com/elastic/apm-agent-rum-js/pull/1131), we introduced [npm-force-resolutions](https://github.com/rogeriochaves/npm-force-resolutions) to the project.

We did to force some downstream dependencies to resolve to particular versions, such as lodash and semver-regex.

**Why removing this now?**

1. The developer experience when doing commits (including amends) is not good. The process is quite slow and feels intrusive. 
2. Apart from that, none of the dependencies locked end up on the bundle used by our users/customers, all the locked versions are devDependencies.
3. It requires from us to keep track of **two** kind of versions: the ones of the installed packages and the ones related to the resolution.

Those are the main reasons to remove it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -522,7 +522,7 @@
             "cli-width": "^3.0.0",
             "external-editor": "^3.0.3",
             "figures": "^3.0.0",
-            "lodash": "4.17.21",
+            "lodash": "^4.17.21",
             "mute-stream": "0.0.8",
             "run-async": "^2.4.0",
             "rxjs": "^6.6.6",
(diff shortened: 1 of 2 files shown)


kyungeunni (approved):
it was bugging me, too. Thanks Alberto.
LGTM 🎉

vigneshshanmugam (approved):
LGTM, Didn't you get any changes in the package-lock files?

devcorpio:
> LGTM, Didn't you get any changes in the package-lock files?

thanks!

PR #1268, opened 2022-08-26
hotfix: change event capture timing

baishiwen9:
The error messages of http requests (such as js files, css files, img) cannot be captured at present, and the event monitoring timing should be changed to the capture stage.

### DIFF
--- a/packages/rum-core/src/error-logging/error-logging.js
+++ b/packages/rum-core/src/error-logging/error-logging.js
@@ -165,7 +165,7 @@ class ErrorLogging {
   registerListeners() {
     window.addEventListener('error', errorEvent =>
       this.logErrorEvent(errorEvent)
-    )
+    , true)
     window.addEventListener('unhandledrejection', promiseRejectionEvent =>
       this.logPromiseEvent(promiseRejectionEvent)
     )


devcorpio:
Hi @baishiwen9,

Before proceeding with this PR I would like to clarify a few points:

`The error messages of http requests (such as js files, css files, img) cannot be captured at present`

What do you mean exactly? Could you share a few example?

` and the event monitoring timing should be changed to the capture stage.`

Could you elaborate more on this, please?

Thanks,
Alberto

baishiwen9:
> 
If a JavaScript script fails to load, no error message can be captured.

For example: < script src=“https://www.xxx.com/static/js/123.js "></script > failed to load. The HTTP status is 404, but no error message was captured.

devcorpio:
Hi @baishiwen9,

Unfortunately, the change you want to introduce is not modifying the behaviour as you would expect. The `window.addEventListener('error', ...` handler does not detect if a file does not load properly.

I have a few questions:

- Have you seen any change on your application after applying this change?
- If so, could you share with us the html snippet you have been using to test this?

Thanks,
Alberto

devcorpio:
Thanks again for the time you have invested on this!

I'm closing PR because the made changes are not modifying the current behaviour and there has not been activity for a while.

Cheers,
Alberto

PR #1273, opened 2022-09-01
The iframe sharing session

dragonTour:
# context
In some older systems, the IFrame embedding subsystem was used. However, sessions are stored in sessionStorage and cannot use the same session

 # summary
Use postMessage to make Iframe collect the same session.a


dragonTour:
image
How to deal with this?

dragonTour:
run elasticsearch-ci/docs

PR #1283, opened 2022-09-28
run npm audit as a part of build process

gaurav-ingalkar:


devcorpio:
Hi @gaurav-ingalkar,

Thanks for your contribution!

I would like to highlight that we don't use "**npm audit**" mainly because two reasons:

- we rely on [dependabot](https://github.com/dependabot) to get security notifications.
- the agent doesn't have practically prod dependencies, most of our dependencies are just for development purposes.

For now we would like to like keep such strategy when it comes to security. Thanks again for the time you have put on this!

Cheers,
Alberto

PR #1306, opened 2022-11-07
chore: replace bundlesize with bundlesize2

devcorpio:
# Summary

**bundlesize** depends on a version of **brotli** that depends on **iltorb 2.4.5** which is an archived repo and it's failing with Node.js 14

This change is [recommended](https://github.com/siddharthkp/bundlesize/pull/370#issuecomment-1038968845) by the author of bundlesize.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -90,7 +90,7 @@ pipeline {
               deleteDir()
               unstash 'source'
               script{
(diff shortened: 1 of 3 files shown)


kyungeunni (approved):
LGTM! :shipit:

PR #1311, opened 2022-11-08
chore: use chrome stable to execute benchmarks

devcorpio:
## Context

Chrome 109 was causing Karma to crash when executing the benchmarks. The dependency that was installing such version is **google-chrome-unstable**.

The only visible log in the CI was **"Cannot start ChromeHeadless",** no useful details related to the error were provided.

Screenshot 2022-11-08 at 18 46 47

## Applied action

To fix this situation we have replaced **google-chrome-unstable** with **google-chrome-stable**. Now, the version installed is **107**, which works perfectly.

## Thoughts

We are not exactly sure why Chrome **109** was failing, that version will be [released](https://chromiumdash.appspot.com/schedule) in 2023, so probably there are bugs that Chrome contributors need to fix. If the benchmarks start failing in **2023** for the same reason we will need to approach this issue in a different way.

### DIFF
--- a/.ci/docker/node-playwright/Dockerfile
+++ b/.ci/docker/node-playwright/Dockerfile
@@ -99,13 +99,15 @@ RUN apt-get -qq install -y  --no-install-recommends \
     libxml2 \
     libxslt1.1
 
-# 4. Install Chrome unstable to run karma benchmark tests inside puppeteer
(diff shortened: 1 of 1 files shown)


kyungeunni (approved):
LGTM 🎉

PR #1344, opened 2023-02-24
feat(rum-core): run tasks that block Angular stabilization in a separate Zone

### LINKED ISSUE #983: throttle() function causes Angular to not get stable
When using your library with Angular, I noticed that the setTimeout() in your throttle() function is causing Angular to not get 'stable' for the given interval (default is 60s).

Angular considers itself 'unstable' when ZoneJS is seeing pending micro or macro tasks in the Angular zone. ('isStable' doc: https://angular.io/api/core/ApplicationRef#is-stable-examples)

The problem, as i understand it, here is that the throttle() gets executed within the Angular zone if Angular is present.

We use this 'isStable' flag for certain features that need to know if the given route has finished loading and the page rendering/processing is done.

So, as the general solution to this is, that you run your long running tasks outside of the Angular zone, my idea was to also run your throttle setTimeout function in a separate, new zone if there is window.Zone available.

Something like:
```
export default function throttle(fn, onThrottle, opts) {
  var context = this;
  var limit = opts.limit;
  var interval = opts.interval;
  var counter = 0;
  var timeoutId;
  var throttleZone = typeof Zone !== 'undefined' && Zone.root.fork({name: 'apm-throttle-zone'});

  return function () {
    counter++;

    if (typeof timeoutId === 'undefined') {
        var timeoutFn = function () {
          timeoutId = setTimeout(function () {
            counter = 0;
            timeoutId = undefined;
          }, interval);
        };

        if (throttleZone) {
          throttleZone.run(timeoutFn)
        } else {
          timeoutFn();
        }
    }

    if (counter > limit && typeof onThrottle === 'function') {
      return onThrottle.apply(context, arguments);
    } else {
      return fn.apply(context, arguments);
    }
  };
}
```

What do you guys think?

pappkamerad:
Fixes #983 

This adds a `runInOwnZone` helper function to wrap tasks that would block Angular stabilization otherwise.

Check the issue for more details.

### DIFF
--- a/packages/rum-core/src/common/after-frame.js
+++ b/packages/rum-core/src/common/after-frame.js
@@ -23,6 +23,8 @@
  *
  */
 
+import { runInOwnZone } from './utils'
+
 const RAF_TIMEOUT = 100
 
 /**
@@ -43,7 +45,7 @@ export default function afterFrame(callback) {
     cancelAnimationFrame(raf)
     setTimeout(callback)
   }
-  const timeout = setTimeout(handler, RAF_TIMEOUT)
+  const timeout = runInOwnZone(() => setTimeout(handler, RAF_TIMEOUT))
 
   const raf = requestAnimationFrame(handler)
 }
--- a/packages/rum-core/src/common/http/fetch.js
+++ b/packages/rum-core/src/common/http/fetch.js
@@ -24,6 +24,7 @@
  */
 
 import { HTTP_REQUEST_TIMEOUT } from '../constants'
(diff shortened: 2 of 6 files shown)


devcorpio:
Hi @pappkamerad,

Thanks a lot for your contribution!

Currently, we don't have enough availability to review your PR - due to project priorities within the team - we'll let you know once we are ready to review it

Thanks for your time,

Alberto

mitchschedulicity:
Would love to see this approved and merged. It is currently a huge obstacle for our automated testing team.

PR #1359, opened 2023-04-03
Migrate main CI to GitHub Actions

amannocci:
## Description

* Migrate main CI to GitHub Actions.
* Fix bundlesize GitHub check.

## Motivation & Context

* We are moving away from Jenkins to GitHub Actions.

## Next steps

* Migration of Benchmark stages.
* Migration of Release stages.
* Better support for coverage reporting.

### DIFF
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -1,18 +1,22 @@
 #!/usr/bin/env bash
-set -xeo pipefail
+set -xo pipefail
 STACK_VERSION=${STACK_VERSION:-8.6.1}
 
 pip install docker-compose>=1.25.4
 
-USER_ID="$(id -u):$(id -g)" \
-NODEJS_VERSION=$(cat .nvmrc) \
-STACK_VERSION=${STACK_VERSION} \
-docker-compose \
-  -f ./dev-utils/docker-compose.yml \
-  --log-level INFO \
-  up \
-  --quiet-pull \
-  --exit-code-from node-puppeteer \
-  --remove-orphans \
-  node-puppeteer
+for i in {1..3};
+do
+  USER_ID="$(id -u):$(id -g)" \
+  NODEJS_VERSION=$(cat .nvmrc) \
+  STACK_VERSION=${STACK_VERSION} \
+  docker-compose \
+    -f ./dev-utils/docker-compose.yml \
+    --log-level INFO \
+    up \
+    --quiet-pull \
+    --exit-code-from node-puppeteer \
+    --remove-orphans \
+    node-puppeteer && break;
+  sleep 5;
+done
 echo "Exit code from docker-compose $?"
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -23,6 +23,11 @@ do
     --exit-code-from node-puppeteer \
     --remove-orphans \
     node-puppeteer && break;
+    status=${?}
+    if [ "${status}" -eq "0" ]; then
+      break
+    fi
   sleep 5;
 done
-echo "Exit code from docker-compose $?"
+echo "Exit code from docker-compose ${status}"
+exit ${status}
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -xo pipefail
-STACK_VERSION=${STACK_VERSION:-8.6.1}
+
+STACK_VERSION=${STACK_VERSION:-"8.6.1"}
+APM_SERVER_PORT=${APM_SERVER_PORT:-"8200"}
+APM_SERVER_URL=${APM_SERVER_URL:-"http://apm-server:8200"}
+KIBANA_URL=${KIBANA_URL:-"http://kibana:5601"}
 
 pip install docker-compose>=1.25.4
 
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -4,6 +4,8 @@ STACK_VERSION=${STACK_VERSION:-8.6.1}
 
 pip install docker-compose>=1.25.4
 
+# Tests are run within the node-puppeteer container and can fails.
+# To avoid flakiness, we retry up to 3 times to run them.
 for i in {1..3};
 do
   USER_ID="$(id -u):$(id -g)" \
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -15,6 +15,9 @@ do
   USER_ID="$(id -u):$(id -g)" \
   NODEJS_VERSION=$(cat .nvmrc) \
   STACK_VERSION=${STACK_VERSION} \
+  APM_SERVER_PORT=${APM_SERVER_PORT} \
+  APM_SERVER_URL=${APM_SERVER_URL} \
+  KIBANA_URL=${KIBANA_URL} \
   docker-compose \
     -f ./dev-utils/docker-compose.yml \
     --log-level INFO \
@@ -22,7 +25,7 @@ do
     --quiet-pull \
     --exit-code-from node-puppeteer \
     --remove-orphans \
-    node-puppeteer && break;
+    node-puppeteer;
     status=${?}
     if [ "${status}" -eq "0" ]; then
       break
--- a/.github/workflows/ci-extends.yml
+++ b/.github/workflows/ci-extends.yml
@@ -1,6 +1,6 @@
 ---
 
-name: ci
+name: ci-extends
 
 on:
   workflow_dispatch:
--- a/.github/workflows/ci-extends.yml
+++ b/.github/workflows/ci-extends.yml
@@ -1,0 +1,81 @@
+---
+
+name: ci
+
+on:
+  workflow_dispatch:
+    inputs:
+      saucelab_test:
+        description: 'Enable run a Sauce lab test'
+        default: true
+        required: false
+        type: boolean
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
+
+# Limit the access of the generated GITHUB_TOKEN
+permissions:
+  contents: read
+  statuses: write
+
+env:
+  DOCKER_ELASTIC_SECRET: secret/observability-team/ci/docker-registry/prod
+  SAUCELABS_SECRET: secret/apm-team/ci/apm-agent-rum-saucelabs@elastic/apm-rum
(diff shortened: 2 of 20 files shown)


dev-utils/test-config.js
```diff
@@ -40,7 +40,7 @@ function getTestEnvironmentVariables() {
     branch: process.env.BRANCH_NAME,
     mode: process.env.MODE,
     sauceLabs: process.env.MODE && process.env.MODE.startsWith('saucelabs'),
-    isJenkins: process.env.JENKINS_URL,
+    isCI: process.env.JENKINS_URL || process.env.GITHUB_API_URL,
```
reakaleek, line 43: We could also use the `CI` environment variable. It's supported in both Jenkins and GitHub workflow. Wdyt?

reakaleek (commented):
We also need to add the status checks if it's a "docs only" PR. Similar to https://github.com/elastic/apm-server/pull/10002.

reakaleek (approved):
LGTM

.ci/scripts/test.sh
```diff
@@ -1,18 +1,22 @@
...
-STACK_VERSION=${STACK_VERSION} \
-docker-compose \
-  -f ./dev-utils/docker-compose.yml \
-  --log-level INFO \
-  up \
-  --quiet-pull \
-  --exit-code-from node-puppeteer \
-  --remove-orphans \
-  node-puppeteer
+for i in {1..3};
```
v1v, line 7: Can you explain what's the reason for this with a comment?
OTOH, do you think a trap will be needed to clean up the previous run? Or will `--remove-orphans` is enough?

.github/workflows/main.yml
```diff
@@ -0,0 +1,156 @@
...
+            });
+
+  test-puppeteer:
+    name: Test Puppeteer
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        stack-version:
+          - '8.0.0-SNAPSHOT'
+          - '8.6.1'
```
v1v, line 78: https://github.com/elastic/apm-agent-rum-js/blob/main/.ci/bump-elastic-stack.yml is also affected

.github/workflows/main.yml
```diff
@@ -0,0 +1,156 @@
...
+  coverage:
+    name: Coverage
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run coverage
+        uses: ./.github/workflows/run-test
+        with:
+          goal: 'coverage'
+          mode: 'none'
```
v1v, line 137: what do you think if not explicitly set since `none` is the default value?

.github/workflows/main.yml
```diff
@@ -0,0 +1,156 @@
...
+          stack-version: '8.6.1'
```
v1v, line 138: updatecli might also need this

.github/workflows/main.yml
```diff
@@ -0,0 +1,156 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run puppeteer tests
+        uses: ./.github/workflows/run-test
+        with:
+          goal: 'test'
+          mode: 'none'
+          scope: ${{ matrix.scope }}
+          stack-version: ${{ matrix.stack-version }}
+          vault-url: ${{ secrets.VAULT_ADDR }}
```
v1v, line 94: This won't work on PRs from forked repos, what do you think by splitting this PR in two, the one without secrets and the one with secrets?

For the latter one, we might need to syncup with the Rum team, I'd say to be supported on merge to main for the time being, and if needed we can support the case for forked PRs, but I'd prefer to be more conservative in favour of non-security access

.github/workflows/run-test/action.yml
```diff
@@ -0,0 +1,77 @@
...
+        secretId: ${{ inputs.vault-secret-id }}
+        method: approle
+        secrets: |
+          ${{ env.SAUCELABS_SECRET }} SAUCE_USERNAME | SAUCE_USERNAME ;
+          ${{ env.SAUCELABS_SECRET }} SAUCE_ACCESS_KEY | SAUCE_ACCESS_KEY
+    - name: Run tests Elastic Stack ${{ matrix.stack-version }} - ${{ matrix.scope }} - ${{ fromJSON('{"none":"Puppeteer"}')[inputs.mode] || inputs.mode }}
+      run: './.ci/scripts/test.sh'
+      shell: bash
+      env:
+        APM_SERVER_PORT: '8200'
```
v1v, line 64: If these env variables are not needed ouside but required by the `./.ci/scripts/test.sh`, then can these en variables be defined within the `./.ci/scripts/test.sh` context?

.github/workflows/main.yml
```diff
@@ -0,0 +1,156 @@
...
+        env:
+          BUNDLESIZE_PASS: ${{ steps.bundlesize.outputs.bundlesize_pass }}
+          BUNDLESIZE_FAIL: ${{ steps.bundlesize.outputs.bundlesize_fail }}
+        with:
+          script: |
+            const { BUNDLESIZE_PASS, BUNDLESIZE_FAIL } = process.env;
+            let state = 'success';
+            if (BUNDLESIZE_FAIL > 0) {
+              state = 'failure';
+            }
```
v1v, line 62: NIT, can `${{ steps.bundlesize.outputs.bundlesize_pass }}` be used within the `script` context? something like


```
...
context: `${context.workflow} / Bundlesize: ${{ steps.bundlesize.outputs.bundlesize_pass }} pass / ${{ steps.bundlesize.outputs.bundlesize_fail }} fail`,
```

v1v (changes requested):

.github/workflows/main.yml
```diff
@@ -0,0 +1,156 @@
+---
+
+name: main
```
v1v, line 3: OTOH, I'm not fully convinced with `main` for the `ci` specific workflows. Somehow it looks like it refers to the `main` branch. What do you think by using `test` or `ci`? I've been using `ci` in the migrations I've been doing but if I recall @reakaleek used `test` in some other places.

What are your thoughts?

.ci/scripts/test.sh
amannocci: I added a comment to explain the logic.
The `--remove-orphans` flag should be enough to clean up everything before retrying.
We were using the same logic in the Jenkins implementation.

.github/workflows/main.yml
reakaleek: I've used `test` in the beginning but slowly started to prefer `ci`. The main reason was the status check being named e.g `test / test`, whereas `ci / test` looked more pleasant.

amannocci: I renamed the workflow to `ci`.

.github/workflows/main.yml
amannocci: It seems possible to use workflow variables within the generated script.
However, I prefer to use env vars everywhere as recommended in the [documentation](https://github.com/actions/github-script#use-env-as-input).

.github/workflows/ci-extends.yml
```diff
@@ -0,0 +1,81 @@
+---
+
+name: ci-extends
```
v1v, line 3: If it runs only the saucelabs integration, maybe `ci-saucelabs` can be more accurate than `ci-extends`?

.github/workflows/ci-extends.yml
```diff
@@ -0,0 +1,81 @@
...
+
+jobs:
+  test-saucelabs:
+    name: Test SauceLabs
+    # Inputs aren't defined on some events
+    # When undefined, we need to emulate the default value
+    if: inputs.saucelab_test == true || inputs.saucelab_test == null
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 1
```
v1v, line 48: can you add a comment explaining the reason for using one concurrency?

.github/workflows/ci-extends.yml
```diff
@@ -0,0 +1,81 @@
...
+        type: boolean
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
+  pull_request:
+    branches:
+      - main
```
v1v, line 21: As mentioned earlier, this won't work on PRs coming from forked repositories.

.github/workflows/ci-extends.yml
```diff
@@ -0,0 +1,81 @@
...
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
+
+# Limit the access of the generated GITHUB_TOKEN
+permissions:
+  contents: read
+  statuses: write
```
v1v, line 29: What's the reason for this write access?

.github/workflows/run-test/action.yml
```diff
@@ -0,0 +1,67 @@
+---
+
+name: common tasks
+description: Install specific JDK and run a command
```
v1v, line 4: JDK is needed?

v1v (changes requested):

.github/workflows/ci-extends.yml
amannocci: Let's remove the `pull_request` trigger and rely on checks as gatekeepers if needed.
We can trigger the workflow manually using the `workflow_dispatch` event.

.github/workflows/ci-extends.yml
amannocci: I removed the unneeded permission.

.github/workflows/run-test/action.yml
```diff
@@ -0,0 +1,67 @@
...
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@v3
+      with:
+        node-version-file: '.nvmrc'
+    - name: Pull and build docker infra
+      run: './.ci/scripts/pull_and_build.sh'
+      shell: bash
+    - uses: hashicorp/vault-action@v2.5.0
+      if: inputs.mode == 'saucelabs'
```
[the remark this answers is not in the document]
amannocci: I added some checks for those inputs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+      - test-puppeteer
+      - coverage
+    steps:
+      - id: check
+        uses: elastic/apm-pipeline-library/.github/actions/check-dependent-jobs@current
+        with:
+          needs: ${{ toJSON(needs) }}
+      - run: ${{ steps.check.outputs.isSuccess }}
+
+  coverage:
```
v1v, line 97: minor NIT: since this job runs in parallel and it's needed by the `all-tests-puppeteer`, maybe it could be moved above so `all-tests-puppeteer` will be the last job since it needs the other ones?

v1v (approved):
👍

.github/workflows/ci.yml
amannocci: I removed the need for lint and coverage. We only need to wait for matrix jobs to generate the right check.

Issue #1345, opened 2023-02-28
ngcc is removed in Angular 16


stefannikolei:
When i install my dependencies i see see following warning when targeting Angular 15:

Processing legacy "View Engine" libraries:
- @elastic/apm-rum-angular [module/esm2015] (https://github.com/elastic/apm-agent-rum-js.git)

So i think that there will be problems with Angular 16 because the ngcc Compiler got removed. 

See --> https://github.com/angular/angular/blob/main/CHANGELOG.md#compiler (https://github.com/angular/angular/pull/49101)

stefannikolei:
Any Updates on this. All developers who wanna update to Angular V16 will be blocked by this. We rather wait for this to be resolved or must abandon elastic as our RUM Monitoring. :(

stefannikolei:
Angular V16 is released. But no words from elastic 😰

drewpost:
Hello to everyone following this issue. Apologies for the slow replies. I can confirm that we intend to deliver this support with an update coming to our RUM-JS agent this summer.

ArthurIbias:
![image](https://github.com/elastic/apm-agent-rum-js/assets/37701365/12bd5ac3-e55f-449a-8402-3a7bc3a3154c)

Are there any plans to update this module with compatibility for Angular v16+?

aikrez:
@drewpost Can we have a more precise timeline than 'this summer' as that could be months from now.
The company I consult for is blocked by this for upgrading their applications.

drewpost:
> @drewpost Can we have a more precise timeline than 'this summer' as that could be months from now.
> The company I consult for is blocked by this for upgrading their applications.

It's about as accurate as I can be at the moment. My best guess would be July BUT that could shift.

A bit of context: The people working on this are also the same team who are delivering our new synthetics product which is going GA in the 8.8 stack release. We anticipate there will be work reacting to things from that release but the plan is that there will be bandwidth to tackle this RUM work in June. Whilst we have stack release cycles and the teams working tend to organise around these, the RUM-JS agent can be released outside of the normal stack releases. It's out plan to ship this updated agent as soon as the work is done to minimise any further delay. 

I hope that helps with the planning. If there's any updates to that schedule either way, I'll update this issue.

devcorpio:
Hi everyone,

The PR with the changes is now merged.

Will add a new comment here once we release the new version of the angular package

Thanks,
Alberto

hm-00:
[...]

PR #1380, opened 2023-06-29
feat(rum-angular)!: support angular ivy

devcorpio:
Related to https://github.com/elastic/apm-agent-rum-js/issues/1345

# Summary

The changes in this PR will make it possible to use Angular RUM integration in Angular **Ivy** Applications, including Angular 16.

E2E tests work properly in each of the browsers we support.

**Angular 16 app using RUM agent WITHOUT the changes**

angular-16-before-change

**Angular 16 app using RUM agent WITH the changes**

angular-16-after-change

--
**The new changes are only compatible with Angular >=12.**

**Note**: From Angular 12 is not possible to disable Ivy

See below the behaviour with older versions:

Angular 9 output:
angular-9

Angular 10 output

angular-10

Angular 11 output
angular-11

## How to test

Bear in mind that there are other ways to test this, but this is my favourite way (closer to the reality, users install packages):

1. Checkout the PR's branch
2. Install [Verdaccio](https://verdaccio.org/docs/installation/) and start it (make sure you are [authenticated](https://verdaccio.org/docs/setup-npm#creating-user/))
3. Publish rum agent in Verdaccio
      - Change the registry value from [lerna.json file](https://github.com/elastic/apm-agent-rum-js/blob/main/lerna.json#L6) to  http://localhost:4873
      - add ```"angular_ivy_support"```in the [allowBranch](https://github.com/elastic/apm-agent-rum-js/blob/main/lerna.json#L9) property
      - execute ```npx lerna publish --no-git-tag-version --no-push```
4. You can confirm that everything has been published by going to http//localhost:4873
5. In your angular app install the RUM angular package: ```npm install @elastic/apm-rum-angular@latest --registry http://localhost:4873```

### DIFF
--- a/packages/rum-angular/package.json
+++ b/packages/rum-angular/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elastic/apm-rum-angular",
-  "version": "2.1.7",
+  "version": "2.1.7-lala-3",
   "description": "Elastic APM Real User Monitoring for Angular applications",
   "homepage": "https://www.elastic.co/guide/en/apm/agent/rum-js/current/index.html",
   "license": "MIT",
--- a/packages/rum-angular/package.json
+++ b/packages/rum-angular/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elastic/apm-rum-angular",
-  "version": "2.1.7-lala-3",
+  "version": "2.1.7",
   "description": "Elastic APM Real User Monitoring for Angular applications",
   "homepage": "https://www.elastic.co/guide/en/apm/agent/rum-js/current/index.html",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


vigneshshanmugam (commented):
Alberto and myself had a chat offline, we will update the E2E to angular 16 to make it easier for users to reference our example.

sergey-tihon:
May we expect some alpha npm package any time soon to test Ivy support?

devcorpio:
> Alberto and myself had a chat offline, we will update the E2E to angular 16 to make it easier for users to reference our example.

We had a follow-up, we will not do this, the reasons:

- Angular > 12 drop support for IE11 (which wasn't particularly a problem, this wasn't the blocker)
- Angular 16 doesn't support Node.js 14 - which is the version we use in the agent, so we needed to upgrade to Node.js 16
- Update to Node.js implies supporting NPM 7, which opens a new can of worms

That's why it was decided to do this in a separate tech debt issue, now it's the right time to do this update

devcorpio:
@sergey-tihon, I added a "how to test" section in the PR's description

vigneshshanmugam (approved):
LGTM 🎉 , Tested with Angular 16 and works as expected.

Issue #1360, opened 2023-04-04
Breakdown Timings are reported in milliseconds not microseconds


simonhearne:
Breakdown timings are [reported as](https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/performance-monitoring/breakdown.js#L145) millisecond values in the `span.self_time.sum.us` field.

These values should be microseconds to match the [APM field definitions](https://www.elastic.co/guide/en/apm/server/current/exported-fields-apm-transaction-metrics.html#_self_time).

drewpost:
I can confirm that we intend to remedy this issue with an update coming to our RUM-JS agent this summer.

devcorpio:
This has been released: https://www.elastic.co/guide/en/apm/agent/rum-js/current/release-notes-5.x.html#release-notes-5.13.0

PR #1381, opened 2023-07-04
fix(rum): report breakdown span time in microseconds

devcorpio:
Fixes https://github.com/elastic/apm-agent-rum-js/issues/1360

# Summary

From now on, the breakdown metric `span.self_time.sum.us` will be calculated in microseconds.

By the way, although this is not a 100% sure breaking change - we don't know what users might do with the ingested data), it can be a "potentially breaking change", so we should indicate it.

For instance, I like how the Java APM agent team expresses that concept in their release notes:

Screenshot 2023-07-04 at 19 03 02

Example [here](https://www.elastic.co/guide/en/apm/agent/java/current/release-notes-1.x.html#_potentially_breaking_changes)

**Important:** 

You will see that I've not updated the calculation for `transaction.duration.sum.us`. The reason is that almost two years ago, [APM server stopped recording that value](https://github.com/elastic/apm-server/pull/6180). In fact, everything related to transaction breakdown metrics [was removed from all APM agents.](https://github.com/elastic/apm/issues/505)

I believe that for us is a good time to do so since we are solving this bug.

I will do that in a different PR, so we can mention that separately in the release notes and also reference to reference it [here](https://github.com/elastic/apm/issues/505)

Edit: The related PR: https://github.com/elastic/apm-agent-rum-js/pull/1382

### DIFF
--- a/packages/rum-core/src/performance-monitoring/breakdown.js
+++ b/packages/rum-core/src/performance-monitoring/breakdown.js
@@ -142,7 +142,7 @@ function getSpanBreakdown(
     span: details,
     samples: {
       'span.self_time.count': getValue(count),
-      'span.self_time.sum.us': getValue(duration)
+      'span.self_time.sum.us': getValue(duration * 1000)
     }
   }
 }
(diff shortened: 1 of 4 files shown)


vigneshshanmugam (approved):
LGTM, I like the idea of expressing as a potential breaking change as per the suggestion 💯

PR #1382, opened 2023-07-04
fix(rum): stop recording transaction breakdown metrics

devcorpio:
Follow up / Continuation of https://github.com/elastic/apm-agent-rum-js/pull/1381

# Summary

Almost two years ago,  [APM server stopped](https://github.com/elastic/apm-server/issues/6176) recording **transaction.duration.sum.us** and **transaction.duration.count**. 

The [context](https://github.com/elastic/apm/issues/505) for that is this:

> Agents capture and send transaction.duration.sum.us and transaction.duration.count metrics to apm-server, which we have been dutifully recording as part of transaction breakdown metrics. The original intention behind these was to enable metrics-based UI.
> 
> We now capture transaction latency distributions in apm-server, so the above mentioned metrics are no longer needed. They're not used anywhere, so let's stop recording them.
> 
> Additionally, **transaction.breakdown.count** is also not used (anymore). See https://github.com/elastic/apm/issues/505#issuecomment-943325753 for more details.

You will also see that all agents implemented that change in that period, too:

Screenshot 2023-07-04 at 18 51 12

Like all the other agents, I wouldn't treat this as a breaking change. Although it would **be pretty interesting** to do what the APM java team does, treat this as a "potentially breaking change", since we cannot be sure what users do with the ingested data. 

example: https://www.elastic.co/guide/en/apm/agent/java/current/release-notes-1.x.html#_potentially_breaking_changes_7

There are other agents (Node.js for instance) that treated this as a [bug](https://www.elastic.co/guide/en/apm/agent/nodejs/current/release-notes-3.x.html#_bug_fixes_22) and other ones (like the .NET) that treated this as a [feature](https://www.elastic.co/guide/en/apm/agent/dotnet/current/release-notes-1.x.html#_features_11) 

 If you disagree, let me know (that's also why I created this separate PR, to avoid blocking the [other one](https://github.com/elastic/apm-agent-rum-js/pull/1381) in case your opinion differs)

P.S. Tested with apiVersion 2 and 3

### DIFF
--- a/packages/rum-core/src/common/compress.js
+++ b/packages/rum-core/src/common/compress.js
@@ -290,30 +290,14 @@ export function compressError(error) {
 
 export function compressMetricsets(breakdowns) {
   return breakdowns.map(({ span, samples }) => {
-    const isSpan = span != null
-    if (isSpan) {
-      return {
-        y: { t: span.type },
-        sa: {
-          ysc: {
-            v: samples['span.self_time.count'].value
-          },
-          yss: {
-            v: samples['span.self_time.sum.us'].value
-          }
-        }
-      }
-    }
     return {
+      y: { t: span.type },
       sa: {
-        xdc: {
-          v: samples['transaction.duration.count'].value
+        ysc: {
+          v: samples['span.self_time.count'].value
         },
-        xds: {
-          v: samples['transaction.duration.sum.us'].value
-        },
-        xbc: {
-          v: samples['transaction.breakdown.count'].value
+        yss: {
+          v: samples['span.self_time.sum.us'].value
         }
       }
     }
--- a/scripts/ci-bundlesize.sh
+++ b/scripts/ci-bundlesize.sh
@@ -12,8 +12,8 @@ BUNDLESIZE_PATH="${WORKSPACE}/bundlesize.txt"
 npm run bundlesize | tee "${BUNDLESIZE_PATH}" || true
 
 # Extract passed and failed
-BUNDLESIZE_PASS=$(grep "✔" "${BUNDLESIZE_PATH}" | wc -l)
-BUNDLESIZE_FAIL=$(grep "✖" "${BUNDLESIZE_PATH}" | wc -l)
+BUNDLESIZE_PASS=$(grep "✔" "${BUNDLESIZE_PATH}" | wc -l) || true
+BUNDLESIZE_FAIL=$(grep "✖" "${BUNDLESIZE_PATH}" | wc -l) || true
 if [ ! -z "${GITHUB_OUTPUT}" ]; then
   echo "bundlesize_pass=${BUNDLESIZE_PASS}" >> "${GITHUB_OUTPUT}"
   echo "bundlesize_fail=${BUNDLESIZE_FAIL}" >> "${GITHUB_OUTPUT}"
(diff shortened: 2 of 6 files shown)


packages/rum-core/src/common/compress.js
```diff
@@ -289,9 +289,9 @@ export function compressError(error) {
 }
 
 export function compressMetricsets(breakdowns) {
-  return breakdowns.map(({ span, samples }) => {
-    const isSpan = span != null
-    if (isSpan) {
+  return breakdowns
+    .filter(({ span }) => span != null)
```
devcorpio, line 293: now that I'm thinking, this filter shouldn't be necessary anymore, since now all breakdowns are for spans, I will check this.

devcorpio:
Hi @amannocci,

I've noticed something with the bundlesize step (within the ci / Lint step)

If the [output](https://github.com/elastic/apm-agent-rum-js/actions/runs/5456920221/jobs/9930360710?pr=1381) of the bundlesize is (1 check passed, 1 check failed), things work as expeected

Screenshot 2023-07-04 at 20 02 25


But if the [output](https://github.com/elastic/apm-agent-rum-js/actions/runs/5457295484/jobs/9931144869?pr=1382)  is (2 checks passed):

Screenshot 2023-07-04 at 20 02 55

Then the .sh throws this error:

`./scripts/ci-bundlesize.sh: Error on line 16: BUNDLESIZE_FAIL=$(grep "✖" "${BUNDLESIZE_PATH}" | wc -l)`

Do you know what is going on?

scripts/ci-bundlesize.sh
```diff
@@ -12,8 +12,8 @@ BUNDLESIZE_PATH="${WORKSPACE}/bundlesize.txt"
 npm run bundlesize | tee "${BUNDLESIZE_PATH}" || true
 
 # Extract passed and failed
-BUNDLESIZE_PASS=$(grep "✔" "${BUNDLESIZE_PATH}" | wc -l)
-BUNDLESIZE_FAIL=$(grep "✖" "${BUNDLESIZE_PATH}" | wc -l)
+BUNDLESIZE_PASS=$(grep "✔" "${BUNDLESIZE_PATH}" | wc -l) || true
+BUNDLESIZE_FAIL=$(grep "✖" "${BUNDLESIZE_PATH}" | wc -l) || true
```
amannocci, line 16: It's due to the bash flag.
We can disable it before trying to grep.

scripts/ci-bundlesize.sh
```diff
@@ -12,8 +12,8 @@ BUNDLESIZE_PATH="${WORKSPACE}/bundlesize.txt"
 npm run bundlesize | tee "${BUNDLESIZE_PATH}" || true
 
 # Extract passed and failed
-BUNDLESIZE_PASS=$(grep "✔" "${BUNDLESIZE_PATH}" | wc -l)
-BUNDLESIZE_FAIL=$(grep "✖" "${BUNDLESIZE_PATH}" | wc -l)
+BUNDLESIZE_PASS=$(grep "✔" "${BUNDLESIZE_PATH}" | wc -l) || true
```
amannocci, line 15:
```suggestion
set +e
BUNDLESIZE_PASS=$(grep "✔" "${BUNDLESIZE_PATH}" | wc -l)
```

scripts/ci-bundlesize.sh
amannocci, line 16:
```suggestion
BUNDLESIZE_FAIL=$(grep "✖" "${BUNDLESIZE_PATH}" | wc -l)
set -e
```

vigneshshanmugam (approved):
LGTM 

Given RUM agent does not capture breakdown metrics by default, Lets not call it as breaking change and go ahead with a patch release as other APM agents.

Issue #1286, opened 2022-09-29
improve naming strategy for user-interaction transactions


devcorpio:
# Context

Currently, users can set a custom name for user-interaction transactions via the [data-transaction-name ](https://www.elastic.co/guide/en/apm/agent/rum-js/5.x/supported-technologies.html#user-interactions)attribute.

For instance:

`
<button data-transaction-name="add to cart">cart icon</button>
`

Clicking on that button would generate a transaction with the name **Click - add to cart**

# To improve

The current technique does not work well with a nested html structure:

```
<button data-transaction-name="my transaction name">
  <span>Some text</span>
  <svg>button icon...</svg>
</button>
```

The user would expect to see **Click - my transaction name** but instead the transaction's name that they will see is either **Click - span** or **Click - svg**

There are workarounds like setting **data-transaction-name** in each node, but that can be cumbersome and even affect the performance of the website since the payload might be bigger. 

Not only that, some libraries generate extra html code dynamically making then impossible or **pretty difficult** to modify. Just for illustration, when using [Material UI](https://mui.com/) it's common for a developer to write code like this one: 

`<DeleteIcon data-transaction-name="delete" />`

But what we see in the browser is a nested structure:

```
<svg ... data-transaction-name="DeleteSltd">
      <path d="..."></path>
</svg>
```

Causing then the naming issues commented above.

# Why is it important to improve this?

This issue has been reported multiple times by users and customers and for quite a while. [This](https://elastic.slack.com/archives/C02N6H8HPM2/p1654248897066909) slack thread is just an example of those reports.

nicolasguilhem:
In addition, many frameworks like Angular and React generate code dynamically (like the example in description with Material UI).

Would be a great feature.

As a suggestion, looking recursivly in parent of target for attribute `data-transaction-name` in https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/common/observers/page-clicks.js#L77 should do the job ?

devcorpio:
it has been released: https://www.elastic.co/guide/en/apm/agent/rum-js/current/release-notes-5.x.html#release-notes-5.13.0

PR #1352, opened 2023-03-16
fix(rum-core): add `'click'` to instrumentation types

jlp-craigmorten:
Adds the `'click'` value to the `InstrumentationTypes` type.

Re-ordered to mirror the documentation at https://www.elastic.co/guide/en/apm/agent/rum-js/current/configuration.html#disable-instrumentations

PR #1390, opened 2023-07-10
feat(rum-core): enhance user transaction name obtaining

devcorpio:
Related to https://github.com/elastic/apm-agent-rum-js/issues/1286

# Summary

We have improved the user-interaction transaction name obtaining. From now on, if a user performs click on an element from a nested HTML structure like this:

```
<button data-transaction-name="my transaction name">
  <span>Some text</span>
  <svg>button icon...</svg>
</button>
```

The users will always get  `Click - my transaction name ` rather than `Click - span` or `Click - svg`

For more details see the [issue](https://github.com/elastic/apm-agent-rum-js/issues/1286)

**Important**: 

The agent will keep applying the older "algorithm" in browsers not supporting the closest method. E.g. IE11

In addition, this will work if the target is inside a button or link - the ones we define as interactive selectors. See [the associated tests](https://github.com/elastic/apm-agent-rum-js/blob/933ffa6345feb86bebce7024a4c0a9b87c314ab8/packages/rum-core/test/common/observers/page-clicks.spec.js#L145,L193) in case you need more context

### DIFF
--- a/.github/workflows/ci-saucelabs.yml
+++ b/.github/workflows/ci-saucelabs.yml
@@ -16,6 +16,12 @@ on:
     paths-ignore:
       - '**.md'
       - '**.asciidoc'
+  pull_request:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
 
 # Limit the access of the generated GITHUB_TOKEN
 permissions:
--- a/.github/workflows/ci-saucelabs.yml
+++ b/.github/workflows/ci-saucelabs.yml
@@ -41,7 +41,11 @@ jobs:
     name: Test SauceLabs
     # Inputs aren't defined on some events
     # When undefined, we need to emulate the default value
-    if: inputs.saucelab_test == true || inputs.saucelab_test == null
+    if: |
+      inputs.saucelab_test == true || inputs.saucelab_test == null
+      && (github.event_name != 'pull_request'
+        || github.event_name == 'pull_request'
+        && github.event.pull_request.head.repo.fork == false)
     runs-on: ubuntu-latest
     strategy:
       # Saucelab doesn't support running test in //.
--- a/packages/rum-core/src/common/observers/page-clicks.js
+++ b/packages/rum-core/src/common/observers/page-clicks.js
@@ -104,7 +104,7 @@ function buildTransactionName(target) {
   }
 
   // Just use the tagName
-  return tagName.toLowerCase()
+  return tagName
 }
 
 function findCustomTransactionName(target) {
--- a/packages/rum-core/src/common/observers/page-clicks.js
+++ b/packages/rum-core/src/common/observers/page-clicks.js
@@ -74,20 +74,7 @@ function getTransactionMetadata(target) {
     context: null
   }
 
-  const tagName = target.tagName.toLowerCase()
-
-  // use custom html attribute 'data-transaction-name' - otherwise fall back to "tagname" + "name"-Attribute
-  let transactionName = tagName
-  if (!!target.dataset.transactionName) {
-    transactionName = target.dataset.transactionName
-  } else {
-    const name = target.getAttribute('name')
-    if (!!name) {
-      transactionName = `${tagName}["${name}"]`
-    }
-  }
-
-  metadata.transactionName = transactionName
+  metadata.transactionName = buildTransactionName(target)
 
   let classes = target.getAttribute('class')
   if (classes) {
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-saucelabs.yml
```diff
@@ -16,6 +16,12 @@ on:
     paths-ignore:
       - '**.md'
       - '**.asciidoc'
+  pull_request:
```
devcorpio, line 19: enabling saucelabs e2e.

packages/rum/src/index.d.ts
```diff
@@ -192,6 +192,7 @@ type TransactionEvents = 'transaction:start' | 'transaction:end'
 type InstrumentationTypes =
   | 'page-load'
   | 'eventtarget'
+  | 'click'
```
devcorpio, line 195: add missing type.

Related to: https://github.com/elastic/apm-agent-rum-js/pull/1352#event-9772322079

.github/workflows/ci-saucelabs.yml
```diff
@@ -41,7 +41,11 @@ jobs:
     name: Test SauceLabs
     # Inputs aren't defined on some events
     # When undefined, we need to emulate the default value
-    if: inputs.saucelab_test == true || inputs.saucelab_test == null
+    if: |
+      inputs.saucelab_test == true || inputs.saucelab_test == null
+      && (github.event_name != 'pull_request'
+        || github.event_name == 'pull_request'
+        && github.event.pull_request.head.repo.fork == false)
```
devcorpio, line 48: thanks for this change!

.github/workflows/ci-saucelabs.yml
```diff
@@ -35,7 +41,11 @@ jobs:
     name: Test SauceLabs
     # Inputs aren't defined on some events
     # When undefined, we need to emulate the default value
-    if: inputs.saucelab_test == true || inputs.saucelab_test == null
+    if: |
+      inputs.saucelab_test == true || inputs.saucelab_test == null
+      && (github.event_name != 'pull_request'
```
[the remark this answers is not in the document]
devcorpio: thanks! 🎉

packages/rum-core/src/common/observers/page-clicks.js
```diff
@@ -96,3 +83,34 @@ function getTransactionMetadata(target) {
...
+
+  const tagName = target.tagName.toLowerCase()
+  // "tagname" + "name"-Attribute
+  const name = target.getAttribute('name')
+  if (!!name) {
+    return `${tagName}["${name}"]`
+  }
+
+  // Just use the tagName
+  return tagName.toLowerCase()
```
vigneshshanmugam, line 104: toLowerCase not required. Already lowercased.

packages/rum-core/src/common/observers/page-clicks.js
```diff
@@ -96,3 +83,34 @@ function getTransactionMetadata(target) {
...
+}
+
+function findCustomTransactionName(target) {
+  if (target.closest) {
+    // Leverage closest API to traverse the element and its parents
+    const element = target.closest('[data-transaction-name]')
```
vigneshshanmugam, line 110: How far do we want to traverse? We should at-least restrict to a known tags that can cause the click effect to be on safer side. 
* Button
* Anchor tags

Doing this for every observe on click is expensive and not worth it.

devcorpio: Vignesh and I had a chat offline:

We will perform the following optimization:

Instead of 
```target.closest('[data-transaction-name]')```

We will do:

```target.closest('a[data-transaction-name], button[data-transaction-name]')```

This way we make sure the API just looks for interactive selectors and ignore the rest of them.

File: packages/rum-core/src/common/observers/page-clicks.js (hunk 5 above)
Comment (devcorpio) [reply packages/rum-core/src/common/observers/page-clicks.js:107]: leftover 🍕

File: dev-utils/test-config.js [hunk 7]
```diff
@@ -67,6 +67,12 @@ function getGlobalConfig(packageName = 'rum') {
  * The list below is based purely on the market share distribution.
  */
 function getDefaultBrowsers() {
+  return [
```
devcorpio, line 70: I'm checking if saucelabs works with just one browser

packages/rum-core/test/common/observers/page-clicks.spec.js
```diff
@@ -139,4 +140,58 @@ describe('observePageClicks', () => {
...
+        {
+          name:
+            'should not use customTransactionName when click target is not descendant from <button> nor <a>',
+          parentTagName: 'div',
+          customTransactionName: 'div-custom-name-html',
+          expected: 'Click - span'
+        }
+      ].forEach(({ parentTagName, customTransactionName, expected }) => {
+        it('should respect the custom transaction name in a nested HTML structure', () => {
+          unobservePageClicks = observePageClicks(transactionService)
```
vigneshshanmugam, line 172: shouldnt we call `unObserve` at the end?

vigneshshanmugam (approved):
LGTM 🎉

packages/rum-core/test/common/observers/page-clicks.spec.js
devcorpio: it's being called in line 51:

```
  afterEach(() => {
    document.body.removeChild(containerElement)
    // Remove the added listener since window object is shared among all tests
    unobservePageClicks()
  })
```

PR #1391, opened 2023-07-17
ci: support dry-run within release workflow

amannocci:
* Support release dry-run

* Fully test release process

## How has this been tested?
* https://github.com/elastic/apm-agent-rum-js/actions/runs/5578333578


lerna.json
```diff
@@ -10,7 +10,7 @@
       "conventionalCommits": true,
       "message": "chore(release): publish",
       "changelogPreset": "conventionalcommits",
-      "gitRemote": "upstream",
+      "gitRemote": "origin",
```
devcorpio, line 13: question:

Why is this change needed?

amannocci: It seems that the upstream git remote isn't defined in GitHub Actions.

devcorpio:
LGTM!, thanks for the dry run!

devcorpio (approved):

package-lock.json
vigneshshanmugam, line 1: Looking at this big change, I believe something is going on with this file, Did you do `nvm use` before running npm install?

devcorpio: it's curious, because the package-lock version is still 1, so no new npm version did this.

gonna check locally

.github/workflows/release.yml
```diff
@@ -34,44 +45,46 @@ jobs:
...
+        env:
+          DRY_RUN: ${{ inputs.dry-run }}
+        run: |
+          if [ "${DRY_RUN}" == "false" ]; then
+            echo 'ELASTIC_CDN_BUCKET_NAME=apm-rum-357700bc' >> ${GITHUB_ENV}
+            echo 'ELASTIC_CDN_CREDENTIALS=secret/gce/elastic-cdn/service-account/apm-rum-admin' >> ${GITHUB_ENV}
+          else
+            echo 'ELASTIC_CDN_BUCKET_NAME=oblt-apm-agent-rum-js-ci' >> ${GITHUB_ENV}
+            echo 'ELASTIC_CDN_CREDENTIALS=secret/observability-team/ci/service-account/apm-agent-rum-js' >> ${GITHUB_ENV}
+          fi
```
kuisathaverat, line 77: I am not sure of the goal of this step, if DRY_RUN=true all the rest of the steps that use those env vars will fail. The env vars can be defined at the job level in a `env` block.

amannocci: The `DRY_RUN` env var is used to define which secrets we will use.
We could use either the production secrets or the CI secrets.
Those secrets contain similar values but refer to different accounts.

package-lock.json
amannocci: I used NodeJS 14.21.3 with NPM 6.14.18.

scripts/ci-release.mjs
```diff
@@ -0,0 +1,146 @@
...
+    raiseError("Failed to publish github release")
+  }
+}
+
+// Entrypoint
+;(async () => {
+  try {
+    await main()
+  } catch (err) {
+    console.log(err)
```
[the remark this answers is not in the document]
amannocci: I added a non-zero exit in that case to avoid any issues.

reakaleek (approved):

package-lock.json
devcorpio: For visibility purposes:

The test https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-vue/test/specs/error-handler.spec.js#L32 was failing, both in CI and locally

The plan to make it work again has been:

- add the exact version (14.20.1) of Node.js in `.nvmrc`, to prevent the npm install from messing with the package-lock.json
- remove `execa` and `npm-auth-to-token` packages
- restore `package-lock.json` from `main`
- install `execa` and `npm-auth-to-token` again

devcorpio: I set the node version to 14 the version in .nvmrc again to avoid this in CI.

Screenshot 2023-07-24 at 11 15 17

This is a tech debt that eventually we will need to tackle!

PR #1399, opened 2023-07-25
chore: prevent slack notification in dry-run mode

devcorpio:
# Summary

The PR changes prevent the dry-run release process from sending Slack notification

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -123,7 +123,7 @@ jobs:
           process_gcloudignore: false
 
   status:
-    if: inputs.dry-run == null || !inputs.dry-run
+    if: inputs.dry-run != null && !inputs.dry-run
     needs:
       - release
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -123,7 +123,7 @@ jobs:
           process_gcloudignore: false
 
   status:
-    if: always()
+    if: inputs.dry-run == null || !inputs.dry-run
```
amannocci, line 126:
```suggestion
    if: inputs.dry-run != null && !inputs.dry-run
```
The `null` value is considered as dry-run mode also.

amannocci (changes requested):

.github/workflows/release.yml
devcorpio: done!

vigneshshanmugam (approved):

amannocci (approved):

Issue #1279, opened 2022-09-21
Incorrect transaction duration for pages after redirect


trainings:
We use token handler pattern for enforcing authentication for web applications.
Steps:
1. user types abc.xxx.com and sends a request
2. on the server side it checks for the presence of a token in the cookie. if there is no token, the user is redirected to authorization service auth.xxx.com
3. user enters his credentials
4. user redirected back to the abc.xxx.com

everything is standard, but:

page load transaction measured from point 1 (not 4)
on the sample of a transaction, gaps for items 2 & 3

devcorpio:
Hi @trainings,

Thanks for reaching out!

Bear in mind that the RUM **[page-load](https://www.elastic.co/guide/en/apm/agent/rum-js/current/supported-technologies.html#page-load-metrics)** transaction will end just after the browser fires the page [load event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event). 

The agent do not keep state between **full-page** refreshes which means that every page will have each own page-load transaction. 
Important to mention that it's possible to [disable](https://www.elastic.co/guide/en/apm/agent/rum-js/current/configuration.html#disable-instrumentations) the page-load instrumentation so you can **start** and **end** such transaction [manually](https://www.elastic.co/guide/en/apm/agent/rum-js/5.x/custom-transactions.html), that tends to be common in SPAs given that you might want to decouple "my SPA is fully loaded/bootstrapped" from the load event the browser fires. On your particular scenario this might not help since it seems that you are doing redirects that force the browser to fully refresh.

 In a future version of the agent we will introduce the concept if user sessions which will provide (among other things) with the possibility of seeing every page-load transaction grouped together under the same session id.


Thanks,
Alberto

trainings:
I feel like I've been misunderstood.
Now, if an anonymous user, in the process of gaining access to a web page, goes to an external service (e.g. keycloak) for authentication, then all redirects (to keycloak and back to my service callback, etc.) time are counted as a transaction time.

For example - I have a page with page-load-time 3sec.
And we have the following:
User is opening a browser and typing xxx.com and sending it
My service processes the request and checks that the user isn't authenticated
My service redirect (302 http code) the user to the IDP (keycloack) on another domain auth.com
The user gets to this page and enters his password there (5 sec) and submits it
Auth service verifies user credentials and redirect user back to the xxx.com/callback with a authorization code
[...]

PR #1400, opened 2023-07-27
feat(rum-core): add redirect span for same-origin redirect

devcorpio:
Related to https://github.com/elastic/apm-agent-rum-js/issues/1279

# Summary

A new span named **Redirect** will be available if a **same-origin** redirect has existed. 

Additionally, this PR also fixes a bug. Before these changes, the agent did not consider the redirection timing when calculating **navigation marks**.

Marks were wrongly calculated, but the issue was not obvious because the new span wasn't there.

wrong-marks

With the new span, you can see how the marks were clearly badly calculated:

spans-but-wrong-marks

And after applying all changes, you can see now, the proper waterfall:

full-redirect-marks

**Note**: None of these changes apply for **cross-origin** redirects since the browser doesn't expose that info (at least not explicitly)

Btw, user timing marks have been also tested:

custom-marks


packages/rum-core/src/performance-monitoring/navigation/capture-navigation.js
devcorpio, line 1: The name is the same, because "capture-navigation" has been part of our muscle memory for a very long time 💪

packages/rum-core/test/performance-monitoring/capture-navigation.spec.js
```diff
@@ -23,21 +23,26 @@
  *
  */
 
+import * as navTiming from '../../src/performance-monitoring/navigation/navigation-timing-spans'
```
devcorpio, line 26: this has been extremely helpful to test the **baseTime** logic 🎉 , which it was not tested before this PR

packages/rum-core/src/performance-monitoring/metrics.js
```diff
@@ -265,7 +270,9 @@ export function captureObserverEntries(list, { isHardNavigation, trStart }) {
    * To avoid capturing the unload event handler effect
    * as part of the page-load transaction duration
    */
-  const unloadDiff = timing.fetchStart - timing.navigationStart
+  const unloadDiff = isRedirectInfoAvailable(timing)
```
devcorpio, line 273: this makes sure the FCP startTime includes the `redirectTime`, otherwise the mark would show up in the bad place in the UI

packages/rum-core/src/performance-monitoring/navigation/capture-navigation.js
```diff
@@ -0,0 +1,125 @@
...
+      })
+    }
+    /**
+     * must be zero otherwise the calculated relative _start time would be wrong
+     */
+    const trStart = 0
+    transaction._start = trStart
+
+    const timings = PERF.timing
+    const baseTime = isRedirectInfoAvailable(timings)
```
devcorpio, line 75: also, this change makes possible that spans take the redirect duration into consideration

packages/rum-core/src/performance-monitoring/navigation/marks.js
```diff
@@ -0,0 +1,124 @@
...
+   */
+  if (
+    fetchStart >= navigationStart &&
+    responseStart >= fetchStart &&
+    responseEnd >= responseStart
+  ) {
+    const marks = {}
+    NAVIGATION_TIMING_MARKS.forEach(function (timingKey) {
+      const m = timing[timingKey]
+      if (isRedirectInfoAvailable(timing)) {
```
devcorpio, line 109: also, this change makes it possible that navigation marks to take the duration into consideration.

This change + [this other one](https://github.com/elastic/apm-agent-rum-js/pull/1400/files#r1276504852) make possible to show up all marks properly in the UI (like the third screenshot of the PR description)

packages/rum-core/src/common/utils.js
```diff
@@ -431,6 +431,11 @@ function isBeaconInspectionEnabled() {
   return false
 }
 
+// redirect info is only available for same-origin redirects
+function isRedirectInfoAvailable(timing) {
+  return timing.redirectStart > 0
```
vigneshshanmugam, line 436: Are there cases when this data can be buggy? Like case with fetchStart before redirectStart or being in unix vs zero based offset timestamps. Check the buggy navigation data test cases to see if its relevant. Not a blocker though, just want to call out.

vigneshshanmugam (commented):
LGTM, great to see new tests 🎉 

Just want to make sure there are no buggy navigation data with Redirects, Otherwise we should be good to go.

packages/rum-core/src/performance-monitoring/navigation/marks.js
devcorpio: done! and also added the comment

packages/rum-core/src/common/utils.js
devcorpio: thanks for highlighting this.

I've not found anything that might affect the agent, the only bug I found is [this](https://bugs.chromium.org/p/chromium/issues/detail?id=805831), but doesn't seem relevant for us and nobody has reproduced it after Chromium 30 🧐

vigneshshanmugam (approved):
LGTM 🎉

PR #1401, opened 2023-07-31
ci: correct inputs coerces

amannocci:
* Correct input coerce
* https://docs.github.com/en/actions/learn-github-actions/expressions#operators

* The release process isn't working as expected.


devcorpio (approved):
LGTM

PR #1404, opened 2023-07-31
ci: correct github token configuration

amannocci:
* The auto-generate github token isn't enough to push on protected branches.

* We should be able to push on protected branches for release process.


devcorpio (approved):
LGTM

PR #1406, opened 2023-08-03
ci: avoid to persist the wrong GitHub token

amannocci:
* Use the PAT GitHub Token to checkout the repository.
* It will persist the PAT token instead of the default one.

* lerna isn't able to push on protected branches.


devcorpio (approved):
LGTM

Issue #1176, opened 2022-03-03
support react-router-dom v17


vrecan:
When using v17 of react-router-dom I get the following error:
```
apm Uncaught Error: "[ApmRoute2]" is not a <Route> component. 
All component children of <Routes> must be a <Route> or <React.Fragment>
```

Here is the code example, this works when I switch out the apmRoute for a normal Route

```
  return (
    <>
      {!isIdle && <Heartbeat setIsLoggedIn={setIsLoggedIn} />}
      <Router>
        <Routes>
          <ApmRoute path="/login" element={<Login />} />
          <ApmRoute path="/" element={<Desktop />}>
            <ApmRoute index element={<AgentDashboard />} />
            <ApmRoute
              path="/conversations/:conversationId"
              element={<CurrentConversation />}
            />
            <ApmRoute path="/conversations/new" element={<NewConversation />} />
          </ApmRoute>
        </Routes>
      </Router>
    </>
  );
};
```

Here are my imports:
```
import { useState } from 'react';
import { QueryClientProvider } from 'react-query';
import { ReactQueryDevtools } from 'react-query/devtools';
import { BrowserRouter as Router, Routes, Route } from 'react-router-dom';
import { ApmRoute } from '@elastic/apm-rum-react';

import { Login } from './components';
import { AgentDashboard } from './domains/AgentDashboard';
import { CurrentConversation } from './domains/CurrentConversation';
import { ModalProvider } from './domains/Modal';
import { NewConversation } from './domains/NewConversation';
import { Desktop } from './domains/Desktop';
import { queryClient, state, hooks } from './utils';
```

From what I can tell apmRoute isn't extending Route.

devcorpio:
Hi @vrecan ,

Thanks for reaching out. At this moment the agent does not have out-of-the-box support for `react-router-dom` v6. It supports v4 and v5.

In this [issue](https://github.com/elastic/apm-agent-rum-js/issues/1163) we offer possible tradeoffs that might helpful for you.

We don't have plans **yet** to start the development to add the v6 support. 

Thanks,
Alberto

jbournonville:
Any updates on plan to make devs for react-router v6 support ?

puunukk:
Any updates on plan for _react-router-dom_ v6 support?

LeaTaka:
Hey @drewpost , 
Do you have an update on your 'RUM-JS agent release' delivery planning regarding v6 support?

devcorpio:
[...]

PR #1410, opened 2023-08-07
feat(rum-react)!: support react-router 6

devcorpio:
Related to: https://github.com/elastic/apm-agent-rum-js/issues/1176

# Summary

This PR makes it possible to instrument applications with react-router 6.

The API we expose will look like the following snippet:

```
      <YourRouter>
        <ApmRoutes>
          <Route path="/home" element={<Component name="Elastic" />} />
          <Route path="/new" element={<div>new</div>} />
        </ApmRoutes>
      </YourRouter>
```

All the routes defined within `<ApmRoutes>` will be instrumented.

**IMPORTANT**:

This PR will not instrument the mechanism named [RouteProvider](https://reactrouter.com/en/main/upgrading/v6-data)  and there are no plans to do so in the midterm.

**Note**: <strike>E2E are still pending to finish</strike>

react-router 6 [doesn't support IE11.](https://github.com/remix-run/react-router/issues/8220#issuecomment-961326123) - E2E tests for React will no longer be executed for that browser.


packages/rum-react/src/get-apm-route.js
```diff
@@ -1,86 +0,0 @@
...
-
-import React from 'react'
-import { Route } from 'react-router-dom'
-import { getWithTransaction } from './get-with-transaction'
-
-/**
- * If the path/name is given as array, use the computed path
- * to get the current transaction name
- */
-function getTransactionName(name, props) {
```
devcorpio, line 34: FYI, react-router 6 doesn't allow to define array paths

packages/rum-react/src/get-with-transaction.js
```diff
@@ -97,25 +98,7 @@ function getWithTransaction(apm) {
             return tr
           })
 
-          /**
```
vigneshshanmugam, line 100: seems like you missed to carry over these comments to the utility.

packages/rum-react/test/specs/get-apm-routes.spec.js
```diff
@@ -0,0 +1,240 @@
...
+  TRANSACTION_SERVICE
+} from '@elastic/apm-rum-core'
+import { getApmRoutes } from '../../src/get-apm-routes'
+import { getGlobalConfig } from '../../../../dev-utils/test-config'
+
+let navigate
+function Component(props) {
+  // React-router util which allows navigating between routes
+  // Cleaner than handling the history api manually
+  navigate = useNavigate()
```
vigneshshanmugam, line 56: Does  it have to be inside the component?

packages/rum-react/test/specs/get-apm-routes.spec.js
```diff
@@ -0,0 +1,240 @@
...
+      'route-change',
+      {
+        canReuse: true,
+        managed: true
+      }
+    )
+
+    transactionService.startTransaction.calls.reset()
+
+    // /**
```
vigneshshanmugam, line 137: nit: remove double comments.

packages/rum-react/src/get-apm-routes.js
```diff
@@ -0,0 +1,142 @@
...
+          // make sure we don't add an extra slash when concatenating
+          const hasSlash =
+            path[0] === '/' || builtPath[builtPath.length - 1] === '/'
+          path = hasSlash ? path : `/${path}`
+          builtPath += path
+
+          // E.g `to` property from `<Link>` component matches to window.location.pathname
+          if (branch.pathname === location.pathname) {
+            // make sure wildcard routes are also handled
+            const isWildcardRoute = builtPath.slice(-2) === '/*'
```
vigneshshanmugam, line 114: Does wildcard paths always end in this way? What about `path="/*"` ?

packages/rum-react/src/get-apm-routes.js
```diff
@@ -0,0 +1,142 @@
...
+    const navigationType = useNavigationType()
+
+    React.useLayoutEffect(
+      () => {
+        const routes = createRoutesFromChildren(props.children)
+        handleRouteChange(apm, routes, location)
+      },
+      // query param changes should not trigger this effect
+      // that's why we use `location.pathname` instead of `location`
+      [location.pathname, navigationType, props.children]
```
vigneshshanmugam, line 49: query params might not affect it, can you add a test where we add a new children based on query param?

vigneshshanmugam (commented):
Did first pass. generally looks good. Waiting for the E2E tests to finalize the review.

packages/rum-react/test/specs/get-apm-routes.spec.js
devcorpio: yes, you can't use an effect outside a component

packages/rum-react/src/get-with-transaction.js
devcorpio: oops, it seems I was hungry and I ate the comment, gonna restore it

packages/rum-react/src/get-apm-routes.js
devcorpio: > A * wildcard may be used only at the end of a path, not in the middle.

From the docs "upgrading from v5"
https://reactrouter.com/en/main/upgrading/v5#note-on-route-path-patterns

packages/rum-react/src/get-apm-routes.js
devcorpio: I added that change because that was something we had in react-router 5

The code was this:

https://github.com/elastic/apm-agent-rum-js/blob/aeb75dd282a999b684f530fd0c3955bad40e6deb/packages/rum-react/src/get-apm-route.js#L60

And this was the test:

https://github.com/elastic/apm-agent-rum-js/blob/aeb75dd282a999b684f530fd0c3955bad40e6deb/packages/rum-react/test/specs/get-apm-route.spec.js#L186


The test for react-router 6 is this one:

https://github.com/elastic/apm-agent-rum-js/blob/f63ebb8a86214d694e3915d687e0c5290a746ffa/packages/rum-react/test/specs/get-apm-routes.spec.js#L137

vigneshshanmugam: Sorry if i was not clear, The `location.pathname` is correct. I was asking about the `props.children`. There are cases when component adds say `<div>Current Query: ${path.query}>` that would trigger a transaction. If yes, lets add a test to cover this case. Because our current test uses query param but does not update the component children itself.

packages/rum-react/test/e2e/with-router/routes.js
```diff
@@ -51,10 +63,13 @@ function App() {
         </li>
       </ul>
       <Suspense fallback={<div>Loading...</div>}>
-        <Switch>
```
vigneshshanmugam, line 54: Is this file even required, given there is no switch functionality in the new router, I would be fine with deleting this part of the code.

vigneshshanmugam (approved):
LGTM 🎉

Issue #1423, opened 2023-09-06
Custom serializer for Unhandled promise rejection: <object> 


aizerin:
I want to send more details when unhandled promise rejection occurs with an object as a rejection reason. But the problem is that I'm not able to modify it somehow. This function [logPromiseEvent](https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/error-logging/error-logging.js#L174) will just throw it away.

It would be nice to have a function like https://www.elastic.co/guide/en/apm/agent/rum-js/current/agent-api.html#apm-add-filter to alter a promise payload.

I can create PR for this change if you think that it will be nice to have it.

currently, I'm using this ugly workaround:

```javascript
if (typeof window !== 'undefined') {
  const originalAddEventListener = window.addEventListener;
  window.addEventListener = (type, fn) => {
    if (type === 'unhandledrejection' && fn.toString().includes('logPromiseEvent')) {
      originalAddEventListener(type, (event) => {
        let modifiedEvent = event;
        if (isPlainObject(event.reason) && !event.reason?.message) {
          modifiedEvent = {
            reason: {
              message: 'some custom message',
              serializedData: JSON.stringify(event.reason),
            },
          };
        }
        fn(modifiedEvent);
      });
    } else {
      originalAddEventListener(type, fn);
    }
  };
}
```

devcorpio:
Hi @aizerin,

Thanks for using the RUM agent!

For further clarification, I would like to ask you a few questions:

- What are you seeing in Kibana UI if don't apply the workaround? I mean, what information are you seeing?
- What information is your workaround adding/modifying?

Additionally, could you share a snippet where you throw the kind of error you are mentioning? and showing us how things differ with and without the workaround? Something like "current behaviour", "expected/wished behaviour", for instance.

That would help us to better understand the scenario you are describing. 

Thanks,
Alberto

aizerin:
Hi,
this is the current behavior. I can see unhandled promise rejection in Kibana but I cannot find any helpful information about it.
![current](https://github.com/elastic/apm-agent-rum-js/assets/18553162/aedab0d5-8d3e-46a8-bdb9-5c038cb9f047)

with the workaround I can have more detailed informations.

I can set custom message based on data

<img width="1375" alt="image" src="https://github.com/elastic/apm-agent-rum-js/assets/18553162/c886c2bf-95ca-4f51-b390-aa913d956e9f">

And I can see serialized object in metadata of error.

[...]

PR #1428, opened 2023-09-18
feat(rum-core): stringify object rejected by a promise

devcorpio:
Solves: https://github.com/elastic/apm-agent-rum-js/issues/1423

# Context

Before this change, if the promise rejection reason was an object:

```
       const obj = {
              foo: "bar",
              rum: "Agent",
              hello: "world"
       };

       Promise.reject(obj)
```

The RUM agent was capturing the error this way:

`Unhandled promise rejection: <object>`. Additionally, none of the properties were included in the error custom properties, making it difficult for users to identify the error's context.

There was an exception to that rule. Objects containing a "message" property:

Screenshot 2023-09-18 at 17 03 29

But unfortunately, this is not something that users can always control.

# Enhancement

From now onwards, if the rejection reason of a promise is an object (with no message property), then two things will happen:

1. The RUM agent will `JSON.stringify` the obj so that the properties will be visible in the error message. This is similar to the native behaviour of browsers like Google Chrome.
2. The non-standard properties will also be added to the error custom properties

You can see an example in the screenshots below:

Screenshot 2023-09-18 at 16 48 10

Screenshot 2023-09-18 at 16 49 21

# Notes

1. the behaviour of other types remains unchanged.
    - arrays -> `<object>`
    - functions -> `<function>`

2. The "stringify" of the object will fail at least in [two circumstances](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions):
     - it contains a circular reference
     - one of the values is a BigInt

If that happens, a fallback will be applied, meaning that the error will be captured as if was being captured before this change.

### DIFF
--- a/packages/rum-core/src/error-logging/error-logging.js
+++ b/packages/rum-core/src/error-logging/error-logging.js
@@ -33,6 +33,7 @@ import stackParser from 'error-stack-parser'
  * List of keys to be ignored from getting added to custom error properties
  */
 const IGNORE_KEYS = ['stack', 'message']
+const PROMISE_REJECTION_PREFIX = 'Unhandled promise rejection: '
 
 function getErrorProperties(error) {
   /**
@@ -172,7 +173,6 @@ class ErrorLogging {
   }
 
   logPromiseEvent(promiseRejectionEvent) {
-    const prefix = 'Unhandled promise rejection: '
     let { reason } = promiseRejectionEvent
     if (reason == null) {
       reason = '<no reason specified>'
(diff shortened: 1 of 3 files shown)


packages/rum-core/src/error-logging/error-logging.js
```diff
@@ -211,6 +203,32 @@ class ErrorLogging {
...
+      } catch (error) {
+        // fallback. JSON.stringify can throw exceptions in different circumstances.
+        // please, see this link for more info:
+        // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions
+        errorEvent.message += '<object>'
+      }
+    } else if (typeof reason === 'function') {
+      errorEvent.message += '<function>'
+    } else {
+      errorEvent.message += reason
```
vigneshshanmugam, line 227: Did we handle this case before?

vigneshshanmugam (approved):
LGTM

packages/rum-core/src/error-logging/error-logging.js
devcorpio: yes, this case was:

https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/error-logging/error-logging.js#L197

PR #1434, opened 2023-09-22
chore: major dev maintenance for RUM agent

devcorpio:
# Summary

## Node and npm

- Node.js 18 support
- npm >= 8 support + `legacy-peer-deps=true` in a new .npmrc. This allows us to update things gradually. In a nutshell, that flag makes npm behave like it was doing in versions 4-6. 

## E2E Developer Experience

- updated `webdriverio` and `wdio` dependencies to the latest one. Mainly for improving the reporting and exit code behaviour, more info:
           -  https://github.com/webdriverio/webdriverio/issues/9440
           - https://github.com/webdriverio/webdriverio/issues/8307 (we are victims of this one every now and then and it's so frustrating, it's so time-consuming)
- all E2E tests are now also working with the new `webdriver` async mechanism https://webdriver.io/docs/async-migration/
- all of them keep working perfectly 🎉 , you can see that in the [CI pipeline](https://github.com/elastic/apm-agent-rum-js/actions/runs/6283001540/job/17063182739) of [this commit.](https://github.com/elastic/apm-agent-rum-js/pull/1434/commits/cfba9994850992d62b3f5a873431709cf0a0f214) The commits after that were more focused on making sure that benchmarking, load-testing, and dry-run-release also worked properly.

## Webpack compatibility with Node 18
 - Make webpack and Node.18 compatible (https://github.com/webpack/webpack/issues/14532).
        - our `webpack` config uses now the first solution(`xxhash64`) offered by the creator of `webpack`. More info [here](https://github.com/webpack/webpack/issues/14532#issuecomment-947525539)
        - for the `angular package` the second solution(`--openssl-legacy-provider` flag) is the one applied since updating` angular-cli `dependencies is not an option given the huge the task is.

## Final thoughts

This lack of maintenance was causing several DX problems and with the passing of time to fix this was becoming more and more difficult because of the enormous gap between the versions.

Possible future maintenance candidates: **lerna**, **angular cli**, **benchmarking, and load-testing** (the last two still use Node 14. Adapting Playwright and its related dependencies to work with Node.js 18 is not a piece of cake, maybe this should be the one to do first because the source from where we installed Node.js 14 is going to become inactive sooner rather than later, you can see that warning in the logs), etc. No need to rush though, it can be done gradually since this PR now solves the main blockers we had.


.ci/docker/node-playwright/Dockerfile
```diff
@@ -10,7 +10,7 @@ ARG NODEJS_VERSION
 
 # 1. Install node
 RUN apt-get -qq update && apt-get -qq install -y curl && \
-    curl -sL https://deb.nodesource.com/setup_${NODEJS_VERSION}.x | bash - && \
+    curl -sL https://deb.nodesource.com/setup_14.x | bash - && \
```
vigneshshanmugam, line 13: Given that you have set the NODE_OPTIONS on benchmark scripts, can you remove the hardcoded version here?

.ci/docker/node-puppeteer/Dockerfile
```diff
@@ -1,14 +1,33 @@
-#Dockerfile for docker.elastic.co/observability-ci/node-puppeteer:12
-# When changing Node version, please build the Docker image for RUM here:
-#   https://apm-ci.elastic.co/job/apm-shared/job/apm-docker-images-pipeline/build?delay=0sec
-# Remember checking the 'rum' parameter
 ARG NODEJS_VERSION
-FROM node:${NODEJS_VERSION}
+FROM ubuntu:jammy
```
vigneshshanmugam, line 2: Did you talk to the Robots team about using this custom image? 

The other option is to fix our obserability ci image which is from our shared pipeline.

vigneshshanmugam (approved):
LGTM 🎉

.ci/docker/node-playwright/Dockerfile
devcorpio: Actually, I will let the hardcoded version here and remove the NODE_OPTIONS from benchmark scripts until we decide to apply this update to the benchmarking, too. My surprise was big when I saw that this dockerfile is also built in a different step (which provides the Node 18 version) of the CI: https://github.com/elastic/apm-agent-rum-js/actions/runs/6284894505/job/17066717885

screenshot

And btw, just for more visibility, this is the capture of the Node.js 14 warning I mentioned in the PR's description:

Screenshot 2023-09-26 at 10 40 39

.ci/docker/node-puppeteer/Dockerfile
devcorpio: I didn't mention anything explicitly because, in the other image, we were using `ubuntu:bionic` which is another custom "image".

I can share this with Robots, and discuss if it would be required to align approaches in future iterations

amannocci (approved):
LGTM

.ci/docker/node-puppeteer/Dockerfile
amannocci: It's fine to use any distro base image from a trusted provider.

PR #1437, opened 2023-09-26
feat(ci): honour dry-run for CDN publishing

v1v:
This should help with running on dry-run mode


amannocci:
I don't understand why it doesn't honor dry-run yet.
AFAIR, a different bucket is used to test the workflow in dry-run.

v1v:
> I don't understand why it doesn't honor dry-run yet.

I didn't see the logic `Setup credentials` but assumed CDN should not run when using `dry-run`.

I still think it's misleading to use `dry-run` and still produce artifacts. I'll close this now

PR #1438, opened 2023-09-26
feat(ci): run CDN if dist files

v1v:
Upload artifacts only if artifacts exists. This should help with the corner case when one of the components in the mono-repo does not produce bundles.

 as explained in https://github.com/orgs/community/discussions/25790 and https://github.com/google-github-actions/upload-cloud-storage/issues/332


devcorpio (approved):
LGTM, thanks for doing this


Like I said in the other PR:

 just in case modify the feat(ci) to chore(ci) before merging, I'm not sure if feat(ci) will affect the semantic versioning of our next release and we don't want these changes to affect that.

amannocci (approved):
LGTM

PR #1452, opened 2023-10-30
feat(rum-core): add fallback to click transaction name

kareemaly:
I came across a scenario in [findCustomTransactionName](https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/common/observers/page-clicks.js#L110), where it returns `null` if it doesn't find a closest element matching the `INTERACTIVE_SELECTOR`. In our frontend, we utilize clickable elements like images that are not wrapped in `button` or `a` tags, but they do have the `data-transaction-name` attribute defined. However, the current implementation still leads to a `Click - img` transaction name.

I noticed the current behavior is covered in unit tests which suggests it might be by design? If not, then I'm proposing a slight modification to the `findCustomTransactionName` function to return `target.dataset.transactionName` as a fallback when no closest matching element is found.

Looking forward to your feedback. Thank you!

### DIFF
--- a/packages/rum-core/src/common/observers/page-clicks.js
+++ b/packages/rum-core/src/common/observers/page-clicks.js
@@ -112,7 +112,9 @@ function findCustomTransactionName(target) {
     // Leverage closest API to traverse the element and its parents
     // only links and buttons are considered.
     const element = target.closest(INTERACTIVE_SELECTOR)
-    return element ? element.dataset.transactionName : null
+    return element
+      ? element.dataset.transactionName
+      : target.dataset.transactionName
   }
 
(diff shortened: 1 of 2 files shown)


devcorpio:
Hi @kareemaly,

Thanks a lot for your contribution!

Currently, we have limited availability to review your PR - due to project priorities within the team - we'll let you know once we are ready to review it.

Thanks for your time,

Alberto

devcorpio:
Hi @kareemaly,

just letting you know that we are fixing this in this PR: https://github.com/elastic/apm-agent-rum-js/pull/1511

Thanks again for your initiative

Alberto

Issue #1293, opened 2022-10-20
Add support for Interaction Next Paint (INP)


devcorpio:
# Summary

[INP](https://web.dev/inp/) is a metric whose goal is to measure responsiveness throughout the page lifecycle.

> Chrome usage data shows that 90% of a user's time on a page is spent after it loads, Thus, careful measurement of responsiveness throughout the page lifecycle is important. This is what the INP metric assesses.

To strategy followed by Google to measure this on the field is to rely on the `visibilitystate` event. The logic that decides if the metric should be reported is performed every time the page's visibility state changes to hidden. Due to that fact, it's important to keep in mind that the metric can be reported multiple times on the same page.

The nature of this new metric doesn't **seem** to fit perfectly with any of the existing transactions on the agent. Since the **main** goal of this metric is to measure the responsiveness after the page's load we cannot attach this information to the **page-load** transaction.

**Possible** approaches to tackle this:

- Create a new type of transaction for this
- Maybe it's time to revisit the [custom metrics feature](https://github.com/elastic/apm-agent-rum-js/issues/197)
- [... more ideas ...]

Please, feel free to add comments, ideas, etc.

CC: @kyungeunni @vigneshshanmugam

paulb-elastic:
INP to replace FID as a core web vital metric in March 2024 https://web.dev/inp-cwv/

paulb-elastic:
Some useful links:
- https://developer.chrome.com/blog/inp-tools-2022/
- https://web.dev/inp/#measure-inp-in-javascript
- https://web.dev/inp/#in-the-lab

StuGray86:
Do you have any plans for dealing with this? Google are rolling this out in March, so it would be good to see some plans around how Elastic RUM will deal with this.

RobDTech:
We are an Enterprise Retail customer, this metric is key to all Retailers and the lack of visibility on this metric is worrying. Please can you indicate if this is something that is coming in the next release?

devcorpio:
Hi @StuGray86 @RobDTech,

A new version of the agent was released, please see the [release notes](https://www.elastic.co/guide/en/apm/agent/rum-js/current/release-notes-5.x.html#release-notes-5.16.0) for more details.

Thanks,
Alberto

PR #1462, opened 2023-12-12
feat(rum): report INP metric

devcorpio:
# Summary

Related to https://github.com/elastic/apm-agent-rum-js/issues/1293

From now on, the RUM agent will start reporting the INP metric.

- INP will not be reported if there is no user activity
- INP will be reported as 0 if the slowest detected interaction duration is < 40
- INP will pick the slowest interaction if there have been less than 50 interactions
- INP will result from calculating the P98 if there have been more than 50 interactions.
- The metric will be reported when the user leaves the page, which can happen multiple times.
- We will "reset" the tracked interactions every time we report the metric
- INP will be attached to the page where the user "hard-navigated to".

For further details, please see the code which is heavily documented given the experimental nature of this metric.

**Screenshot where we can see the INP metric on the User Experience page:**

INP metric


devcorpio:
hi @v1v,

it seems our test steps are [failing](https://github.com/elastic/apm-agent-rum-js/actions/runs/7194615188/job/19595512188?pr=1462) with this error:

Screenshot 2023-12-13 at 13 05 37

Looking for info I found that might be related with Docker 7.0.0, [link](https://docker-py.readthedocs.io/en/stable/change-log.html#id1) to the changelog

and a link to a conversation that a few people are having: https://github.com/docker/docker-py/issues/3194

Do you know what can we do to fix this?

v1v:
> Do you know what can we do to fix this?

As far as I see `python3.10` is the one set in the GitHub runners. I don't know whether we can use a pinned version for the GitHub runner or do some magic with downgrading the docker version.

Let me see if I can figure out how to solve this

v1v:
https://github.com/elastic/apm-agent-rum-js/pull/1462/commits/871d1a088aaf3ec64cda1339dcfa3959005317d0 might help, I followed the steps explained [here](https://github.com/ahembree/ansible-hms-docker/issues/48#issuecomment-1851258396).

packages/rum-core/src/performance-monitoring/transaction.js
```diff
@@ -196,6 +198,15 @@ class Transaction extends SpanBase {
   isManaged() {
     return !!this.options.managed
   }
+
+  /*
+    Makes sure that a transaction ended while the page was becoming hidden
+    it's not discarded for that very same reason.
+  */
+  endPageHidden(endTime) {
```
devcorpio, line 206: I created this function because while I was doing the last exploratory testing session I spotted a bug:

The logic where we detect if the website becomes hidden to end the ongoing transaction had a race condition:

1. transaction.end()
2. update lastHiddenStart

unfortunately, the [code where we check](https://github.com/elastic/apm-agent-rum-js/blob/7cd5f5ecbe02ee4ddc3e648927fa4ac0ce3dbb28/packages/rum-core/src/performance-monitoring/transaction-service.js#L258) if a transaction should be discarded because "the page was hidden during the transaction" doesn't happen in a sync way in transaction.end(), it happens after executing a javascript Promise. 

So, assuming that calling `transaction.end()` in that part of the code was a synchronous process was wrong. It was updating the lastHiddenStart before doing the end transaction check (even if the end was invoked before)

**Fortunately**, this was not affecting the page-load transaction, I only reproduced the described behaviour with the INP transaction and also a few times (not all the time, I guess that depended on who was "winning" the race) with clicks transactions.

dev-utils/test-config.js
```diff
@@ -172,7 +172,8 @@ function getBrowserList(pkg = 'default') {
       },
       {
         browserName: 'firefox',
-        browserVersion: 'latest',
+        // beware that if we update to 99 or more we will need to cope with https://github.com/karma-runner/karma-sauce-launcher/issues/275
+        browserVersion: '98',
```
devcorpio, line 176: for now, avoid investing time in this annoying issue where saucelabs, webdriverio, and, karma-sauce-launcher integration are causing an error.

screenshot

Related info:

https://github.com/karma-runner/karma-sauce-launcher/issues/275
https://github.com/elastic/apm-agent-rum-js/pull/1303

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * Updates the total count of different user interactions occurred in a page
+ * Once `performance.interactionCount` becomes globally available the "helper/polyfill logic" function
+ * will not be needed anymore.
+ *
+ * the reasons why interactionCount is important are:
+ * 1. To get the proper percentile 98 when calculating the INP value.
+ * 2. To report INP as 0 (rather than undefined) in those circumstances where the PerformanceObserver doesn't report entries.
+ *    That can happen because the min value for durationThreshold is 16 then interactions faster than that will not be reported.
```
vigneshshanmugam, line 210: its a bit confusing, i think we can word it as `interactions faster than 16ms will not be taken in to account`

packages/rum-core/test/performance-monitoring/metrics/inp/report.spec.js
```diff
@@ -0,0 +1,96 @@
...
+      const tr = reportInp(transactionService)
+      expect(tr).toBeUndefined()
+    })
+
+    describe('when metric is >= 0', () => {
+      beforeEach(() => {
+        calculateInpSpy.and.returnValue(100)
+      })
+
+      it('should create page-exit transaction', () => {
```
vigneshshanmugam, line 59: would just merge all these tests together instead of separate it statements.

vigneshshanmugam (commented):
Great job @devcorpio 🎉, Thanks a ton for adding in lots of details around the core INP calculation logic and also adding relevant links to the spec where necessary. It was really nice to see them side by side and also understanding the spec/browser reporting.

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
```diff
@@ -0,0 +1,256 @@
...
+  inpState.minInteractionId = Infinity
+  inpState.maxInteractionId = 0
+  inpState.interactionCount = 0
+  inpState.longestInteractions = []
+}
+
+// Stores user interaction
+function storeUserInteraction(entry) {
+  // Ignore interactions that are faster (or equal) than the least slow known interaction
+  const leastSlow =
```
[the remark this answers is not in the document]
devcorpio: this is just an "optimization" (maybe a bit premature since the data structure cannot grow so much...). I would be open to delete this code if needed.

What it does:

if the newly captured entry is quicker than the least slow one then don't store it. The slowest one is the position 0 of the array, the least slowest one is in the last position (because of the sorting)

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /**
+   * different entries can share the same interactionId
+   * Please see the EventTiming spec for more details about
+   * the algorithm: https://www.w3.org/TR/event-timing/#sec-computing-interactionid
+   */
+  let foundIndex = -1
+  let i = 0
+  while (i < inpState.longestInteractions.length && foundIndex === -1) {
```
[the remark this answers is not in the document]
devcorpio: The data structure stores a few entries... so I agree, I also prefer to prioritize readability over "performance". will change

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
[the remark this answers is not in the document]
devcorpio: thought again about this, I'm going to remove it, we are talking just about a few entries (10~). Let's write optimizations only when they are needed. "Nuke" mode activated 😄

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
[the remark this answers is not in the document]
devcorpio: oh I see, just remember why I used that:

`Array.prototype.some` returns a boolean and we want the index
Array.prototype.find (same with findIndex) not IE11 compatible (I know this code will not run in IE11, but even in this case...)

I would advocate to keep the while then

packages/rum-core/src/performance-monitoring/metrics/inp/report.js
```diff
@@ -0,0 +1,68 @@
...
+    })
+
+    // make sure that transaction's duration is > 0
+    // we do the +1 because INP can be also reported as 0
+    const endTime = startTime + inp + 1
+    inpTr.endPageHidden(endTime)
+
+    // restart INP tracking information
+    // since users can enter and leave the page multiple times
+    restoreINPState()
```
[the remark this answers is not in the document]
devcorpio: `reportINP` executes when a page becomes hidden, so the answer to your question would be: "Every time we hide the page we should restore to the initial state." Besides, if `restoreINPState()` gets executed is because there is an INP to value report ( >= 0)

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
devcorpio: I agree, besides there was a typo "it should slower than" :D

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
```diff
@@ -0,0 +1,256 @@
...
+  const inp = inpState.longestInteractions[interactionIndex].duration
+
+  return inp
+}
+
+export function interactionCount() {
+  return performance.interactionCount || inpState.interactionCount
+}
+
+export function inpCandidates() {
```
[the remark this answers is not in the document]
devcorpio: I will need to export the state like we do in metrics

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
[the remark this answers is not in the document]
devcorpio: I added two additional tests to make sure this knowledge if represented in code 🚀

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
```diff
@@ -0,0 +1,256 @@
...
+  recorder = new PerfEntryRecorder(processUserInteractions)
+) {
+  /**
+   * We set the lowest threshold (16) supported by browsers
+   * this way we can estimate the total number of different user interactions which is essential to calculate P98, among other things.
+   * Once browsers start exposing the `performance.interactionCount` property the
+   * threshold will be 40, which is the value Google recommends as the default threshold
+   * https://github.com/GoogleChrome/web-vitals/blob/3806160ffbc93c3c4abf210a167b81228172b31c/src/onINP.ts#L203
+   */
+  const durationThreshold = isPerfInteractionCountSupported()
```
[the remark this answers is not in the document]
devcorpio: I declared a variable on top of this function. (so we don't call it twice there)

But I would keep this as a util since makes testing more flexible. (at the end of the day we are only invoking it twice anyway)

packages/rum-core/src/performance-monitoring/metrics/inp/process.js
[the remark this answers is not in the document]
vigneshshanmugam: I was going for the wording, The logic itself feels good to me. But thanks for adding the extra tests

packages/rum-core/src/performance-monitoring/transaction.js
vigneshshanmugam: We discussed this offline and planned to add another event `transaction:skipped` to update the last hidden start time.

devcorpio: this is the related commit https://github.com/elastic/apm-agent-rum-js/pull/1462/commits/df7423dd9e46545f683d6a6e6db6f5358282d018

packages/rum-core/src/common/observers/page-visibility.js
```diff
@@ -76,9 +76,26 @@ export function observePageVisibility(configService, transactionService) {
  * @param transactionService
  */
 function onPageHidden(configService, transactionService) {
-  reportInp(transactionService)
+  const inpTr = reportInp(transactionService)
+  // we don't want to flush the queue for every transaction that is ended when page becomes hidden
+  // and given the async nature of the ending process of transactions
```
vigneshshanmugam, line 85: nit: might reword it as transaction ends are scheduled async as microtasks(promise), so we are coordinating the flushing process

packages/rum-core/src/common/constants.js
```diff
@@ -110,6 +110,7 @@ const TRANSACTION_END = 'transaction:end'
 const CONFIG_CHANGE = 'config:change'
 const QUEUE_FLUSH = 'queue:flush'
 const QUEUE_ADD_TRANSACTION = 'queue:add_transaction'
+const TRANSACTION_DISCARD = 'transaction:discard'
```
vigneshshanmugam, line 113: should we rename it as trasaction:ignore/skip ? mainly suggesting because ignored also aligns with our config settings.

vigneshshanmugam (approved):
LGTM, Added small suggestions. Nothing blocking.

packages/rum-core/src/common/constants.js
devcorpio: yes, it makes sense 🚀

shahzad31 (approved):

PR #1511, opened 2024-06-27
fix(rum): make sure custom transaction name logic is IE11 compliant

### LINKED ISSUE #1510: user-interactions transactions don't work in IE11
# Summary

It has been reported that user-interaction transactions (aka click transactions) are not working in IE11. To be more explicit, every time a user clicks the RUM agent reports:

```
TypeError: Cannot get property 'transactionName', which is undefined or a null reference.
```

The reason for the issue is that to extract the `data-custom-transaction-name` value we use [dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) which is not IE11 compliant.

devcorpio:
Resolves #1510 

### Fix:

We stop using [dataset](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) (even in browsers where it works) and we use [getAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute)

Since we are modifying this area, we also take the opportunity to fix this other [issue](https://github.com/elastic/apm-agent-rum-js/pull/1452) that was reported a while ago

### DIFF
--- a/packages/rum-core/src/common/observers/page-clicks.js
+++ b/packages/rum-core/src/common/observers/page-clicks.js
@@ -108,13 +108,15 @@ function buildTransactionName(target) {
 }
 
 function findCustomTransactionName(target) {
+  const trCustomNameAttribute = 'data-transaction-name'
+  const fallbackName = target.getAttribute(trCustomNameAttribute)
   if (target.closest) {
     // Leverage closest API to traverse the element and its parents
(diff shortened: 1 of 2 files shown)


vigneshshanmugam (approved):
LGTM. Good find 🎉

PR #1525, opened 2024-08-23
feat: Add callback option to config to capture context when starting transactions and spans

cjr125:


vigneshshanmugam (commented):
Not sure whats the use-case we are solving with this. Why not use Global Labels for this functionality - https://www.elastic.co/guide/en/apm/agent/rum-js/current/agent-api.html#apm-add-labels

cjr125:
> Not sure whats the use-case we are solving with this. Why not use Global Labels for this functionality - https://www.elastic.co/guide/en/apm/agent/rum-js/current/agent-api.html#apm-add-labels

The specific use case for the callback is to capture context specific data which is different between the call stacks when firing / handling the transaction / span events from the auto-instrumentation. For example, when an application uses a combination of javascript files in which some import the RUM agent and others do not, there will be different stack traces when comparing the stack frame where the event was fired as opposed to handled. The ultimate goal would still be to assign labels, the gap being the context when the event is fired in cases where that context does not import the RUM agent library

cjr125:
@vigneshshanmugam any update on this? do you understand the use case?

malek-andrew:
@vigneshshanmugam Customer filed a support case today regarding this PR. Would you mind give it another review?

david-luna:
Hi @cjr125 

While I run some tests locally to understand better the intent of this new config I'm going to do a 1st review on this. Thanks for the contribution :)

david-luna:
Hi @cjr125

I did finish the testing on this change-set. Thanks for being patient and for your contribution. Here are some questions and comments regarding this.

### use case
> The specific use case for the callback is to capture context specific data which is different between the call stacks when firing / handling the transaction / span events from the auto-instrumentation. For example, when an application uses a combination of javascript files in which some import the RUM agent and others do not, there will be different stack traces when comparing the stack frame where the event was fired as opposed to handled. The ultimate goal would still be to assign labels, the gap being the context when the event is fired in cases where that context does not import the RUM agent library

Is the call stack the only difference? 

### 2 options instead of 1

The same option is used to extract context whenever a new transaction or span is created by the RUM agent. One potential problem I see is that there is no way to differentiate is the function is being called for a transaction or for a span. Also I think to leave the user to deal with this detection may be cumbersome. I propose to have a dedicated option per type (transaction or span) so we have more control on what info we add into transactions and spans and.

### be defensive about user input

In your implementation the callback function is called if the configuration option is defined but it does not check that the value passed is a callable function. A bad configuration option will raise an error when the agent tries to call an expression which is not a function and it may break the user's app (it certainly breaks the instrumentation).

Also the RUM agent needs to make sure the function call does not raise an error (try/catch) and check if the returned value from the callback is of the type we expect to add it to the labels/tags

### make use of the API

similar to the above section setting the return object as tags property may lead to issues when processing the data and displaying it in kibana. The public API `addLabels` from both transaction and span performs some checks and makes sure the tag names have valid chars in it ([ref](https://github.com/elastic/apm-agent-rum-js/blob/65ebcabe9cd1472d6b0b8da7b0a5489fe7d126f3/packages/rum-core/src/performance-monitoring/span-base.js#L67-L75))

cjr125:
> Hi @cjr125
> 
> I did finish the testing on this change-set. Thanks for being patient and for your contribution. Here are some questions and comments regarding this.
> ### use case
> 
> > The specific use case for the callback is to capture context specific data which is different between the call stacks when firing / handling the transaction / span events from the auto-instrumentation. For example, when an application uses a combination of javascript files in which some import the RUM agent and others do not, there will be different stack traces when comparing the stack frame where the event was fired as opposed to handled. The ultimate goal would still be to assign labels, the gap being the context when the event is fired in cases where that context does not import the RUM agent library
> 
> Is the call stack the only difference?
> ### 2 options instead of 1
> 
> The same option is used to extract context whenever a new transaction or span is created by the RUM agent. One potential problem I see is that there is no way to differentiate is the function is being called for a transaction or for a span. Also I think to leave the user to deal with this detection may be cumbersome. I propose to have a dedicated option per type (transaction or span) so we have more control on what info we add into transactions and spans and.
> ### be defensive about user input
> 
> In your implementation the callback function is called if the configuration option is defined but it does not check that the value passed is a callable function. A bad configuration option will raise an error when the agent tries to call an expression which is not a function and it may break the user's app (it certainly breaks the instrumentation).
> 
> Also the RUM agent needs to make sure the function call does not raise an error (try/catch) and check if the returned value from the callback is of the type we expect to add it to the labels/tags
> ### make use of the API
> 
> similar to the above section setting the return object as tags property may lead to issues when processing the data and displaying it in kibana. The public API `addLabels` from both transaction and span performs some checks and makes sure the tag names have valid chars in it ([ref](https://github.com/elastic/apm-agent-rum-js/blob/65ebcabe9cd1472d6b0b8da7b0a5489fe7d126f3/packages/rum-core/src/performance-monitoring/span-base.js#L67-L75))
> 
> I think these 3 points should be addressed. I'm expecting to have bandwidth in the following days so if you prefer I can jump in the implementation :)

@david-luna I believe I have addressed each of your comments. Can you please review the changes again when you have a chance?

packages/rum-core/src/performance-monitoring/span.js
```diff
@@ -39,6 +39,19 @@ class Span extends SpanBase {
...
+    if (
+      this.options.spanContextCallback &&
+      typeof this.options.spanContextCallback === 'function'
+    ) {
+      let tags
+      try {
+        tags = this.options.spanContextCallback()
+        this.addLabels(tags)
+      } catch (e) {
+        console.error('Failed to execute span context callback', e)
```
david-luna, line 52: although this message may be really helpful the RUM agent should not use the console API directly. Instead we have a logger service that is aware of the log level configured in the agent.

Injecting that service here is not a good idea IMHO. However the transaction service already has access to the logger so it seems a good place to wrap the function and log the errors.

packages/rum-core/src/performance-monitoring/transaction.js
```diff
@@ -54,6 +54,19 @@ class Transaction extends SpanBase {
 
     this.sampleRate = this.options.transactionSampleRate
     this.sampled = Math.random() <= this.sampleRate
+
+    if (
```
david-luna, line 58: if we delegate to the transaction service the wrapping if the function (as suggested above) this will we simplified to only check for existence & call it without worrying for errors

```js
if (this.options.transactionContextCallback) {
  this.setLabels(this.options.transactionContextCallback())
}
```

packages/rum-core/src/performance-monitoring/transaction.js
```diff
@@ -96,7 +109,17 @@ class Transaction extends SpanBase {
     if (this.ended) {
       return
     }
-    const opts = extend({}, options)
+    let opts = extend({}, options)
+
+    if (
```
david-luna, line 114: if we delegate to the transaction service the wrapping if the function (as suggested above) this will we simplified to only check for existence & call it without worrying for errors

```js
if (this.options.spanContextCallback) {
  this.setLabels(this.options.spanContextCallback())
}
```

packages/rum-core/test/performance-monitoring/transaction-service.spec.js
```diff
@@ -651,6 +651,61 @@ describe('TransactionService', function () {
...
+      const transactionService = new TransactionService(logger, config)
+
+      const sp1 = transactionService.startSpan('span1', 'span')
+
+      sp1.onEnd = () => {
+        expect(sp1.context.tags.stack).toBeTruthy()
+        done()
+      }
+      sp1.end()
+    })
```
david-luna, line 708: maybe testing that if passing a function that does throw an error the agent does not break (+1 if we can check the log is done)

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -101,7 +101,36 @@ class TransactionService {
...
+          tags = config.transactionContextCallback()
+        } catch (err) {
+          this._logger.error(
+            'Failed to execute transaction context callback',
+            err
+          )
+        }
+        return tags
+      }
+      presetOptions.transactionContextCallback = tCC.bind(this)
```
david-luna, line 120: I think is a good idea to bond a context to the callback function but in this case I would prefer to protect the RUM agent from external code by binding to a `null` context. This way the user defined callback has no access possibility to manipulate the transaction service via the `this` keyword.
```suggestion
      presetOptions.transactionContextCallback = tCC.bind(null)
```

packages/rum-core/src/performance-monitoring/transaction-service.js
```diff
@@ -101,7 +101,36 @@ class TransactionService {
...
+      const sCC = function () {
+        let tags = {}
+        try {
+          tags = config.spanContextCallback()
+        } catch (err) {
+          this._logger.error('Failed to execute span context callback', err)
+        }
+        return tags
+      }
+      presetOptions.spanContextCallback = sCC.bind(this)
```
david-luna, line 132: same here 
```suggestion
      presetOptions.spanContextCallback = sCC.bind(null)
```

packages/rum/src/index.d.ts
```diff
@@ -108,6 +108,8 @@ declare module '@elastic/apm-rum' {
       payload?: string
       headers?: Record<string, string>
     }) => boolean
+    transactionContextCallback?: (...args: any[]) => any
+    spanContextCallback?: (...args: any[]) => any
```
david-luna, line 112: let's be more explicit about the types here:
- we know for sure the RUM agent will call them with no args
- we have an explicit type for labels

```suggestion
    transactionContextCallback?: () => Labels
    spanContextCallback?: () => Labels
```

packages/rum-core/src/performance-monitoring/transaction-service.js
cjr125: Binding to a null context seems to protect access against using the logger. I get the following error from transaction-service.spec.js if I try this change:
TypeError: Cannot read properties of null (reading '_logger')
Are there other changes needed to still access the logger? Or do we want to leave the binding set to this?

cjr125: I could also try binding to a new object which just defines the logger if that will be acceptable: e.g.
```
presetOptions.transactionContextCallback = tCC.bind({
  _logger: this._logger
})
```

david-luna: sorry, I missed you're using `this` to access the logger. In order to be able to `.bind(null)` you can refer to the logger within de closure like is done with `config`. 

```js
const config = this._config.config
const logger = this._logger // get a reference to the logger

...
if (...) {
  const tCC = function () {
    try {
      ...
    } catch (err) {
      logger.error('...', err) // then you can use it without `this.`
    }
  }
}

```

cjr125: I think we actually do not need to use the this keyword at all within the wrapper function

david-luna:
run docs-build

david-luna (approved):
LGTM. Thanks for your contribution :)

david-luna:
@vigneshshanmugam do you mind to do a quick check. Thanks

docs/configuration.asciidoc
```diff
@@ -419,3 +419,65 @@ This is useful on scenarios where the APM server is behind a reverse proxy that
 
 NOTE: If APM Server is deployed in an origin different than the page’s origin, you will need to
 <<configuring-cors, configure Cross-Origin Resource Sharing (CORS)>>.
+
+
+[function]
```
david-luna, line 424:
```suggestion
[float]
```

docs/configuration.asciidoc
```diff
@@ -419,3 +419,65 @@ This is useful on scenarios where the APM server is behind a reverse proxy that
...
+      stack = (error as Error).stack || ''
+    }
+    stack = stack.split('\n').map(function (line) { return line.trim(); })
+    return { stack };
+  }
+}
+----
+
+
+[function]
```
david-luna, line 455:
```suggestion
[float]
```

david-luna (approved):
a couple of formatting changes to make the build pass

docs/configuration.asciidoc
```diff
@@ -419,3 +419,65 @@ This is useful on scenarios where the APM server is behind a reverse proxy that
...
+`transactionContextCallback` allows the agent to specify a function to be called when starting automatically instrumented transactions and return context to 
+be set as tags. This enables the agent to capture data such as call stack frames and variable values from the scope when instrumented events are fired from 
+files  which do not import the RUM agent library.
+
+The following example illustrates an example which captures the stack trace:
+
+[source,js]
+----
+var options = {
+  transactionContextCallback: () => {
```
vigneshshanmugam, line 440: I am not a big fan of this approach and exposing configurations feels unncessary. Why not adding custom properites to transaction context via

### apm.SetCustomContext

You can basically call this [function apm.SetCustomContext](https://www.elastic.co/guide/en/apm/agent/rum-js/current/agent-api.html#apm-set-custom-context) anywhere and get the context added to the transaction/error events 

### apm.Observe

You can rely on the events when any transaction gets started/ended and add the relevant tags/context specific information. 

```ts
apm.observe('transaction:start', function (transaction) {
    stack = "" // error stack
    transaction.addLabels({ stack })
})
```

Let me know what you think.

cjr125: @vigneshshanmugam the callback approach makes an important difference when context needs to be captured from an auto-instrumented transaction which starts in a file which is loaded by the application but does not import the RUM agent library. Using the built-in setCustomContext approach will not show the URL of the file where such a transaction started in a predictable order in the stack trace if we are to capture it as you have described. The callback approach will guarantee it will occur at the top of the stack trace. The use case we are interested in is quickly identifying the developer responsible for maintaining the code where a problematic transaction starts, so the callback approach is the only way to guarantee that we can immediately filter out the other files (when the original file does not import the library). Please let me know if you have other questions.

cjr125: @vigneshshanmugam is the explanation above clear? are there any other questions? It is not possible to solve the problem described above using the existing observe / setCustomContext methods

PR #1582, opened 2025-03-13
cleanup: remove unused apm-server Dockerfile

kruskall:
this is causing quite a few alerts in snyk

the file is unused and broken since it's using an old version of go


vigneshshanmugam (approved):
Thank you @kruskall , mainly used for dev, we can remove this.

Issue #1608, opened 2025-05-13
Missing MAP files


darrensherman:
I noticed the `.map` files are missing from the Releases here, but are listed in [UNPKG](https://app.unpkg.com/@elastic/apm-rum@5.17.0/files/dist/bundles). 

Was that intentional?

PR #1619, opened 2025-06-16
chore: add missing assets in github release

david-luna:
The github release for `rum-core` does not add the `*.map` files. 

Fixes: #1608

### DIFF
--- a/scripts/github-release.js
+++ b/scripts/github-release.js
@@ -75,14 +75,8 @@ function createRelease(token) {
 }
 
 function uploadAssets(uploadUrl, token) {
-  /**
-   * Exclude the files that are not required for releasing
-   *
-   * TODO: Remove this filtering logic once we revisit bundling steps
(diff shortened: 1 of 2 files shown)


scripts/github-release.js
```diff
@@ -75,14 +75,8 @@ function createRelease(token) {
 }
 
 function uploadAssets(uploadUrl, token) {
-  /**
-   * Exclude the files that are not required for releasing
-   *
-   * TODO: Remove this filtering logic once we revisit bundling steps
```
david-luna, line 81: note for reviewer: it seems we forgot this TODO and that's why we do not have `*.map` files

vigneshshanmugam (approved):
LGTM

scripts/github-release.js
david-luna: no, just `.js` and `.map` files

PR #1665, opened 2025-12-02
feat(rum-core): allow sending transactions with no spans using config

david-luna:
The route change transaction is a type of transaction that can be interesting eve if it does not contain spans. Can be used to track the journey of the users across the instrumented web application. Is specifically useful for single page applications (SPA) where the URL can change without performing any HTTP request (hence no spans) because of the app caching data or because there is no need to fetch data.

~The change introduces a new flag in configuration named `sendAllRouteChanges` which defaults to `false` keeping the current behavior. Apps that wants to record all transactions of type `route-change` can set this config to `true`.~

UPDATE: changed the config options to allow all transactions without spans.

The change introduces a new flag in configuration named `reportTransactionsWithoutSpans` which defaults to `false` keeping the current behavior. Apps that wants to record all transactions regardless if they have spans or not can set this config to `true`.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -368,7 +368,7 @@ If APM Server is deployed in an origin different than the page’s origin, you w
 ::::
 
 
-### `recordTransactionsWithoutSpans` [record-transactions-without-spans]
+### `reportTransactionsWithoutSpans` [report-transactions-without-spans]
 
 * **Type:** Boolean
 * **Default:** `false`
--- a/packages/rum-core/src/common/config-service.js
+++ b/packages/rum-core/src/common/config-service.js
@@ -96,7 +96,7 @@ class Config {
       session: false,
       apmRequest: null,
       sendCredentials: false,
-      recordTransactionsWithoutSpans: false
+      reportTransactionsWithoutSpans: false
     }
 
     this.events = new EventHandler()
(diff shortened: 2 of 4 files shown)


docs/reference/configuration.md
```diff
@@ -368,3 +368,9 @@ If APM Server is deployed in an origin different than the page’s origin, you w
 ::::
 
 
+### `sendAllRouteChanges` [send-all-route-changes]
```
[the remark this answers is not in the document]
david-luna: My suggestion is `recordEmptyTransactions`. Empty meaning with no spans. WDYT?

docs/reference/configuration.md
```diff
@@ -368,9 +368,9 @@ If APM Server is deployed in an origin different than the page’s origin, you w
 ::::
 
 
-### `sendAllRouteChanges` [send-all-route-changes]
+### `recordEmptyTransactions` [record-empty-transactions]
```
vigneshshanmugam: nit: reportTransactionsWithoutSpans? - fine with Empty as well. Just provided as a suggestion.

david-luna: commit https://github.com/elastic/apm-agent-rum-js/pull/1665/commits/a4c2fca4767d45261c4552b2a5e8002fb6adba93

Issue #1672, opened 2025-12-08
captureError sending labels not working


TomaszSz368:
Description:
Adding labels to error through `captureError(Error, { labels: ...})` doesn't work because they are ignored when are send to server because only labels/tags from configService in [createMetadata](https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/common/apm-server.js#L325) function are taken into consideration.

Expected result:
`options.label` labels added in `captureError` function call are send to server

Related Issue: #1591

PR #1673, opened 2025-12-08
Add tags to compressed context object

Mighty683:
Solves issue: https://github.com/elastic/apm-agent-rum-js/issues/1672

Adding tags from error context to payload send to intake api


david-luna:
@Mighty683 thanks for your contribution. I think this PR misses a couple of things:
- CLA signed
- a test to ensure data is sent properly (I can help with that)

PR #1682, opened 2026-01-14
fix: create Request object with init

DevooKim:
Updates the fetch patching logic to better handle cases where a `Request` object is used as the input along with an options object.

Changes
* Updated the `patchFetch` function in `fetch-patch.js` to create a new `Request` object when both a `Request` input and an `init` options object are provided, ensuring fetch is called with the correct parameters.

### DIFF
--- a/packages/rum-core/test/common/fetch-patch.spec.js
+++ b/packages/rum-core/test/common/fetch-patch.spec.js
@@ -136,10 +136,17 @@ describe('fetchPatch', function () {
         expect(typeof promise.then).toBe('function')
         expect(events.map(e => e.event)).toEqual(['schedule'])
 
-        // Validate that fetch has been called with the proper data
-        const data = events[0].task.data
-        validateFetchData(data, newFetchArgs)
-        done()
+        promise.then(function (resp) {
+          expect(resp).toBeDefined()
+          Promise.resolve().then(function () {
+            // Validate that fetch has been called with the proper data
+            const data = events[0].task.data
+            validateFetchData(data, newFetchArgs)
+
+            expect(events.map(e => e.event)).toEqual(['schedule', 'invoke'])
+            done()
+          })
+        })
       })
 
       it('should produce task events when fetch fails', function (done) {


david-luna:
@DevooKim thanks for your contribution to the RUM agent. Would you mind to sing the CLA while I'm reviewing your changes?

DevooKim:
> @DevooKim thanks for your contribution to the RUM agent. Would you mind to sing the CLA while I'm reviewing your changes?

@david-luna I already signed it.

packages/rum-core/src/common/patching/fetch-patch.js
```diff
@@ -87,7 +87,7 @@ export function patchFetch(callback) {
         url = input
       }
     } else if (input) {
-      request = input
+      request = init ? new Request(input, init) : input
```
david-luna: I think not only the change but the surrounding if/else block are missing a scenario. Checking the [fetch docs in MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#parameters) I see the 1st param could be:
- a string
- a URL object
- an object with a stringifier <-- this case is not covered in the if/else blocks
- a Request <-- this PR tries to cover a variation this case

The Request constructor accepts the following params according to [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#parameters):
- a string
- a Request object
NOTE: testing a bit in developer tools I've verified the constructor also accepts an object with a stringifier

So we may end up in this `else if` block with a `input` param that is not a `Request` object but rather an different object that returns a valid URL string when stringified. For example the following fetch call will cause this scenario

```js
const obj = { toString: () => 'https://elastic.co' };
const init = { ... }; // it can also be undefined 
window.fetch(obj, init);
```

The result is the task being scheduled does not have a reference to a request but to a plain object. This happens regardless if the second parameter (init) is defined or not. It is important to get the request right because its reference is used to inject `tracestate` headers if configuration allows it. 
https://github.com/elastic/apm-agent-rum-js/blob/main/packages/rum-core/src/performance-monitoring/performance-monitoring.js#L252-L267

Thanks @DevooKim. You PR surfaced this issue. IMO I think we should always wrap in a `Request` object so we cover all scenarios. I think we can move forward with a couple of changes

always wrap in a request so we cover the plain object scenario
```suggestion
      request = new Request(input, init)
```
add a new test for a new input type
```js
{
  name: 'should fetch correctly when using URL object as input',
  fetchArgs: createFetchArgs('PUT'),
  fetchFn(fetchArgs) {
    const obj = { toString: () => fetchArgs.url }
    return window.fetch(obj, fetchArgs.options)
  }
}
```
this should go in  `fetch-patch.spec.js` L106 as another test case

Thank you very much for you contribution. I'll be around if you have questions or you push new changes :)

david-luna (changes requested):

